PR #266, opened 2012-08-02
added a JsonpHandler

### LINKED ISSUE #263: jsonp support for json format
Hi, im using this bundle for making an api, and i need to call it from a browser with jquery. but need to use jsonp callbacks...

i made a fast fix for achieve the functionality like this

``` php
if ($request->getRequestFormat() && $request->query->get("callback"))
{
     $response->setContent($request->query->get('callback') . "(" . $response->getContent() .")");            
}
```

in line 122 on ViewResponseListener.php i want to know if there is a better way to achieve this, or if in the future you will add the feture...

if any of you give me some hints in where will be the best place to add this i can make the feature and made a pull request.

my guesses are: or in the listener, or (i know its better) at the the end of createResponse (line 336) in ViewHandler.

and maybe make it configurable....

lsmith77:
fixes #263
- [x] rebase
- [x] decide on how to recognize jsonp requests

The issue with recognizing jsonp requests is the following:
If users will just send `application/javascript` in their Accept headers, we will not know if they want `js` or `jsonp` since `application/javascript` will always be mapped back to just `js`. There are four solutions:

1) use `application/javascript+jsonp` or some similar mime type that is mapped exclusively to `jsonp`
2) simply assume that `js` means `jsonp` when the `jsonp` handler is active
3) require that controllers set `jsonp` explicitly as the format `$view->setFormat('jsonp')`
4) make the mime type configurable (currently implemented)

I should not that using 1) of course does not prevent someone form using 3). However in that case the Accept header would become irrelevant and also the content would still be served with `Content-Type` set as `application/javascript+jsonp`


spwin:
I had the latest version #6edf1ee however I keep getting the error
`You have requested a non-existent service "fos_rest.view_handler"`

I solved it by changing view.xml:
`<service id="fos_rest.view_handler.default" class="FOS\RestBundle\View\ViewHandler" public="false">`
to
`<service id="fos_rest.view_handler.default" class="FOS\RestBundle\View\ViewHandler" public="false">`

And FOSRestController.php
`$this->viewhandler = $this->container->get('fos_rest.view_handler');`
to
`$this->viewhandler = $this->container->get('fos_rest.view_handler.default');`

My Symfony version is: 2.8.17

xabbuh:
@spwin see #1671

PR #580, opened 2013-10-09
[Alternative Proposal] Change redirection status code check

merk:
This is an alternative idea for handling status codes and redirects with bodies.

It should be considered instead of #575


lsmith77:
this never came up again .. so closing

PR #637, opened 2013-12-14
Fix in composer.json (require willdurand/negotiation ~1.1)

therealmikz:
Changed willdurand/negotiation version in composer.json to ~1.1 because in ~1.0 method normalizePriorities can't be called (doesn't exist)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,7 +25,7 @@
         "php":                            ">=5.3.2",
         "symfony/framework-bundle":       "~2.2",
         "doctrine/inflector":             "1.0.*",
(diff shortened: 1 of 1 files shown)


dimitrilahaye:
Same issue is back - I see it just now :-/

xabbuh:
I do not understand why this should be an issue with the `normalizePriorities()` method since `1.0.x` wasn't allowed before. However, the change still makes sense since it allows to user higher `1.x` versions too.

lsmith77:
this is already fixed in 1.7 https://github.com/FriendsOfSymfony/FOSRestBundle/blob/1.7/composer.json

Issue #666, opened 2014-01-16
Discussion about the AccessDeniedListener and 401, as defined in RFC 2616


SimonSimCity:
Hi,

I want to start a discussion about the AccessDeniedListener, introduced in #411 ...

As written in #434, this listener blocks the correct behavior for the two implemented HTTP authentications.

What does the listener actually handle? Are there use-cases for a 403 exception, that just this listener can handle, or is it just used when the `form` firewall-handler or a similar one is activated?

I would really like to have #434 fixed, or mentioned in the docs, that you invalidate the 401 responses by activating the AccessDeniedListener.

Because, according to the specs, any 401 response, that does not provide a `WWW-Authenticate` header field, is not valid. Read http://www.ietf.org/rfc/rfc2616.txt 10.4.2 for more information on that.

@lsmith77 I see this implementation now as improper as #596 in this bundle :) What do you think?

Issue #74, opened 2011-06-16
add a listener to prevent redirect to login page


lsmith77:
see http://groups.google.com/group/symfony-devs/browse_thread/thread/5af1120c1a6b2ab5 and http://groups.google.com/group/symfony-devs/browse_thread/thread/ebcceb9eb846e08b

PR #730, opened 2014-03-28
[WIP] Fixed incompatibility with firewalls implementation for handling AuthenticationFailureHandlerInterface

### LINKED ISSUE #434: AccessDeniedListener doesn't set correct header on 401 status code
Since #411, AuthenticationException are handled by AccessDeniedListener to return an HTTP 401 Unauthorized.
The problem it that according to [the spec](http://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.2) this should come with an _WWW-Authenticate_ header specifying authentication method. This looks like this `WWW-Authenticate: Basic realm="Restricted Area"` for a basic auth scheme.

The purpose of #411 is to prevent redirection to the login form when we simply expect JSON from client-side for example. But this redirection only occurs in the specific case of using Login form method. The problem is that we need to know what this header value needs to be according to authentication method.
This seems to be already handled at SecurityBundle level with "entry points".

I'm not sure how we should proceed here.

SimonSimCity:
Fixed #434, becuase the original AuthenticationFailureHandler defined in the firewall is used instead of this one.

Details are provided in #666 ...

### DIFF
--- a/EventListener/AccessDeniedListener.php
+++ b/EventListener/AccessDeniedListener.php
@@ -79,7 +79,7 @@ class AccessDeniedListener extends ExceptionListener
     public static function getSubscribedEvents()
     {
         return array(
-            KernelEvents::EXCEPTION => array('onKernelException', 5),
+            KernelEvents::EXCEPTION => array('onKernelException', -16),
         );
     }
 }


stof:
@lsmith77 I would need to understand the purpose of this listener first. I don't know what it was designed for:

If the goal was to prevent the redirection to the login form for people using the same firewall for the web and the API, I know what the issue is: the listener in Symfony Security is checking the current exception or any linked exception (the reason was to be able to handle security thrown inside Twig because of the wrapping in a Twig_Error_Runtime). Given your listener sets the AccessDeniedException as the linked exception of your new instance, the listener of the security component handles it again after your own. Changing the priority to run after the listener of the Security component would not prevent redirections to a login form (running too late). The only way to fix it would be to avoid linking the exception in the new one.

If the goal is to handle the case of API-only firewalls which don't have to redirect at all, this should not be handled by a custom exception listener at all but using the listener shipped in the component to call an appropriate entry point (i.e. not the entry point redirecting to a login page). In this case, the listener here will indeed hurt you (and changing the priority is only part of the fix, as it will not be called anymore for these exceptions as they will be covered by the previous listener).

lsmith77:
see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/74 for why the listener was added

lsmith77:
@stof can you have another look?

stof:
OK, so it seems that the use case is to handle the case of firewalls covering both the web interface and the API. This means that this PR is the wrong fix (it simply makes this listener useless by putting it at a priority where it will always consider that it has nothing to do).
See my previous comment about the actual way to make it work.

This listener would deserve adding some functional tests to ensure its behavior (and better documentation about its goal).

lsmith77:
ok thx .. removing the 1.5.0 milestone for now ..

stof:
Well, the fix is not complex. It could go in 1.5 (writing functional tests would take more time). But in any case, this PR should be rejected, as it is not the right approach to the fix.

lsmith77:
sure .. if someone gets it done soon .. I will of course it put it into 1.5.0

Issue #796, opened 2014-06-11
Import routes from a controller directory


jorgebg:
With `SensioFrameworkExtraBundle` you can [import routes from a controller directory](http://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/annotations/routing.html#activation):

```yaml
# app/config/routing.yml
blog:
    resource: "@SensioBlogBundle/Controller"
    type:     annotation
```

Can [`FosRestBundle`](https://github.com/FriendsOfSymfony/FOSRestBundle) do the same?

```yaml
# app/config/routing.yml
blog:
    resource: "@SensioBlogBundle/Controller"
    type:     rest
```

Ener-Getick:
Fixed by https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1339

PR #443, opened 2013-04-20
[WIP] Support Using Directories as a Routing Resource

ChadSikorra:
This change lets you import all controllers in a directory using a syntax like `@AcmeDemoBundle/Controller/Api`. I recently wanted to do this in a project and realized it wasn't currently possible. This fixes https://github.com/FriendsOfSymfony/FOSRestBundle/issues/189

PR #798, opened 2014-06-18
[WIP] Import routes from a controller directory

jorgebg:
Import routes from a controller directory #796

### DIFF
--- a/Routing/Loader/RestRouteLoader.php
+++ b/Routing/Loader/RestRouteLoader.php
@@ -69,6 +69,10 @@ class RestRouteLoader extends Loader
      */
     public function load($controller, $type = null)
     {
+        if (substr($controller, -1) === '/') {
+            return $this->loadDirectoryResource($controller, $type);
+        }
+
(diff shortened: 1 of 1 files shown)


xabbuh:
What is the state here? Do you still work on it @jorgebg?

lsmith77:
btw there is also https://github.com/FriendsOfSymfony/FOSRestBundle/pull/443

@xabbuh I would say just take it over :)

JHGitty:
ping @xabbuh Do you like to take it over? #798 (this) and #443 don't have time for it.

Ener-Getick:
:+1: would be great but IMO we should create a ``RestDirectoryLoader`` instead of using the ``RestRouteLoader``.
Maybe even create a specific type ``rest_directory`` to simplify the new loader (but this second idea is not really user friendly...).

xabbuh:
@JHGitty @Ener-Getick I'll look into this this week.

xabbuh:
closing here in favour of #1339

PR #875, opened 2014-09-25
Use standard bundle installation template

WouterJ:
*The Symfony Documenators started an initiative to make all bundle installation documentation consistent and  easy to understand for beginners. That'll improve the DX for starters using 3rd party bundles. The template can be found here: http://symfony.com/doc/current/cookbook/bundles/best_practices.html#installation-instructions*

I would like to ask you to use this template in your documentation.

If you have any comments on the template, please open an issue in the `symfony/symfony-docs` repository. Feedback will brings us to the ultimate template.

### DIFF
--- a/Resources/doc/1-setting_up_the_bundle.rst
+++ b/Resources/doc/1-setting_up_the_bundle.rst
@@ -1,8 +1,18 @@
 Step 1: Setting up the bundle
 =============================
 
-A) Install FOSRestBundle
-------------------------
+A) Download the Bundle
+----------------------
+
+Open a command console, enter your project directory and execute the
+following command to download the latest stable version of this bundle:
+
+.. code-block:: bash
+
+    $ composer require friendsofsymfony/rest-bundle "~1.4"
+
+This command requires you to have Composer installed globally, as explained
+in the `installation chapter`_ of the Composer documentation.
 
 .. note::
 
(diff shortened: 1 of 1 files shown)


Resources/doc/1-setting_up_the_bundle.md
```diff
@@ -12,30 +24,28 @@ Step 1: Setting up the bundle
...
+            // ...
+            new FOS\RestBundle\FOSRestBundle(),
+        );
+    
         // ...
-        new FOS\RestBundle\FOSRestBundle(),
-        
-        // if you choose to use JMSSerializer, make sure that it is registered in your application
-        
-        // new JMS\SerializerBundle\JMSSerializerBundle(),
```
WouterJ: in the note about JMSerializer, it states that the user should follow its own installation instructions to set-up and install it. I don't think this bundle should cover that.

WouterJ:
Updated it (and also rebased).

Resources/doc/1-setting_up_the_bundle.rst
```diff
@@ -1,8 +1,18 @@
...
-------------------------
+A) Download the Bundle
+----------------------
+
+Open a command console, enter your project directory and execute the
+following command to download the latest stable version of this bundle:
+
+.. code-block:: bash
+
+    $ composer require friendsofsymfony/rest-bundle "~1.4"
```
lsmith77: isnt there a new syntax that automatically picks a good version?

lsmith77:
btw don't newer versions of Symfony2 provide a core serializer service?

Resources/doc/1-setting_up_the_bundle.rst
WouterJ: Yes, but we (sf doc team) recommend to use the cersion in the docs, as you want users to install the version for which the documentation is created (otherwise, 1.0 docs would have a command that installs 2.0)

WouterJ:
Yes, since Symfony 2.3 (https://github.com/symfony/FrameworkBundle/commit/98f30b0df147786f125cb0722c9db40cc91e5278), which is disabled by default btw.

lsmith77:
ah ok .. can you then update the sentence that tells users that they need to define the serializer service themselves if they do not use JMSSerializerBundle to instead tell them to simply activate the service in the framework bundle?

lsmith77:
will do this myself

lsmith77:
see https://github.com/FriendsOfSymfony/FOSRestBundle/pull/933

PR #883, opened 2014-10-07
Enable or disable removing false data for JsonToFormDecoder

meyerbaptiste:
The `JsonToFormDecoder` class remove false data like a classic form.
But, in the case where we have a form with just one checkbox or patch just one checkbox it does not work because form is not submitted due to missing of parameters in the request and so it is invalid.
This issue can be fixed by replacing the value by null instead of remove it.

So in this PR I propose that we add a container parameter for this service which can enable or disable removing false data.

I use a container parameter because it is a tricky case which does not require to implement a configuration system of the decoder but which leaves the possibility to the developer to modify it if necessary.

### DIFF
--- a/Decoder/JsonToFormDecoder.php
+++ b/Decoder/JsonToFormDecoder.php
@@ -19,19 +19,6 @@ namespace FOS\RestBundle\Decoder;
 class JsonToFormDecoder implements DecoderInterface
 {
     /**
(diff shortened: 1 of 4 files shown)


Adapik:
OK, now how I can distinguish empty field from false field? No chance

PR #902, opened 2014-11-11
close #901, we can add a incompatibles attribute

### LINKED ISSUE #901: annotation for query params that are mutually exclusives
Hello let's admit i have this api 

```
    /tickets?search=keyword
   /tickets?status=open
```

however right now I don't have 

```
    /tickets?search=keyword&status=open
```

Of course for that simple case i can either 

  * put a if in my code and do a "bad request"  if both are present => but when i start to have 5 kind of different filtering, it starts to add a lot of if
  * tell in the documentation which one has priority => misleading

so would it be possible in the QueryParam annotation to have an additional attribute something like 

      /**
       * @Annotations\QueryParam(
       *    name="status",
       *    default=null,
       *    nullable=true,
       *    incompatible_with=["search"]
       *    description=""
       * )
       */

?

allan-simon:
to QueryParam to detect parameters that can't appears
together in the same API call

### DIFF
--- a/Controller/Annotations/Param.php
+++ b/Controller/Annotations/Param.php
@@ -37,6 +37,8 @@ abstract class Param
     public $nullable = false;
     /** @var bool */
     public $allowBlank = true;
+    /** @var array */
+    public $incompatibles = array();
 
     /**
      * @return string
--- a/Request/ParamFetcher.php
+++ b/Request/ParamFetcher.php
@@ -195,6 +195,8 @@ class ParamFetcher implements ParamFetcherInterface
             );
         }
 
+        $this->checkNotIncompatibleParams($config);
+
         if (null === $config->requirements || ($param === $default && null !== $default)) {
             return $param;
         }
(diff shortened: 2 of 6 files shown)


lsmith77:
also needs a rebase

allan-simon:
by the way I was wondering if it would make sense to add explanation in the documentation about what each fields does,  for example i'm a bit confused by the difference between "key" and "name" , i mean i see there can be a difference between the name use to get in php code and the name in the actual  GET parameters 
(and I think it will be the same for that "incompatibles"  field i'm adding)

allan-simon:
I still have one test left but I can't get the phpunit to run on my machine (i got it running for other project) but here it gives me 

```
PHP Fatal error:  Class 'Doctrine\Common\Annotations\AnnotationRegistry' not found in /XXX/git/github/FOSRestBundle/Tests/bootstrap.php on line 20

```
when I try to run it

allan-simon:
test are now passing , the new added function is fully covered , if it's ok (with maybe a little more doc?) then I will squash my commits

lsmith77:
yeah .. there is just one extra line I noted to remove. with docs I shall merge.

allan-simon:
I've rebased, for the extra new line can you tell me if it's still there ? I've also added doc

Issue #868, opened 2014-09-16
Camel keys normalizer not working


Sukhrob:
I know it is not a place for asking questions. But, I had to create an issue as there is obviously something wrong or missing in camel keys normalizer. Please, refer to this link in stackoverflow. Thanks.
http://stackoverflow.com/questions/25794050/fosrestbundle-camel-keys-normalizer-not-working

PR #903, opened 2014-11-13
Added an option to normalize form data

florianv:
The normalizer is only applied when the data is decoded which is not the case with `multipart/form-data`.

Fix issue https://github.com/FriendsOfSymfony/FOSRestBundle/issues/868.


DependencyInjection/FOSRestExtension.php
```diff
@@ -160,9 +160,11 @@ public function load(array $configs, ContainerBuilder $container)
...
 
             $arrayNormalizer = $config['body_listener']['array_normalizer'];
-            if (null !== $arrayNormalizer) {
-                $container->getDefinition($this->getAlias().'.body_listener')
-                    ->addMethodCall('setArrayNormalizer', array(new Reference($arrayNormalizer)));
+
+            if (null !== $arrayNormalizer['service']) {
+                $bodyListener = $container->getDefinition($this->getAlias().'.body_listener');
+                $bodyListener->addArgument(new Reference($arrayNormalizer['service']));
+                $bodyListener->addArgument($arrayNormalizer['forms']);
```
florianv: Thanks for the suggestion, but `replaceArgument()` will throw an exception if the argument was not added (it only replaces it).

xabbuh: Hm, actually you're right.

lsmith77:
this just needs a rebase .. right?

florianv:
@lsmith77 yes. The rebase has been done

PR #906, opened 2014-11-18
[DX] Improved error message on wrong error

WouterJ:
Someone on IRC got `Warning: constant(): Couldn't find constant \FOS\RestBundle\Util\Codes:: in /../vendor/friendsofsymfony/rest-bundle/FOS/RestBundle/DependencyInjection/FOSRestExtension.php line 137` when configuring an empty code. This will make the error a bit more debuggable and nicer :)


WouterJ:
@lsmith77 added some the tests

Tests/DependencyInjection/FOSRestExtensionTest.php
```diff
@@ -399,6 +400,39 @@ public function testLoadOkMessagesClass()
...
+        $data = array(
+            null,
+            'HTTP_NOT_EXISTS',
+            'some random string',
+            true,
+        );
+
+        return array_map(function ($i) {
+            return array($i);
+        }, $data);
```
xabbuh: What is the advantage in using `array_map()` here?

WouterJ: tests have to return an array of arrays. I don't like it for single values

PR #910, opened 2014-11-22
Fixed validation for Request parameter that is an array without default value

eugef:
If Request parameter is an array without default value and parameter is not set - then an error should be thrown that "parameter value is not an array".

Steps to reproduce:
- Create a Request parameter with the following config

```
@RequestParam(name="clients", array=true)
```
- In a post request do not send this parameter at all

Actual result (before the fix):
- 'clients' parameter is set to an empty array 

Expected result (with the fix):
- Error "Request parameter value of 'clients' is not an array"

If empty array should be a default value if parameter is not set, then 'strict' should be set to false, example:

```
@RequestParam(name="clients", array=true, strict=false)
```


lsmith77:
@eugef can you have a look at https://github.com/FriendsOfSymfony/FOSRestBundle/issues/960 ?

PR #935, opened 2015-01-15
Update index.rst

dnagirl:
fix help document links.

### DIFF
--- a/Resources/doc/index.rst
+++ b/Resources/doc/index.rst
@@ -6,7 +6,7 @@ Installation
 
 Installation is a quick (I promise!) one-step process:
 
-1. :doc:`Setting up the bundle <1-setting_up_the_bundle>`_
+1. `Setting up the bundle <./1-setting_up_the_bundle.rst>`_
 
 Bundle usage
 ------------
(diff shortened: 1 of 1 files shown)


xabbuh:
@dnagirl The documentation recently changed to the reStructuredText markup so that it can be rendered on symfony.com. The syntax you used for the links is Markdown which won't be parsed correctly by Sphinx (the build tool for rst docs).

xabbuh:
@javiereguiluz Is there a schedule for when the docs for FOSRestBundle and FOSUserBundle will be available on symfony.com?

dnagirl:
@xabbuh, I used the format described here: http://sphinx-doc.org/rest.html#hyperlinks  That's not correct?

In any event, is there a format that will work for both Symfony.com and github? Those broken links are annoying.

xabbuh:
@dnagirl The linked format should only be used for external links. The reason is that the referenced document won't be validated when used for internal documents (opposed to what is done when using the `doc` or `ref` roles).

I'm afraid that there is nothing that could be done for now. Hopefully, the docs will be rendered soon on symfony.com so that the link can be updated in the readme file.

Issue #299, opened 2012-09-17
Allow pass additional variables to the Twig template


waiting-for-dev:
I don't know if I'm missing something, but I think that what I'm trying to do is very common, but I think it's not possible with FOSRestBundle.

I'm using the same action to deal with an `html` and a `json` response, just suffixing the later one route with `.json`.

The `json` is working with no problem, but when the response is `html` I need to pass an additional variable to the `Twig` template, and I don't know and I don't find how to do it. At the moment, my code is:

    // $entity = get my data;

    $name = 'foo';
    
    $view = $this->view($entity, 200)
       ->setTemplate('MyBundle:MyController:myTemplate.html.twig')
       ->setTemplateVar('entity')
       ;
    
    return $this->handleView($view);

How can I pass `$name` to `myTemplate.html.twig`? What I would need is something like:

    //...
    ->setTemplate('MyBundle:MyController:myTemplate.html.twig', array('name' => $name))
    //...

PR #936, opened 2015-01-15
Pass additional variables in template system

Cript:
Hello. I want to pass additional variables in template. For example category page with goods. 
I think it is good way to do this.

### DIFF
--- a/View/View.php
+++ b/View/View.php
@@ -25,6 +25,7 @@ use JMS\Serializer\SerializationContext;
 class View
 {
     private $data;
+    private $templateData;
     private $template;
     private $templateVar;
     private $engine;
@@ -127,6 +128,21 @@ class View
     public function setData($data)
     {
         $this->data = $data;
+
+        return $this;
+    }
+
+    /**
+     * Set template variable
+     *
+     * @param $variableName
+     * @param $data
+     *
+     * @return View
+     */
+    public function setTemplateData($variableName, $data)
+    {
+        $this->templateData[$variableName] = $data;
 
         return $this;
     }
@@ -315,6 +331,16 @@ class View
     public function getData()
     {
         return $this->data;
+    }
+
+    /**
+     * Gets the template data.
+     *
+     * @return mixed|null
+     */
+    public function getTemplateData()
+    {
+        return $this->templateData;
     }
 
     /**
--- a/View/View.php
+++ b/View/View.php
@@ -25,7 +25,7 @@ use JMS\Serializer\SerializationContext;
 class View
 {
     private $data;
-    private $templateData;
+    private $templateData = array();
     private $template;
     private $templateVar;
     private $engine;
--- a/View/View.php
+++ b/View/View.php
@@ -135,14 +135,13 @@ class View
     /**
      * Set template variable
      *
-     * @param $variableName
-     * @param $data
+     * @param  array|callable $data
      *
      * @return View
      */
-    public function setTemplateData($variableName, $data)
+    public function setTemplateData($data = array())
     {
-        $this->templateData[$variableName] = $data;
+        $this->templateData = $data;
 
         return $this;
     }
--- a/View/ViewHandler.php
+++ b/View/ViewHandler.php
@@ -380,29 +380,24 @@ class ViewHandler extends ContainerAware implements ConfigurableViewHandlerInter
     public function prepareTemplateParameters(View $view)
     {
         $data = $view->getData();
-        $templateData = $view->getTemplateData();
-        if ($data instanceof FormInterface) {
-            $preparedData = array_merge(
-                array($view->getTemplateVar() => $data->getData(), 'form' => $data->createView()),
-                $templateData
-            );
-            return $preparedData;
-        }
 
-        if (empty($data) || !is_array($data) || is_numeric((key($data)))) {
-            $preparedData = array_merge(
-                array($view->getTemplateVar() => $data),
-                $templateData
-            );
-            return $preparedData;
+        if ($data instanceof FormInterface) {
+            $data = array($view->getTemplateVar() => $data->getData(), 'form' => $data);
+
+        } elseif (empty($data) || !is_array($data) || is_numeric((key($data)))) {
+            $data = array($view->getTemplateVar() => $data);
         }
 
         if (isset($data['form']) && $data['form'] instanceof FormInterface) {
             $data['form'] = $data['form']->createView();
-            $data = array_merge($data, $templateData);
         }
 
-        return $data;
+        $templateData = $view->getTemplateData();
+        if (is_callable($templateData)) {
+            $templateData = call_user_func($templateData, $this, $view);
+        }
+
+        return array_merge($data, $templateData);
     }
 
     /**
--- a/View/ViewHandler.php
+++ b/View/ViewHandler.php
@@ -380,16 +380,26 @@ class ViewHandler extends ContainerAware implements ConfigurableViewHandlerInter
     public function prepareTemplateParameters(View $view)
     {
         $data = $view->getData();
+        $templateData = $view->getTemplateData();
         if ($data instanceof FormInterface) {
-            return array($view->getTemplateVar() => $data->getData(), 'form' => $data->createView());
+            $preparedData = array_merge(
+                array($view->getTemplateVar() => $data->getData(), 'form' => $data->createView()),
+                $templateData
+            );
+            return $preparedData;
         }
 
         if (empty($data) || !is_array($data) || is_numeric((key($data)))) {
-            return array($view->getTemplateVar() => $data);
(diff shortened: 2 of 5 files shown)


lsmith77:
interesting .. so your use case is that you only want to send additional variables to the template but not when using the serializer? I guess right now my answer for such a requirement would be to have a wrapper object that simply does not include the unwanted data in the serialization mapping.

Cript:
> interesting .. so your use case is that you only want to send additional variables to the template but not when using the serializer?

Yes

> I guess right now my answer for such a requirement would be to have a wrapper object that simply does not include the unwanted data in the serialization mapping.
Could you explain more?
Wrapper object for what?

lsmith77:
in jmsserializer you can choose which properties should be serialized and which should be skipped.

Cript:
For example, let's discuss the product page.
In a json view I only need array with products. I can do it like this:
```
$view = $this->view($products, 200)
        ->setTemplate("SiteViewBundle:Category:show.html.twig")
        ->setTemplateVar('products')
```

But in a html view I need not only products array, but also the category object with name, children and other properties to show them in the html page.
It is possible to transfer category object only into html template now?
If it is not possible I suggest do it like this:

```
$view = $this->view($products, 200)
            ->setTemplate("SiteViewBundle:Category:show.html.twig")
            ->setTemplateVar('products')
            ->setTemplateData('category', $category);
```

In then in twig for example:

```
<h1>{{ category.name }}<h1>
{% for product in products %}
    {{ product.name }}
{% endfor %}
```

Cript:
I have found this discussion just now with the same proplem.
https://github.com/FriendsOfSymfony/FOSRestBundle/issues/299

giosh94mhz:
:+1: on this feature.

I had a similar case when creating yet another `UserController`. REST implementation (i.e. serializer) needed only the user data, while the twig template needed some more form, and data to create a rich interface.

Please consider also some sort of "lazy" initialization of template data, to avoid overheads. I see at least two approach for this:

1) simply add a a shortcut in `FOSRestController` like:
```php
private function isFormatTemplating($_format = null)
{
    return $this->get('fos_rest.view_handler')->isFormatTemplating(
        $_format ?: $this->get('request')->getRequestFormat()
    );
}
```
so that developer can easily do
```php
    if ($this->isTemplatingFormat()) {
        // slowest query ever
        $view->setTemplateData('slow_query_result', $values);
    }

```
2) allow callback for `setTemplateData` so that something like this is possible:
```php
$view->setTemplateData('slow_query_result', function () use ($doctrine) {
    // slowest query ever
    return $values;
});
```

This second approach is better IMO.

lsmith77:
lazy loading of such optional data indeed seems like what we want to encourage. in theory it is possible that the output format isn't determined yet in the controller but at a later stage. but this does seem highly theoretical .. however especially those who do not inject the ViewHandler and also do not extend the base Controller .. lazy loading would be better so that they do not have to add this dependency.

giosh94mhz:
> output format isn't determined yet in the controller but at a later stage

I agree, that's why I think that the callback approach is better, and can be resolved inside `ViewHandler::prepareTemplateParameters` (which happens only when using templating format). Probably, in my example above, the first parameter ('slow_query_result') can be avoided at all.

> however especially those who do not inject the ViewHandler and also do not extend the base Controller .. lazy loading would be better so that they do not have to add this dependency.

I don't understand what you mean :) If both the `ViewHandler` and `FOS/View` are not used, then the Controller must return a response, and the template rendering already happened (e.g. by calling $this->render). Did I miss something?

lsmith77:
in order to determine if its a templating format inside the controller, the controller needs access to the ViewHandler .. but if you use the ViewResponseListener, you don't have the ViewHandler instance in the Controller.

giosh94mhz:
Ok, maybe I got it now. So you are encouraging solution 2, while avoiding solution 1 since the format is "unpredictable" in the Controller when using a `ViewResponseListener`. Right?

lsmith77:
yeah .. but like I said .. this is likely more a theoretical concern but if we can make the callback approach work nicely then its all good.

BTW another option is of course also to let the templates pull the additional information. of course one should not have business logic there and also not elegant in terms of error handling but this is what I do at times when I encountered the above described needs before.

lsmith77:
did you intentionally close this ticket?

Cript:
No. I did not close.

View/ViewHandler.php
```diff
@@ -380,16 +380,26 @@ public function renderTemplate(View $view, $format)
     public function prepareTemplateParameters(View $view)
     {
         $data = $view->getData();
+        $templateData = $view->getTemplateData();
         if ($data instanceof FormInterface) {
-            return array($view->getTemplateVar() => $data->getData(), 'form' => $data->createView());
+            $preparedData = array_merge(
```
giosh94mhz: I would add the following:
```diff
@@ -381,6 +381,9 @@
     {
         $data = $view->getData();
         $templateData = $view->getTemplateData();
+        if (is_callable($templateData)) {
+            $templateData = call_user_func($templateData, $this, $view);
+        }
         if ($data instanceof FormInterface) {
             $preparedData = array_merge(
                 array($view->getTemplateVar() => $data->getData(), 'form' => $data->createView()),
```
At least the `ViewHandlerInterface` (`$this`) and the `View` need to be passed to the callback, so that the format can be read correctly.

View/View.php
```diff
@@ -132,6 +133,21 @@ public function setData($data)
...
 
     /**
+     * Set template variable
+     *
+     * @param $variableName
+     * @param $data
+     *
+     * @return View
+     */
+    public function setTemplateData($variableName, $data)
```
giosh94mhz: To make the interface clearer (IMO) and compatible with the callback patch below, I would change the method `setTemplateData` like `setData`, so:
```php
public function setTemplateData($data = [])
{
    $this->templateData = $data;

    return $this;
}
```

View/View.php
```diff
@@ -25,6 +25,7 @@
 class View
 {
     private $data;
+    private $templateData = array();
```
giosh94mhz: Minor fix: it's better to move this initialization to the constructor, since it is already present

lsmith77: I prefer keeping the initialization here

View/View.php
lsmith77: note we still support php 5.3 in this bundle

giosh94mhz: Right. Then `[]` should be replaced by `array()`

lsmith77:
ping

View/View.php
Cript: What about this situation?
```
$view = $this->view($products, 200)
    ->setTemplate("SiteViewBundle:Category:show.html.twig")
    ->setTemplateVar('products')
    ->setTemplateData('category', $category);
/* more code */
$view->setTemplateData('other_template_data', $otherTemplateData)
```
I think it`s not necessary to change setTemplateData function.

View/ViewHandler.php
Cript: Theare are may be few variables in ```$templateData```
I think this code is correct:
```
    foreach($templateData as $keyTemplateData => $valueTemplateData) {
        if (is_callable($valueTemplateData)) {
            $templateData[$keyTemplateData] = call_user_func($valueTemplateData, $this, $view);
        }
    }
```

View/View.php
giosh94mhz: I know it is not required, but IMO is more consistent with current interface (only one 'array' passed to `view`/`setData`) and will guide the developer to a cleaner code. Also, if you allow setting one variable at a time as in `Request` object, you need to implement a container interface or it will be impossible to unset properties.

Your example can be refactored as:
```php
$viewData = array('category' => $category);
/* more code */
$viewData['other_template_data'] = $otherTemplateData;

return $this->view($products, 200)
    ->setTemplate("SiteViewBundle:Category:show.html.twig")
    ->setTemplateVar('products')
    ->setTemplateData($viewData)
;
```
or even better with the lazy approach:
```php

return $this->view($products, 200)
    ->setTemplate("SiteViewBundle:Category:show.html.twig")
    ->setTemplateVar('products')
    ->setTemplateData(function () use ($category, $container) {
        $viewData = array('category' => $category);
        /* more code using services & co */
        $viewData['other_template_data'] = $otherTemplateData;
        // adding 'products' key not recommended :)
        return $viewData;
    })
;
```

lsmith77: @giosh94mhz you can also propose an alternative PR based on this PR .. might make the discussion easier.

Cript:
@giosh94mhz , I have tried you code. It is very easy to use.

lsmith77:
@Cript ok .. so theb we can close this PR in favor of the new one?

Cript:
@lsmith77, ok.

giosh94mhz:
Thanks for the review @Cript

PR #943, opened 2015-01-20
Change private access to protected

waldermort:
The class cannot be extended without copying the 'getControllerLocator' method.

### DIFF
--- a/Routing/Loader/RestRouteLoader.php
+++ b/Routing/Loader/RestRouteLoader.php
@@ -98,7 +98,7 @@ class RestRouteLoader extends Loader
      *
      * @throws \InvalidArgumentException
      */
-    private function getControllerLocator($controller)
+    protected function getControllerLocator($controller)
     {
         $class  = null;
         $prefix = null;


lsmith77:
can you explain a bit what your use case is for extending?

waldermort:
@lsmith77 Well, simple actually, I want to dynamically prefix the route names/paths. In creating an admin generator, I have a treeview in the sidebar. Each Entity controller can be assigned (configured in yaml) to an admin group/subgroup. Due to conflicting routes and various permissions, a route prefix, depending on the position in the tree, is required. Doing this dynamically, allows me to specify a controller belongs in the admin zone by statically prefixing it with '/admin', and allows the controller to inherit the permissions of it's group.

lsmith77:
@Tobion / @stof can you give some feedback here on what the best approach is?

stof:
Prefixing the route paths of a controller should be done by putting the prefix when importing the routing for this controller IMO rather than extending the loader. It is already supported in the Routing component itself.
and FOSRestBundle already provides a way to prefix route names

waldermort:
@stof Notice the word 'dynamic' in my reply. You are talking about static prefixes. If I take the static approach I have to update the prefix any time I want to relocate a controller within the tree. This is going against my design goal. I have a simple requirement that when configuring the admin layout, I can say this controller belongs to this, this and that group. I've already coded this component and it works well.

stof:
what do you mean by dynamic prefixes exactly ?

waldermort:
@stof I have a treeview (menu) with categories like E-Commerce, Settings, User etc. When I add entity controllers to the admin, I want each route to include the slug of its parent category. So instead of routes like /admin/groups they will become /admin/user/groups. Yes, this can be done with static prefixes, but when I want to move a controller from say group Foo to group Bar, I will have to also update those route prefixes. My design fetches the slug from the group and prefixes it while building the routes. This way I can control the grouping from a single yaml file.

stof:
Define a custom route loader, which will read your own Yaml file and then import the resources for each controller with the appropriate prefix (which can then depend on this Yaml file)

waldermort:
Defining a custom loader won't solve this. First I will lose the 'rest' type, second I will still have to call into the rest loader to scan the class. I can do exactly what I want by overloading the RestRouteLoader and modifying the load method. The whole point of this pull request is to point out that I cannot redefine the load method without also copying the 'getControllerLocator' method. As you know, a private method in a parent class is not accessible. I'm not asking how to do something, I've already done it.

stof:
No you won't. Your custom loader will import the rest resources with a prefix.

stof:
And we are telling you that there is better ways to do what you have done. Inheritance is often the worse way to extend code, and switching the visibility to protected means we are then forced to preserve BC on this new extension point. this is why we don't blindly introduce such extension points everywhere, especially when there are other better ways to achieve the use case.

waldermort:
@stof Then let me ask you something. Why are all the service classes, in this bundle, defined as parameters? Doing such a thing implies that the end user may redefine a class which the service uses. But, these services are dependent on each other in that the class name of one is defined in the constructor of another. (Wouldn't it be wiser to use interfaces?). The problem now is that we can't replace the class, we MUST extend it. Now, with all the private methods, how is that possible without copying the code?

Here is a prime example, the route names generated by the bundle all include the class name. For example 'createAction' would output something like 'create_user'. That's fine. But, 'moveUpAction' will produce 'move_user_up'. My code requires building an array of action to route names. So, I would have something like 'create' => 'create_user'. A simple explode() on the generated name does the job well. Given the second example above, how do I remove the '_user' part? It's no longer the last item. Please don't tell me preg_replace as it just may include more than one instance. The best approach is to not include it in the first place. Looking into the code, I need to modify the 'getHttpMethodAndResourcesFromMethod' method of the 'RestActionReader' class (line 312). It should be as simple as extending the class, rewriting the method and changing the class name parameter to that of my own. Not possible, it's another private method.

Since you guys seem to be unwilling to use protected methods, the only way I'm going to be able to do this is by forking. Now, to keep up with bug fixes, I have to manually update my fork instead of letting composer handle it for me.

PR #947, opened 2015-01-23
don't use deprecated routing features

xabbuh:


lsmith77:
when were these features deprecated?

xabbuh:
In 2.2 if I am not mistaken.

lsmith77:
ok .. I just noticed that in the composer.json we still support 2.2, while in travis we only test against 2.3:
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/composer.json#L26
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/.travis.yml

I guess we should also bump this to 2.3 in composer. Meaning there might be even more stuff to remove then.

lsmith77:
see also https://github.com/FriendsOfSymfony/FOSRestBundle/pull/952

PR #953, opened 2015-01-27
remove format parameter in ExceptionController::showAction()

lsmith77:
fix for https://github.com/FriendsOfSymfony/FOSRestBundle/issues/946

### DIFF
--- a/Controller/ExceptionController.php
+++ b/Controller/ExceptionController.php
@@ -52,13 +52,12 @@ class ExceptionController extends ContainerAware
      * @param Request                                    $request
      * @param HttpFlattenException|DebugFlattenException $exception
      * @param DebugLoggerInterface                       $logger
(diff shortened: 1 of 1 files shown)


stof:
:+1:

Controller/ExceptionController.php
```diff
@@ -76,7 +75,7 @@ public function showAction(Request $request, $exception, DebugLoggerInterface $l
             ));
         }
 
-        $format = $this->getFormat($request, $format);
+        $format = $this->getFormat($request, $request->getRequestFormat());
```
lsmith77: actually I am wondering if we can do anything smarter here .. obviously we cannot refactor ``getFormat()`` too much without a BC break but obviously now a bunch of logic in the method is now sort of superfluous by default.

PR #954, opened 2015-01-27
[DependencyInjection] Cleaned the extension

florianv:
This PR breaks the loading into different sections which are easier to read.


lsmith77:
looks good .. hard to really see in the PR if nothing is lost .. gotta trust the tests, huh? :)

florianv:
@lsmith77 Yes I think some tests would have failed if I forgot something. I re-checked it and _normally_ it should be good :)

Issue #760, opened 2014-05-08
Disabling listeners per route


florianv:
Let's say you have an api at `/api/*` and you use the body listener with the camel_keys `array_normalizer` option. You also have a `/token` endpoint managed by `FOSOAuthServerBundle`.

The body listener being registered to all requests, it will transform the `grant_type` property into `grantType` making the requests to `/token` invalid.

I think it could be nice to allow blacklisting some listeners for a set of routes.

PR #780, opened 2014-05-23
[BodyListener] introducing "disable_routes" proof of concept

backender:
Resolve https://github.com/FriendsOfSymfony/FOSRestBundle/issues/760 

/cc https://github.com/FriendsOfSymfony/FOSRestBundle/issues/763

This is a proof of concept how we could avoid body listening on specific routes.
It definitely needs to be finalized, especially since `$request->attributes->get('_route')` didn't work (any hints)? For now it's just a hard coded url for demo purposes.

PR #955, opened 2015-01-27
Introduced the zone concept

florianv:
This is an attempt to fix https://github.com/FriendsOfSymfony/FOSRestBundle/issues/760 by using a more global approach than https://github.com/FriendsOfSymfony/FOSRestBundle/pull/780.

To resume the issue is that FOSRest's listeners are registered to all requests and may conflict with other parts of the application (non-REST).

### Feature

This PR introduces a new configuration value allowing to specify request matchers to determine if the listeners will be enabled or not. By default, all listeners are still enabled to all requests.

```yaml
# app/config/config.yml
fos_rest:
    zone:
        - { path: ^/api }
```

In this example, FOSRest listeners will only be enabled for requests whose path matches `^/api`. It is also possible to specify an array of zones:

```yaml
# app/config/config.yml
fos_rest:
    zone:
        - { path: ^/api }
        - { path: /test, ips: 127.0.0.1 }
```

### Inner working

It works with an early request listener matching requests and storing whether the request matches a configured zone as request attribute. This request attribute can be checked by succeeding listeners to determine if they must be enabled.


lsmith77:
This indeed looks interesting and is sort of what I had in mind as well. I realize that the approach you take is essentially BC but for now lets target this for 2.0.

florianv:
@lsmith77 Can this feature also replace the `format_listener` rules if it targets 2.0 ?

It seems the `rules` are solving the same issues than the `zone` but only for the `format_listener`.

lsmith77:
yeah for 2.0 I think it would be most efficient if the zones are responsible for all of this. ideally the zone even registers the listeners. in end effect .. most features of the Bundle can then be configured for specific parts of the API

florianv:
It seems it would be very flexible to be able to configure the listeners per zone. 

In this case, we could use a [service configurator](http://symfony.com/doc/current/components/dependency_injection/configurators.html) for the listeners, that will match requests and configure them accordingly at runtime. The `ZoneMatcherListener` would be dropped.

ahilles107:
:100:

lsmith77:
ok it looks like 2.0 is on the horizon due to #1074 .. so lets get back on this as well

florianv:
@lsmith77 I'm happy to finish this if it's still relevant (I think so)

lsmith77:
Yeah .. I personally think it would be better if people used multiple Kernels .. but the current trend with a single AppBundle seems to be going in the opposite direction. so I think your idea is the best way to address this.

florianv:
So to resume we could have something like that with the listeners configurable per zone:

```yaml
# app/config/config.yml
fos_rest:
    cache_dir: '%kernel.cache_dir%/fos_rest'
    zones:
        api_zone:
            path: ^/api 
            body_listener:
                array_normalizer: fos_rest.normalizer.camel_keys        
        auth_zone:
            path: login
            body_listener: true
```

To reduce runtime configuration and complexity, we could always create the listeners for each zone with a name like `fos_rest.listener.{zone_name}.allowed_methods`. 

Then an early `kernel.request` listener would check which zone is matched and remove the unused listeners using `EventDispatcher::removeListener()`.

lsmith77:
I am not sure what the performance impact is but I also agree that ``removeListener`` is better than ``addListener`` .. +1 from me

florianv:
If we have to care about sub requests, we will have to add the listeners too! For example if we have two zones X and Y:

- master request, zone X is matched, we remove all Y listeners
- sub request, zone Y is matched, we need to add zone Y listeners and remove the ones from zone X

ahilles107:
How it looks? Any progress?

lsmith77:
please rebase on master

Issue #960, opened 2015-02-04
[QueryParam] Nullable array with validation


eliecharra:
In **1.4.2** I had the following behaviour to allow an optional array param and array item validation when array is not null :

```php
use FOS\RestBundle\Controller\Annotations\QueryParam;

     /*
      * @QueryParam(array=true, name="category_ids", requirements="\d+", strict=true, nullable=true, description="Array of container categories ids : category_ids[]=x&category_ids[]=x")
      */
```

Now in 1.5.0 when I submit an empty param I get
```json
{
  "code": 400,
  "message": "Query parameter value of 'category_ids' is not an array"
}
```

Seems like #910 is breaking this in 1.5.0.
I would like to allow an empty array param but dont wanna loose the validation.
If I set strict to false i'm loosing array items validation.

Easy to reproduce with the above annotation.

Is my initial annotation badly crafted or a BC break ?
I have not looked into the code, but it seems like nullable param for an array is not considered.

eugef:
If parameter is array then `strict` is used to determine if parameter is optional or not. 
Probably this is wrong and `nullable` should be used instead (for now `nullable` is not used at all for array parameter validation - see https://github.com/FriendsOfSymfony/FOSRestBundle/blob/2437a4fafb232bccac0dff71f9d2ada97e19155c/Request/ParamFetcher.php#L138).

If we use both `strict` and `nullable` options then following configurations are possible that cover all cases:
* `nullable=true` and `strict=true` - array parameter is optional, items validation is strict
* `nullable=false` and `strict=true` - array parameter is required, items validation is strict
* `nullable=true` and `strict=false` - array parameter is optional, items validation is not strict
* `nullable=false` and `strict=false` - array parameter is required, items validation is not strict

@lsmith77 , what do you think?

eliecharra:
> nullable=true and strict=true - array parameter is optional, items validation is strict
> nullable=false and strict=true - array parameter is required, items validation is strict
> nullable=true and strict=false - array parameter is optional, items validation is not strict
> nullable=false and strict=false - array parameter is required, items validation is not strict

@eugef @lsmith77 This behavior is better.
strict param shoud be user for validation **only**.
and nullable param shoud be user for optional purpose

lsmith77:
+1

ruscon:
+1 
Can you add this hotfix to the current release ?

eugef:
[...]

PR #963, opened 2015-02-05
[QueryParam] Nullable array with validation

eliecharra:
Fixes FriendsOfSymfony/FOSRestBundle#960

### DIFF
--- a/Request/ParamFetcher.php
+++ b/Request/ParamFetcher.php
@@ -123,10 +123,6 @@ class ParamFetcher implements ParamFetcherInterface
             $strict = $config->strict;
         }
 
-        if ($config->array && (null !== $default || !$strict)) {
-            $default = (array) $default;
-        }
-
         if ($config instanceof RequestParam) {
             $param = $this->request->request->get($config->getKey(), $default);
         } elseif ($config instanceof QueryParam) {
@@ -136,8 +132,14 @@ class ParamFetcher implements ParamFetcherInterface
         }
 
         if ($config->array) {
+
+            if (($default !== null || !$strict) || $nullable)
+            {
+                $default = (array) $default;
+            }
+
             if (!is_array($param)) {
-                if (!$nullable) {
+                if ($strict && !$nullable) {
                     throw new BadRequestHttpException(
                         sprintf("%s parameter value of '%s' is not an array", $paramType, $name)
                     );
--- a/Request/ParamFetcher.php
+++ b/Request/ParamFetcher.php
@@ -133,8 +133,7 @@ class ParamFetcher implements ParamFetcherInterface
 
         if ($config->array) {
 
-            if (($default !== null || !$strict) || $nullable)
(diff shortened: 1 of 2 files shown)


eliecharra:
I'm fixing tests ...

Request/ParamFetcher.php
```diff
@@ -136,8 +132,14 @@ public function get($name, $strict = null)
         }
 
         if ($config->array) {
+
+            if (($default !== null || !$strict) || $nullable)
+            {
```
lsmith77: curly bracket needs to be on the previous line

lsmith77:
so with this null will then be converted to array() ?

eliecharra:
@lsmith77 Here is a little summary to clarify the situation, let me know if it's ok for you.

Requirements | Default value | Given value | Strict | Nullable | Result
|-------------------|-------------------|------------------|--------|-------------|---------
| `\d+` | `null`              |      `null`       | `false` |  `false` | `Array()`
| `\d+` | `null`              |      `null`       | `true` |  `true` | `Array()`
| `\d+` | `null`              |      `null`       | `false` |  `true` | `Array()`
| `\d+` | `null`              |      `null`       | `true` |  `false` | Query parameter value of 'X' is not an array
| `\d+` | `42`                |      `null`       | `true` |  `true` | `Array(42)`
| `\d+` | `42`                |      `null`       | `true` |  `false` | Query parameter value of 'X' is not an array
| `\d+` | `42`                |      `null`       | `false` |  `false` | `Array(42)`
| `\d+` | `42`                |      `test`       | `false` |  `true` | `Array(42)`
| `\d+` | `42`                |      `test`       | `true` |  `true` | Query parameter X value 'test' violated a constraint
| `\d+` | `test`              |      `null`       | `true` |  `true` | `Array('test')`
| `\d+` | `null`              |      `test`       | `true` |  `false` | Query parameter X value 'test' violated a constraint
| `\d+` | `null`              |      `Array(1)`       | `false` |  `false` | `Array(1)`
| `\d+` | `null`              |      `Array(1)`       | `true` |  `false` | `Array(1)`
| `\d+` | `null`              |      `Array(1)`       | `true` |  `true` | `Array(1)`

lsmith77:
OK thank you .. maybe something to add to the docs.

eliecharra:
@lsmith77 Yeah I think so, maybe we need to add some other tests, not sure we cover all these cases.
It's quite complicated to reverse the behavior from the code there :wink:

lsmith77:
lets have this sit in master for a bit ..

lsmith77:
I just released 1.5.1 without this fix .. but I also now merged it into the 1.5 branch .. still waiting to see if someone reports issues for a few days since I am worried about causing another regression.

lsmith77:
released as https://github.com/FriendsOfSymfony/FOSRestBundle/releases/tag/1.5.2

PR #966, opened 2015-02-08
[bugfix] Forward requirements and defaults in RestYamlCollectionLoader

vierbergenlars:
This pull request adds tests that the collection loaders (both XML and YAML) forward route requirements and defaults when importing a controller, like the forwarding of options already works.
This mirrors the behavior of the [symfony `YamlFileLoader`](https://github.com/symfony/symfony/blob/ac389b67145e26f99a59a06bad7346ab5e2ec097/src/Symfony/Component/Routing/Loader/YamlFileLoader.php#L172-173) and [`XmlFileLoader`](https://github.com/symfony/symfony/blob/ac389b67145e26f99a59a06bad7346ab5e2ec097/src/Symfony/Component/Routing/Loader/XmlFileLoader.php#L181-182)

Adding these tests uncovered a flaw in `RestYamlCollectionLoader`, which does not forward these requirements or defaults. The `RestXmlCollectionLoader` does not contain this flaw, because it calls the symfony `XmlFileLoader`.

### DIFF
--- a/Routing/Loader/RestYamlCollectionLoader.php
+++ b/Routing/Loader/RestYamlCollectionLoader.php
@@ -68,13 +68,15 @@ class RestYamlCollectionLoader extends YamlFileLoader
         // process routes and imports
         foreach ($config as $name => $config) {
             if (isset($config['resource'])) {
-                $resource   = $config['resource'];
-                $prefix     = isset($config['prefix'])      ? $config['prefix']         : null;
-                $namePrefix = isset($config['name_prefix']) ? $config['name_prefix']    : null;
-                $parent     = isset($config['parent'])      ? $config['parent']         : null;
-                $type       = isset($config['type'])        ? $config['type']           : null;
-                $options    = isset($config['options'])     ? $config['options']        : null;
-                $currentDir = dirname($path);
+                $resource     = $config['resource'];
+                $prefix       = isset($config['prefix'])       ? $config['prefix']         : null;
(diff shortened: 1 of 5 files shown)


Routing/Loader/RestYamlCollectionLoader.php
```diff
@@ -94,9 +96,9 @@ public function load($file, $type = null)
                     $this->collectionParents[$name] = $parents;
                 }
 
-                if ($options) {
```
lsmith77: hmm .. why did we have this check here before? just to prevent a method call?

lsmith77:
thx ... I guess this could also go into 1.5.1

Routing/Loader/RestYamlCollectionLoader.php
vierbergenlars: I guess so. `addOptions()` does nothing spectacular, and that method already checks if `$options` is empty by [itself](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/Routing/RouteCollection.php#L235) before iterating over its routes.

vierbergenlars:
As I see it, it's a bugfix, not a feature addition.
I cannot imagine someone relies on `requirements` and `defaults` configuration not being passed to the routes in the collection, especially since the XML loader does the right thing, and so does the symfony RouteCollection.

lsmith77:
released in 1.5.1

PR #968, opened 2015-02-11
Added useful configSerializationContext() in View class

Strate:
This PR adds useful configurator for serialization context in view class.
Before:
```php
return $this->handleView(
    $this->view($data)->setSerializationContext(
        SerializationContext::create()->setGroups('123')
    )
)
```

After:
```php
return $this->handleView(
    $this->view($data)->configSerializationContext(function(SerializationContext $context) {
        $context->setGroups('123')
    })
)
```

### DIFF
--- a/View/View.php
+++ b/View/View.php
@@ -459,4 +459,22 @@ class View
 
         return $this->serializationContext;
     }
+
+    /**
+     * Useful method for configure serialization context without losing fluent interface.
+     * Usage:
+     *      $view->configSerializationContext(function(SerializationContext $context) {
+     *          $context->setGroups('group1');
+     *      })
+     *
+     * @param callable $configurator
+     *
+     * @return View
+     */
+    public function configSerializationContext(\Closure $configurator)
+    {
+        $configurator($this->getSerializationContext());
+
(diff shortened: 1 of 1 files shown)


lsmith77:
I do not see the advantage of this approach. It would for example make sense if we would then lazily (f.e. for the case when templating is used rather than serialization) do some expensive method calls but the calls to the context class are neither expensive, nor will they be done lazily

Strate:
This is more for passing View object through control flow. For example you need to chnage serialization context config, which comes in `View` class, passed from other method:
```php
public function someAction()
{
    $view = $this->fetchViewObject();
    // Now, you need to set serialization group in serialization context
    $view->getSerializationContext()->setGroups('123');
    return $viewl
}
```
another way to do this with new method:
```php
public function someAction()
{
    return $this->fetchViewObject()
        ->configureSerializationContext(function($context) {
            $context->setGroups('123')
        });
}
```

lsmith77:
hmm I am not really convinced that this is useful enough so that it makes sense to add this. so I would prefer if you would just extend the View class in your code to add this.

Issue #470, opened 2013-06-06
Variables with type definition do not longer work in action function


Jnig:
Hi,

with the latest dev-master it's no longer possible to define variables with type in the action function.

Controller function:

```
/**
* @Rest\View
*/
public function getUsersAction(User $user) {

}
```

In version 0.12 I get the following route:
  get_users                             GET      ANY    ANY  /api/users/{user}.{_format}

With dev-master I get:
  get_users                             GET      ANY    ANY  /api/users.{_format}

PR #969, opened 2015-02-11
Add jms/serializer to required packages

ianfp:
`FOS\RestBundle\Serializer\ExceptionWrapperSerializeHandler`
implements `JMS\Serializer\Handler\SubscribingHandlerInterface`
and throws the following error if `jms/serializer` is not installed:

PHP Fatal error:  Interface 'JMS\\Serializer\\Handler\\SubscribingHandlerInterface' not found in vendor/friendsofsymfony/rest-bundle/FOS/RestBundle/Serializer/ExceptionWrapperSerializeHandler.php on line 13

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,8 @@
         "symfony/framework-bundle":       "~2.3",
         "doctrine/inflector":             "~1.0",
         "willdurand/negotiation":         "~1.2",
-        "willdurand/jsonp-callback-validator": "~1.0"
+        "willdurand/jsonp-callback-validator": "~1.0",
+        "jms/serializer":                 "~0.12"
     },
 
     "require-dev": {


mvrhov:
It's not there by default because the bundle also works with default symfony serializer.
You should search old issues and read the docs, where this is documented.

lsmith77:
indeed .. also in newer versions of Symfony, there is an alternative which this bundle can also use

ianfp:
Then why am I seeing the error above? How can `FOS\RestBundle\Serializer\ExceptionWrapperSerializeHandler` work if `jms/serializer` is not installed, since it implements an interface defined in `jms/serializer`?

lsmith77:
good point .. we also need to then find a solution for the case when JMS serializer is not available, ie. one that works with the core serializer. however we will not add a dependency on JMS serializer.

lsmith77:
@dunglas are you motivated to look into this? essentially we created a wrapper for exceptions so that we can better control the output structure for the JMS serializer. now we need an equivalent alternative for the core serializer.

dunglas:
@lsmith77 why not. It would be cool if the Symfony Serializer become a first-class citizen in FOSRest but too busy to work on that right now.

If someone is interested to work on that, this error normalizer can be good example of how using new features of the serializer: https://github.com/dunglas/DunglasJsonLdApiBundle/blob/master/Serializer/ConstraintViolationListNormalizer.php

xabbuh:
Sorry, I didn't see your replies when submitting #470. However, feel free to merge it as a quick fix for the next 1.5 patch release. A similar handler for the Symfony serializer could still be added afterwards.

lsmith77:
closing in favor of https://github.com/FriendsOfSymfony/FOSRestBundle/pull/970

lsmith77:
fixed in https://github.com/FriendsOfSymfony/FOSRestBundle/releases/tag/1.5.3

PR #970, opened 2015-02-12
remove unusable handler when JMS serializer is not present

xabbuh:
This compiler pass removes the `ExceptionWrapperSerializeHandler` if the JMS serializer is not present.

### DIFF
--- a/DependencyInjection/Compiler/ExceptionWrapperHandlerPass.php
+++ b/DependencyInjection/Compiler/ExceptionWrapperHandlerPass.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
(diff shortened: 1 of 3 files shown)


lsmith77:
I guess this will just mean that the return format will be different without JMS serializer. I guess this should be noted in the docs.

lsmith77:
and yeah .. I will then cherry pick this into 1.5

xabbuh:
@lsmith77 Are there any docs explaining the behaviour when the JMS serializer is available which could be changed accordingly?

lsmith77:
i was refering to just the docs on the exception controller. there it should be noted that the documented structure will be different without JMS serializer

lsmith77:
ie. I am talking about https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Resources/doc/4-exception-controller-support.rst

though the response format is there currently not even mentioned. but imho at the very least we should then mention that the format will be different depending on if one uses JMS serializer or not ..

lsmith77:
tweaked the docs in 40e595a

xabbuh:
Thanks for the tweaks @lsmith77 and sorry for not responding soon enough.

Issue #971, opened 2015-02-12
create an Exception handler for the core serializer 


lsmith77:
see https://github.com/FriendsOfSymfony/FOSRestBundle/pull/969

PR #975, opened 2015-02-27
Exception serialization with core serializer

florianv:
This PR adds support for exception and form errors serialization with Symfony's serializer. It fixes issues https://github.com/FriendsOfSymfony/FOSRestBundle/issues/971.

- merged the `ExceptionWrapperHandlerPass` with the `SerializerConfigurationPass`
- added some functional tests following Symfony's tests structure for core bundles


lsmith77:
very nice!

DependencyInjection/Compiler/SerializerConfigurationPass.php
```diff
@@ -16,6 +16,9 @@
 
 /**
  * Checks if a serializer is either set or can be auto-configured.
+ *
+ * @author Christian Flothmann <christian.flothmann@xabbuh.de>
```
xabbuh: I think this is wrong. I didn't contribute anything to this class if I'm not mistaken. :)

Resources/config/view.xml
```diff
@@ -12,6 +12,7 @@
         <parameter key="fos_rest.view_handler.default.class">FOS\RestBundle\View\ViewHandler</parameter>
         <parameter key="fos_rest.view_handler.jsonp.class">FOS\RestBundle\View\JsonpHandler</parameter>
         <parameter key="fos_rest.serializer.exception_wrapper_serialize_handler.class">FOS\RestBundle\Serializer\ExceptionWrapperSerializeHandler</parameter>
+        <parameter key="fos_rest.serializer.exception_wrapper_normalizer.class">FOS\RestBundle\Serializer\ExceptionWrapperNormalizer</parameter>
```
xabbuh: I think we should remove this parameter. There are better ways for third-parties to overwrite/modify service definitions. Even Symfony itself doesn't introduce class name parameters for new services anymore (removing the parameter wouldn't be a BC break right now since there was no release since the PR has been merged).

DependencyInjection/Compiler/SerializerConfigurationPass.php
florianv: Well, I moved your code from the other pass so I left your copyright

PR #978, opened 2015-02-27
Build against php nightly

florianv:


lsmith77:
was pondering if we should make it an allowed failure .. but as things are green for now it would be good to detect any regression

Issue #976, opened 2015-02-27
Array Normalizer doesn't work for multipart/form-data


AlexStansfield:
Just ran into this problem when testing some code. Works fine with json and x-www-form-urlencoded.

The problem appears to be in this line:
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/EventListener/BodyListener.php#L90

It checks the content type matches multipart/form-data. However any multipart/form-data content type string will also contain a boundary value. 

Perhaps the content-type needs to be exploded on ';' and then the first value matched. I have also noticed some clients adding a charset onto a application/x-www-form-urlencoded content type so it would solve that problem too.

lsmith77:
sounds reasonable .. can you work on a PR?

AlexStansfield:
Sorry only just caught this reply and see @florianv beat me to it.

PR #979, opened 2015-02-27
Fixed array normalization for form post

florianv:
Fix for issue https://github.com/FriendsOfSymfony/FOSRestBundle/issues/976

### DIFF
--- a/EventListener/BodyListener.php
+++ b/EventListener/BodyListener.php
@@ -15,6 +15,7 @@ use FOS\RestBundle\Decoder\DecoderProviderInterface;
 use FOS\RestBundle\Normalizer\ArrayNormalizerInterface;
 use FOS\RestBundle\Normalizer\Exception\NormalizationException;
 use Symfony\Component\HttpFoundation\ParameterBag;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Event\GetResponseEvent;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 use Symfony\Component\HttpKernel\Exception\UnsupportedMediaTypeHttpException;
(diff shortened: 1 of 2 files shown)


lsmith77:
pondering if the code in RequestUtil could be put some where more sensible .. was looking at @willdurand's Negotiation lib as one possible place ..

florianv:
I agree that it could be located somewhere where it's more reusable.

In Symfony 2.6, the `form` request format was introduced but it seems to lack these cases as well (https://github.com/symfony/symfony/pull/7142).

Maybe it would be better to contribute it to Symfony because there is no negotiation logic (it's just basic parsing) so I'm not sure if it should go in Negotiation.

lsmith77:
maybe we can at least add a comment that this util class may be removed in a future release?

lsmith77:
or lets just make it a private method in the listener class?

florianv:
@lsmith77 A private method in the listener will be annoying to test. What about adding an @internal annotation to the class ?

lsmith77:
fine for me

Util/RequestUtil.php
```diff
@@ -0,0 +1,53 @@
...
+    /**
+     * Checks if the request is a form post.
+     *
+     * @param Request $request
+     *
+     * @return boolean
+     */
+    public static function isFormPost(Request $request)
+    {
+        if (!$request->isMethod('POST')) {
```
stof: this check looks wrong to me. What about PUT for instance ?

florianv: If the request is PUT, it will enter the `if` and the normalization will happen (https://github.com/florianv/FOSRestBundle/blob/array-normalizer/EventListener/BodyListener.php#L119). 

When the feature was initially merged, we only normalized requests with a decodable format (non forms). Support for forms was added later through the `forms` option to keep BC.

Maybe this method should be private in the listener as @lsmith77 suggested because it's not really reusable indeed

florianv: Indeed these cases are missing. I will move this code to a private listener method as @lsmith77 suggested and add extra tests for this

florianv:
Ready for review. As per @stof comment a bug has also been fixed regarding form data normalization for `PUT`, `PATCH` and `DELETE` requests.

PR #981, opened 2015-02-28
Removed useless ini_set in tests

florianv:
These lines were copied by mistake from the above tests.

They are needed to hide the logged exception messages when running the tests but useless here because no exception is thrown when submitting the invalid form.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ php:
   - 5.6
   - nightly
   - hhvm
+  - hhvm-nightly
 
 branches:
   only:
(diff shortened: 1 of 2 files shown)


lsmith77:
oddly enough the nightly builds are now failing ..

florianv:
I have no clue why these tests are failing. Should we allow failures for the nightly version ?

lsmith77:
seems like yes.

florianv:
Done. I also added hhvm-nightly to the build matrix

Issue #984, opened 2015-03-02
Form entity field submit (PUT/POST) as array


backender:
Hi guys,

Not sure if this issue is duplicate but I'm facing the same problem as described in the following stackoverflow question: http://stackoverflow.com/questions/19553884/symfony-forms-fos-rest

Is there a proper way to directly use the from a GET received {"array":{"id":1}} in a PUT/POST that is related to a form field type of "entity" with property id – instead of the {array:1}?

In fact, is writing my own EntityToArray transformer the way to go here, or is already something introduced by this bundle that works out of the box?

lsmith77:
I do not think that we have anything for this in the Bundle. Of course you can customize the response structure via the serializer but it seems it would be more elegant to handle this on the server side, so your approach sounds sensible. If you can find a generic enough solution, we can maybe put it into this Bundle

/cc @webmozart

webmozart:
I agree, writing a transformer sounds like a good solution to me. :)

lsmith77:
closing this .. unless you want to keep it open until a solution is found for entities with composite primary keys?

backender:
The case with the composite primary key seems rather rare. So to me it's ok.

PR #987, opened 2015-03-06
DataTransformer to transform serialized objects containing an id to it's entity

backender:
According to the issue (https://github.com/FriendsOfSymfony/FOSRestBundle/issues/984) I created a transformer that can handle serialized objects containing an id and map it to the entity.

The use of this transfomer within a form type would look as follows:
```php
$transformer = new EntityToIdObjectTransformer($this->om, "AcmeDemoBundle:MyEntity");
$builder->add($builder->create('myfield', 'text')->addModelTransformer($transformer))
```
This way, the selialization process stays untouched.

Not sure if this it's generic enough, otherwise it's documented here at least.

### DIFF
--- a/Form/Transformer/EntityToIdObjectTransformer.php
+++ b/Form/Transformer/EntityToIdObjectTransformer.php
@@ -34,6 +34,7 @@ class EntityToIdObjectTransformer implements DataTransformerInterface {
 
     /**
      * @param ObjectManager $om
+     * @param String $entityName
      */
     public function __construct(ObjectManager $om, $entityName)
     {
--- a/Form/Transformer/EntityToIdObjectTransformer.php
+++ b/Form/Transformer/EntityToIdObjectTransformer.php
@@ -54,11 +54,11 @@ class EntityToIdObjectTransformer implements DataTransformerInterface {
             return "";
         }
 
-        return $object->getId();
+        return array_values($this->om->getClassMetadata($this->entityName)->getIdentifierValues($object))[0];
     }
 
     /**
-     * Transforms an array including an id to an object.
+     * Transforms an array including an identifier to an object.
      *
      * @param  array $idObject
      *
@@ -68,21 +68,22 @@ class EntityToIdObjectTransformer implements DataTransformerInterface {
      */
     public function reverseTransform($idObject)
     {
-        if (!is_array($idObject) || !array_key_exists("id", $idObject) || !$idObject['id']) {
+        if (!is_array($idObject)) {
             return null;
         }
 
-        $id = $idObject['id'];
+        $identifier = array_values($this->om->getClassMetadata($this->entityName)->getIdentifier())[0];
+        $id = $idObject[$identifier];
 
         $object = $this->om
             ->getRepository($this->entityName)
-            ->findOneBy(array('id' => $id))
+            ->findOneBy(array($identifier => $id))
         ;
 
         if (null === $object) {
             throw new TransformationFailedException(sprintf(
-                'An object with id "%s" does not exist!',
-                $id
+                'An object with identifier key "%s" and value "%d" does not exist!',
+                $identifier, $id
             ));
         }
 
--- a/Form/Transformer/EntityToIdObjectTransformer.php
+++ b/Form/Transformer/EntityToIdObjectTransformer.php
@@ -1,0 +1,90 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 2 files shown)


lsmith77:
can you add something to the docs and ideally also a test?

Form/Transformer/EntityToIdObjectTransformer.php
```diff
@@ -0,0 +1,90 @@
...
+     *
+     * @param  array $idObject
+     *
+     * @return Object|null
+     *
+     * @throws TransformationFailedException if object is not found.
+     */
+    public function reverseTransform($idObject)
+    {
+        if (!$idObject || !$idObject['id']) {
```
xabbuh: You should make sure that `$idObject` is an array.

Form/Transformer/EntityToIdObjectTransformer.php
```diff
@@ -0,0 +1,90 @@
...
+        if (null === $object) {
+            throw new TransformationFailedException(sprintf(
+                'An object with id "%s" does not exist!',
+                $id
+            ));
+        }
+
+        return $object;
+    }
+}
```
xabbuh: missing newline

Form/Transformer/EntityToIdObjectTransformer.php
```diff
@@ -0,0 +1,90 @@
...
+     * @param  Object|null $object
+     * @return Object
+     */
+    public function transform($object)
+    {
+        if (null === $object) {
+            return "";
+        }
+
+        return $object->getId();
```
xabbuh: Does this make sense? Shouldn't it be `return array($object->getId())`? Does it make sense to check if the object actually has a `getId()` method first?

Form/Transformer/EntityToIdObjectTransformer.php
```diff
@@ -0,0 +1,90 @@
...
+    {
+        $this->entityName = $entityName;
+        $this->om = $om;
+    }
+
+    /**
+     * Do nothing.
+     *
+     * @param  Object|null $object
+     * @return Object
```
xabbuh: should be `string`

Form/Transformer/EntityToIdObjectTransformer.php
```diff
@@ -0,0 +1,90 @@
...
+    private $om;
+
+    /**
+     * @var String
+     */
+    private $entityName;
+
+    /**
+     * @param ObjectManager $om
+     */
```
xabbuh: The docblock should be either complete (it's currently missing the `$entityName` parameter) or should be omitted otherwise.

Form/Transformer/EntityToIdObjectTransformer.php
lsmith77: i guess either we use the ClassMetadata to figure out how to get he id or we need to maybe make this configurable or a clear extension point

lsmith77:
ping

backender:
last week was busy. try to provide docs this week.

Form/Transformer/EntityToIdObjectTransformer.php
backender: My first intention was to provide a constructor argument that holds a default value. This way, also reverseTransform can be configured in the same place. Obviously ClassMetadata would be an option too. I'll investigate here a bit further this week.

lsmith77:
@backender not sure either about how the form layer expects composite keys to be received.

Form/Transformer/EntityToIdObjectTransformer.php
```diff
@@ -0,0 +1,92 @@
...
+     * @param  Object|null $object
+     * @return string
+     */
+    public function transform($object)
+    {
+        if (null === $object) {
+            return "";
+        }
+
+        return array_values($this->om->getClassMetadata($this->entityName)->getIdentifierValues($object))[0];
```
Baachi: This isn't php 5.3 compatible.

Form/Transformer/EntityToIdObjectTransformer.php
```diff
@@ -0,0 +1,92 @@
...
+     *
+     * @throws TransformationFailedException if object is not found.
+     */
+    public function reverseTransform($idObject)
+    {
+        if (!is_array($idObject)) {
+            return null;
+        }
+
+        $identifier = array_values($this->om->getClassMetadata($this->entityName)->getIdentifier())[0];
```
Baachi: Same here

Form/Transformer/EntityToIdObjectTransformer.php
lsmith77: good catch! @backender can you have a look and submit a new PR?

Baachi: @lsmith77 I created a PR #1001

backender: +1

PR #989, opened 2015-03-11
fix links

TorbenBr:


stof:
I suggest linking to the rendered doc on symfony.com instead (the rST rendering on Github is incomplete as it does not support all Sphinx features)

lsmith77:
+1 on @stof's comment

Issue #822, opened 2014-07-24
serializerGroups annotation not behaving as expected


robcaw:
I'm trying to set up exclusion policies for my entities using yaml for the properties, and setting the serializerGroups on my view actions.
```
/**
     * @Rest\View(
     *      serializerGroups={"api_message"}
     * )
     * @param Request $request
     * @return \FOS\RestBundle\View\View
*/
```
However, the annotation serializerGroups, although recognised, doesn't set the groups for the serializer. Instead, the View's  SerializationContext must be set:

`$view->setSerializationContext(SerializationContext::create()->setGroups(['api_message']));`

robertblock:
Same problem here.

I have not been able to upgrade from FOSRestBundle 1.2.* since the @Rest\View annotation serializerGroups action does not work after upgrading to 1.3.*, 1.4.* or 1.5.*.

I have no doubt it's my fault, I'm just not sure what I'm doing wrong.

lsmith77:
can you paste your controller code?

robertblock:
Sure. Let me know if this is enough:
```
use Doctrine\DBAL\DBALException;
use JMS\Serializer\SerializationContext;
use Nelmio\ApiDocBundle\Annotation\ApiDoc;
use FOS\RestBundle\View\View AS FOSView;
use Symfony\Bundle\FrameworkBundle\Controller\Controller;
use Symfony\Component\HttpFoundation\Request;
use FOS\RestBundle\Controller\Annotations as Rest;
use JMS\SecurityExtraBundle\Annotation as Security;

class StudentRestController extends Controller
{
    /**
     * Get student by id.
     *
     * @Rest\View(serializerGroups={"sg-default", "sg-details"})
     *
     * @ApiDoc()
     *
     * @param Student $student
     * @return FOSView
     *
     * @Security\Secure("ROLE_USER")
     */
    public function getStudentAction(Student $student)
    {
        return FOSView::create()->setStatusCode(200)->setData($student);
    }
}
```

robertblock:
Without looking too deeply into the code, it looks like the inclusion of `$customViewDefined` in `ViewResponseListener` is causing this issue. 

Again, without looking too deeply into the code, there may be a logic error on line 89. I think serializer groups should be applied if `$customViewDefined == true`:

```
if ($configuration->getSerializerGroups() && !$customViewDefined) {
```

should be:

```
if ($configuration->getSerializerGroups() && $customViewDefined) {
``` 

[...]

PR #997, opened 2015-03-23
[Issue 822] Removed logic error in checking for custom view before serializing groups

robertblock:
There was a check for a custom view (`$customViewDefined`) added in `1.3.0` in `ViewResponseListener`. The following was the conditional for whether or not to serialize groups:
```
if ($configuration->getSerializerGroups() && !$customViewDefined)
```  
I simply removed the `!` to fix the logic:
```
if ($configuration->getSerializerGroups() && $customViewDefined)
```


lsmith77:
fix for https://github.com/FriendsOfSymfony/FOSRestBundle/issues/822

lsmith77:
see my comment in #822

PR #1001, opened 2015-04-01
Fix PHP 5.3 compatibility in EntityToObjectTransformer

Baachi:


lsmith77:
I guess the Doctrine API guarantees an array return value for us here or do we need to check for null etc?

Baachi:
AFAIK the method `getIdentifierValues` return always an array. If it's not the case, you misconfigured your entity anyways.

PR #1002, opened 2015-04-02
fix BodyListener to skip Soap Request parsing

INSEAD-asim:
Hi FOS Team,

I implemented this fix to exclude parsing for SOAP 1.1 and SOAP 1.2 requests. It would be good if you integrate in FOS RestBundle.

Regards,
Asim

### DIFF
--- a/EventListener/BodyListener.php
+++ b/EventListener/BodyListener.php
@@ -163,9 +163,7 @@ class BodyListener
 
     private function isSoapRequest(Request $request)
     {
-        $soapAction = $request->headers->get('SoapAction');
-
-        $isSoap11 = $soapAction !== '';
+        $isSoap11 = $request->headers->has('SoapAction');
         $isSoap12 = false;
 
         $contentTypeParts = explode(';', $request->headers->get('Content-Type'));
--- a/EventListener/BodyListener.php
+++ b/EventListener/BodyListener.php
@@ -89,7 +89,12 @@ class BodyListener
         $method = $request->getMethod();
         $contentType = $request->headers->get('Content-Type');
         $isFormRequest = $this->isFormRequest($request);
+        $isSoapRequest = $this->isSoapRequest($request);
         $normalizeRequest = $this->normalizeForms && $isFormRequest;
+
+        if ($isSoapRequest) {
+            return;
+        }
 
         if (!$isFormRequest && in_array($method, array('POST', 'PUT', 'PATCH', 'DELETE'))) {
             $format = null === $contentType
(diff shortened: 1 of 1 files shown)


lsmith77:
that feels a bit hacky. maybe a black list of mime types makes more sense?

INSEAD-asim:
I found that issue is only with SOAP 1.1 as it has same content type i.e. application/xml. The SOAP 1.2 is still okay as content type is application/soap+xml. Not sure if we can use mime type as SOAP is also xml. Here is the link with more details on SOAP headers http://www.herongyang.com/Web-Services/Perl-SOAP-1-2-Request-Differences-SOAP-1-1-and-1-2.html

lsmith77:
and so for SOAP 1.1 you need that hack to check if there is a SoapAction header to determine if not to decode. not really sure if I want to add this into the listener itself. that being said, we support overriding the service to use for the body listener. so maybe the best would be to have an extension point to add custom logic to determine if to decode or not ..

INSEAD-asim:
I tried to override XML Decoder but it always asked to return array(). The modification is to just bypass the body listener in case of SOAP request. The SOAP 1.1 don't have anything specific except SoapAction. Happy to implement something more beautiful.

lsmith77:
no I meant overriding the body listener .. not the xml decoder

INSEAD-asim:
OK. Let me check again on my side. Thanks for tip.

INSEAD-asim:
Can you hint how to override body listener? Since it is not available by default in configuration for overriding.

Also, what I feel that it will create lot of extra work on developer side to override and manage body listener with default body listener provided by FOSRestBundle. This fix will help to overcome this situation. Since we are using this bundle as a part of application not full application.

lsmith77:
hmm .. indeed you are right .. we do not make it easy to override the listener classes via Bundle configuration. just not sure how to best handle this.

lsmith77:
I think the best course of action is the following:
1) add a new protected method that default to return true, that if it returns false will stop any decoding
2) make it possible to configure a different service for the body listener (actually maybe we should allow this for any listener)

what do you think about this approach? then for example you can easily filter by path, header or whatever to determine where the body listener should not lead to decoding anything.

lsmith77:
see also https://twitter.com/lsmith/status/586098409515417600

INSEAD-asim:
Your idea seems very good. I will try to work on this.

Meanwhile, I push one fix for same thing. It stopped body_listener with the previous fix. Check it out at https://github.com/INSEAD-asim/FOSRestBundle/commit/16a316364a93338c13fdeeadcda6f820ff345ad8

stof:
@lsmith77 do we really need to make listener services configurable ? If you want your own listener, register your own listener and disable the listener from the bundle.

lsmith77:
well it makes sense as we provide various config options for this listener .. that being said, at least we should provide some sensible extension points so people can easily extend the default listener

lsmith77:
I have created an alternative approach to this that adds the extension point and makes the service configurable.

PR #1003, opened 2015-04-06
Add versionadded directive for setTemplateData

xelaris:
See also https://github.com/symfony/symfony-docs/issues/5151#issuecomment-89974092.

### DIFF
--- a/Resources/doc/2-the-view-layer.rst
+++ b/Resources/doc/2-the-view-layer.rst
@@ -52,6 +52,9 @@ which adds several convenience methods:
         }
     }
 
+.. versionadded:: 1.6
+  The ``setTemplateData`` method was added in 1.6.
+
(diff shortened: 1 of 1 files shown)


lsmith77:
@schmittjoh this is what I mean with scrutinizer failing.

schmittjoh:
Yeah, this is because PHP7 does not yet have the xdebug extension, and cannot generated code coverage. You see this in the nightly job here: https://travis-ci.org/FriendsOfSymfony/FOSRestBundle/jobs/57321300#L272

The only solution at the moment is to not try to upload code coverage if a job cannot generate it.

lsmith77:
@schmittjoh oh .. didn't realize someone has added php7 already to the builds. is there a work around to prevent breaking CI just because we have php7?

xabbuh:
@lsmith77 you can have a look at #1005 (it's already merged, just in case you need it somewhere else)

PR #1005, opened 2015-04-07
don't send code coverage reports on PHP 7 builds

xabbuh:
Currently, PHP 7 doesn't come with the Xdebug extension and thus cannot
generate code coverage reports. Hence, sending them to Scrutinizer CI
must be disabled on such builds.


stof:
thanks

Issue #973, opened 2015-02-20
[BUG] Update from 1.4.2 to 1.5.3 illegal offset type in GenericMetadata.php


entering:
```
Warning: Illegal offset type in isset or empty in /vagrant/vendor/symfony/symfony/src/Symfony/Component/Validator/Mapping/GenericMetadata.php line 216
500 Internal Server Error - ContextErrorException
```

symfony version: v2.5.3

The code:

```php
    /**
     * @QueryParam(
     *      name="letter",
     *      nullable=false,
     *      requirements="[A-Z\#]",
     *      default="A",
     *      strict=true,
     *      description="First letter of the Accommodation Provider's name to filter by"
     * )
     *
     * @param Request $request
     * @param string $letter
     *
     * @return Response
     */
    public function listAction(Request $request, $letter)
{
...
}
```

And the routing:
```yml
accommodation_provider_list:
    pattern:  /accommodation-providers
    defaults: { _controller: accommodation_bundle.accommodation_provider.controller:listAction }
    methods: [POST, GET]
```

lsmith77:
can you investigate this a but or submit a test a failing test case? there has been some refactoring in this area

entering:
@lsmith77 yes, I just opened the issue to avoid getting "lost", I will investigate and come at least with a failing test case

lsmith77:
ping

lsmith77:
did you also open a ticket on Symfony core?

/cc @webmozart

mlimaloureiro:
hello, i was investigating a little bit this issue, and the problem is with symfony LegacyValidator. Somehow, it is expecting an object instead of a value. Anyway this is fixed in 2.5.10, didn't investigate any further since updating symfony fixed the problem

lsmith77:
maybe it was fixed by https://github.com/symfony/Validator/commit/b90adb60265bec674a0a5ac12bff95d93f7d78f1

I guess then we could in theory add a conflict to lower versions. reading the commit info it would then be an issue with 2.5 versions up to exclusing 2.5.5. it never affected versions below 2.5 of course and also didn't affect 2.6 versions.

PR #1007, opened 2015-04-10
due to an issue in LegacyValidator we disallow early 2.5.x releases

lsmith77:
fix for https://github.com/FriendsOfSymfony/FOSRestBundle/issues/973

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,13 +41,14 @@
     },
 
     "suggest": {
-        "sensio/framework-extra-bundle":  "Add support for route annotations and the view response listener",
+        "sensio/framework-extra-bundle":  "Add support for route annotations and the view response listener, requires ~3.0",
         "jms/serializer-bundle":          "Add support for advanced serialization capabilities, recommended, requires ~0.12",
-        "symfony/serializer":             "Add support for basic serialization capabilities and xml decoding, requires ~2.2",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -41,13 +41,14 @@
...
+        "sensio/framework-extra-bundle":  "Add support for route annotations and the view response listener, requires ~3.0",
         "jms/serializer-bundle":          "Add support for advanced serialization capabilities, recommended, requires ~0.12",
-        "symfony/serializer":             "Add support for basic serialization capabilities and xml decoding, requires ~2.2",
-        "symfony/validator":              "Add support for validation capabilities in the ParamFetcher, requires ~2.2"
+        "symfony/serializer":             "Add support for basic serialization capabilities and xml decoding, requires ~2.3",
+        "symfony/validator":              "Add support for validation capabilities in the ParamFetcher, requires ~2.3"
     },
 
     "conflict": {
+        "symfony/framework-bundle": ">=2.5,<2.5.5",
```
stof: what is forbidden is older versions of Validator, not of FrameworkBundle

stof:
@lsmith77 why merging it without fixing it ?

composer.json
lsmith77: ah sorry. overlooked the comment. yeah you are right

PR #1011, opened 2015-04-12
use non-deprecated code when possible

xabbuh:


lsmith77:
retriggered the travis build

stof:
We should remove the usages of deprecated classes

xabbuh:
For the moment, I added some tests to the `legacy` group.

Tests/Functional/SerializerErrorTest.php
```diff
@@ -87,6 +88,7 @@ public function serializeInvalidFormJsonProvider()
 
     /**
      * @dataProvider serializeInvalidFormXmlProvider
+     * @group legacy
```
stof: this should not be added in the legacy group. ``isBound`` was already deprecated in favor of ``isSubmitted`` in Symfony 2.3, so we should just update the code instead

Tests/Request/RequestBodyParamConverterTest.php
```diff
@@ -205,6 +205,9 @@ public function testApplyWithSerializerContextOptionsForSymfonySerializer()
         $this->converter->apply($request, $config);
     }
 
+    /**
+     * @group legacy
```
stof: this does not make sense either. The codebase already sypports using the new validator. You should just detect the availability of the new validator interface and then use it for the mock instead of the old one.

xabbuh:
Thought it would be much more work to replace deprecated code. Finished now.

lsmith77:
@stof good to merge?

Tests/Request/RequestBodyParamConverterTest.php
```diff
@@ -207,14 +207,33 @@ public function testApplyWithSerializerContextOptionsForSymfonySerializer()
...
     {
-        $validator = $this->getMockBuilder('Symfony\Component\Validator\Validator')
-            ->disableOriginalConstructor()
-            ->getMock();
+        $expectedPost = new Post('Post 1', 'This is a blog post');
         $validationErrors = $this->getMock('Symfony\Component\Validator\ConstraintViolationList');
 
+        if (class_exists('Symfony\Component\Validator\Validator\RecursiveValidator')) {
+            $validator = $this
+                ->getMockBuilder('Symfony\Component\Validator\Validator\RecursiveValidator')
```
stof: you should mock the interface, not the class

Tests/Request/RequestBodyParamConverterTest.php
```diff
@@ -207,14 +207,33 @@ public function testApplyWithSerializerContextOptionsForSymfonySerializer()
...
+                ->disableOriginalConstructor()
+                ->getMock();
+            $validator
+                ->expects($this->once())
+                ->method('validate')
+                ->with($expectedPost, null, array('group1'))
+                ->will($this->returnValue($validationErrors));
+        } else {
+            $validator = $this
+                ->getMockBuilder('Symfony\Component\Validator\Validator')
```
stof: same here

PR #1019, opened 2015-04-16
update Travis CI configuration

xabbuh:
* remove unneeded shell wrapper
* run a Travis build with lowest dependency versions
* allow to finish as fast as possible (when either a mandatory build
  failed or when only builds allowed to fail have not completed yet)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,10 @@ branches:
     - /^\d+\.\d+$/
 
 matrix:
+  fast_finish: true
   include:
+    - php: 5.3
+      env: deps="low"
     - php: 5.5
       env: SYMFONY_VERSION='2.3.* symfony/expression-language:2.4.*'
     - php: 5.5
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -9,13 +9,20 @@ php:
   - hhvm
   - hhvm-nightly
 
+env:
+  global:
+    - deps=""
```
stof: this is not even needed

.travis.yml
```diff
@@ -30,7 +37,8 @@ matrix:
 
 before_script:
   - composer self-update
-  - sh -c 'if [ "$SYMFONY_VERSION" != "" ]; then composer require --dev --no-update symfony/symfony=$SYMFONY_VERSION; fi;'
+  - if [ "$SYMFONY_VERSION" != "" ]; then composer require --dev --no-update symfony/symfony=$SYMFONY_VERSION; fi
+  - if [ "$deps" = "low" ]; then composer update --prefer-lowest; fi
   - composer update --prefer-source
```
stof: this should not be done on the deps=low, otherwise the deps won't be low anymore

lsmith77:
guess we need to increase the minimum jms serializer version .. good catch!

stof:
please update the min requirement to a working version

xabbuh:
I had some other issue after updating the serializer version for the tests. I think I can get back to this tomorrow and ping you after everything has been fixed.

xabbuh:
I made some updates to the test dependencies. However, there's still at least one failure. When running the test from `SerializerErrorTest`, the `ExceptionWrapperNormalizer` is not always executed as expected and thus doesn't change the output format. Does anybody of you know which dependency keeps the normalizer from being executed?

xabbuh:
Seems like the failure was related to the minimum version of the HttpKernel component. Though I didn't figure out what changed from 2.3.23 to 2.3.24 that was causing the failure to go away.

ping @lsmith77 @stof

PR #1025, opened 2015-04-30
Fix symfonyserializer groups

soyuka:
:warning: this is only working with symfony 2.7 serializer.

I haven't added any tests on this because I don't know what you'd like to see beeing tested. 
A lot of your current tests are working with the `JMS\Serializer`. Should every test beeing written again for the symfony serializer? 

This is a work in progress and as symfony 2.7 is in a beta stage I don't think it would be merged soon. 

Waiting for further instructions :).

External ref: https://github.com/symfony/symfony/issues/14495

### DIFF
--- a/View/ViewHandler.php
+++ b/View/ViewHandler.php
@@ -13,6 +13,7 @@ namespace FOS\RestBundle\View;
 
 use JMS\Serializer\SerializerInterface;
 use JMS\Serializer\SerializationContext;
+use Symfony\Component\Serializer\SerializerInterface as SymfonySerializerInterface;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpFoundation\Request;
@@ -448,6 +449,15 @@ class ViewHandler extends ContainerAware implements ConfigurableViewHandlerInter
             if ($serializer instanceof SerializerInterface) {
                 $context = $this->getSerializationContext($view);
                 $content = $serializer->serialize($data, $format, $context);
+            } elseif ($serializer instanceof SymfonySerializerInterface) {
+                $context = $this->getSerializationContext($view);
+
+                $newContext = array();
+
+                if(!$context->attributes->get('groups')->isEmpty()) 
+                    $newContext['groups'] = $context->attributes->get('groups')->get();
+
+                $content = $serializer->serialize($data, $format, $newContext);
             } else {
                 $content = $serializer->serialize($data, $format);
             }
--- a/View/ViewHandler.php
+++ b/View/ViewHandler.php
@@ -449,13 +449,16 @@ class ViewHandler extends ContainerAware implements ConfigurableViewHandlerInter
             if ($serializer instanceof SerializerInterface) {
                 $context = $this->getSerializationContext($view);
                 $content = $serializer->serialize($data, $format, $context);
-            } elseif ($serializer instanceof SymfonySerializerInterface) {
+            //       test for symfony 2.7 version 
+            } elseif (class_exists('\Symfony\Component\Serializer\Normalizer\ObjectNormalizer') && $serializer instanceof SymfonySerializerInterface) {
+                    
                 $context = $this->getSerializationContext($view);
 
                 $newContext = array();
 
-                if(!$context->attributes->get('groups')->isEmpty()) 
+                if (!$context->attributes->get('groups')->isEmpty()) {
                     $newContext['groups'] = $context->attributes->get('groups')->get();
+                }
 
                 $content = $serializer->serialize($data, $format, $newContext);
             } else {
--- a/composer.json
+++ b/composer.json
@@ -33,7 +33,7 @@
         "sensio/framework-extra-bundle":  "~3.0",
         "symfony/form":                   "~2.3",
         "symfony/validator":              "~2.3",
-        "symfony/serializer":             "~2.7",
+        "symfony/serializer":             "~2.3",
         "symfony/yaml":                   "~2.3",
         "symfony/security":               "~2.3",
         "symfony/browser-kit":            "~2.3",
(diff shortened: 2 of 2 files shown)


View/ViewHandler.php
```diff
@@ -448,6 +449,15 @@ private function initResponse(View $view, $format)
             if ($serializer instanceof SerializerInterface) {
                 $context = $this->getSerializationContext($view);
                 $content = $serializer->serialize($data, $format, $context);
+            } elseif ($serializer instanceof SymfonySerializerInterface) {
+                $context = $this->getSerializationContext($view);
+
+                $newContext = array();
+
+                if(!$context->attributes->get('groups')->isEmpty())
```
lsmith77: please follow the Symfony2 CS

lsmith77:
is there no way to determine if we have an instance of a 2.7 serializer available so that we do not need to bump the minimal version?

xabbuh:
For example, the [ObjectNormalizer](https://github.com/symfony/symfony/blob/2.7/src/Symfony/Component/Serializer/Normalizer/ObjectNormalizer.php) will be first included in Symfony 2.7.

soyuka:
Indeed, should we sait until it has been released?

xabbuh:
@soyuka If you check for presence of that class (or a similar one that is only present in Symfony 2.7), there will imho be no reason for waiting for a stable release.

lsmith77:
ping

View/ViewHandler.php
```diff
@@ -448,6 +449,18 @@ private function initResponse(View $view, $format)
...
+                    
+                $context = $this->getSerializationContext($view);
+
+                $newContext = array();
+
+                if (!$context->attributes->get('groups')->isEmpty()) {
+                    $newContext['groups'] = $context->attributes->get('groups')->get();
+                }
+
+                $content = $serializer->serialize($data, $format, $newContext);
```
lsmith77: why not just move all of this code into the else clause? it should not hurt to pass an extra parameter

lsmith77:
it would also be good to add a test case for this

View/ViewHandler.php
```diff
@@ -448,6 +449,18 @@ private function initResponse(View $view, $format)
             if ($serializer instanceof SerializerInterface) {
                 $context = $this->getSerializationContext($view);
                 $content = $serializer->serialize($data, $format, $context);
+            //       test for symfony 2.7 version 
+            } elseif (class_exists('\Symfony\Component\Serializer\Normalizer\ObjectNormalizer') && $serializer instanceof SymfonySerializerInterface) {
+                    
+                $context = $this->getSerializationContext($view);
```
stof: getting a JMSerializer context here means you have a dependency on JMSSerializer even when using the Symfony serializer. This is really bad (and it is a BC break)

soyuka: Thought this was related to `fosrest` directly.

stof: The right solution IMO would be to have a method ``getSymfonySerializationContext`` or something like that, returning the context array expected by Symfony (and it could be returned for any version btw, as Symfony will ignore unknown keys and this would allow people to provide other context options too)

lsmith77: puh .. this is getting a bit nasty. we then also need such a method on the ``View`` along with a ``setSymfonySerializationContext()`` method. very ugly but I see no way to not have to duplicate those methods without breaking BC in a big way.

soyuka: Can you confirm that every serialization tests on FOSRest are using the JMSSerializer? What about having those same tests with the Symfony Serializer, would we have to duplicate all those too?

lsmith77: yes .. that being said, maybe some of the tests actually cover too much of how JMS Serializer behaves instead of just ensuring that we pass the right information to the serializer. So it might not be necessary to duplicate all the tests. furthermore, for some tests it will just boil down to a different service injected into the ViewHandler, so we can have a base class for both serialziers.

stof: @lsmith77 the issue with the View API is that it currently exposes some methods dealing with the JMSSerializer class directly. This means that supporting any other serializer needs to provide another way to configure them.

IMO, the clean solution would be to provide a way to configure the serializer without dealing with the JMSSerializer classes directly (except maybe inside a callable passed by the user, which would receive the serializer and decide which logic to apply). this way, we could build an API which does not depend on JMSSerializer.

Ener-Getick: If I create a new context class that implements the jms serializer context methods, would it be ok for bc breaks ?

Ener-Getick:
Finally, I think that the best solution is to make an adapter that transform the jms context into a standard context class and after this standard context class is transformed for a specific serializer.

Like that, if you want for example to change your serializer for the symfony serializer, you will not have to change anything.

So for that, ``View::setSerializationContext`` must accept the jms normalization context and the standard normalization context.

lsmith77:
the problem is then, that users that do not have JMS serializer installed will need to add it to be able to use the core serializer. In order to maintain BC maybe the best option is to create our own ContextInterface, which implemements JMS's interface and if JMS isn't installed it defines it.

Due to difference in license it would be good to get @schmittjoh's permission for this.

Ener-Getick:
I have an other idea, sorry if I didn't good explain.

I created new context classes that do not implement the jms context interface :
https://github.com/Ener-Getick/FOSRestBundle/tree/TICKET_1025/Context

But to fix the BC breaks, the users can use the jms context or the new context in the view.
https://github.com/Ener-Getick/FOSRestBundle/blob/TICKET_1025/View/View.php#L194-209

To don't have to maintain two context classes, I also created some adapters that transform the context class to another format.
https://github.com/Ener-Getick/FOSRestBundle/tree/TICKET_1025/Context/Adapter

What do you think ?

Ener-Getick:
I just don't know if I should transform the JMS context directly into the view or after in the view handler ?

lsmith77:
@Ener-Getick can you create a PR .. makes it easier to review.

lsmith77:
for the record the PR was created https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1074

PR #1027, opened 2015-05-05
Replace http code by constant

skigun:


stof:
-1. the goal is precisely to show that both the code and the constant are accepted (the usage of the constant being documented on the following line

PR #1033, opened 2015-05-12
split out all longer listener sections into their own document

lsmith77:
can you look over this and tell me if it makes sense?

@WouterJ @weaverryan @javiereguiluz

### DIFF
--- a/Resources/doc/3-listener-support.rst
+++ b/Resources/doc/3-listener-support.rst
@@ -25,8 +25,8 @@ enable all listeners:
         view:
             view_response_listener: 'force'
 
-View Response listener
-----------------------
+3.1 View Response listener
+--------------------------
 
 The view response listener makes it possible to simply return a ``View``
 instance from action controllers. The final output will then automatically be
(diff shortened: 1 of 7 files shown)


Resources/doc/3-2-body_listener.rst
```diff
@@ -0,0 +1,88 @@
...
+        body_listener:
+            array_normalizer: fos_rest.normalizer.camel_keys
+
+Sometimes an array contains a key, which once normalized, will override an
+existing array key. For example ``foo_bar`` and ``foo_Bar`` will both lead to
+``fooBar``. If the normalizer receives this data, the listener will throw a
+BadRequestHttpException with the message ``The key "foo_Bar" is invalid as it
+will override the existing key "fooBar"``.
+
+NB: If you use the ``camel_keys`` normalizer, you must be careful when choosing
```
javiereguiluz: Maybe you could reword this phrase as a note:

```
.. note::

    If you use the ``camel_keys`` normalizer, you must be careful when choosing
    your form name.
```

Resources/doc/3-4-format_listener.rst
```diff
@@ -0,0 +1,129 @@
+3.4 Format listener
+===================
+
+The Request format listener attempts to determine the best format for the
+request based on the Request's Accept-Header and the format priority
```
javiereguiluz: `for the request based on the Request's Accept-Header` looks confusing to me. What about rewording it as:

```
for the request based on the HTTP Accept header
```

Resources/doc/3-6-param_fetcher_listener.rst
```diff
@@ -0,0 +1,142 @@
...
+    parameters at once. And also both ``$paramFetcher->get()`` and
+    ``$paramFetcher->all()`` support and optional ``$strict`` parameter to throw
+    a ``\RuntimeException`` on a validation error.
+
+.. note::
+
+    The ParamFetcher requirements feature requires the symfony/validator
+    component.
+
+Optionally the listener can also already set all configured query parameters as
```
javiereguiluz: `[the listener] can also already set all [...]` sounds a bit weird, but I'm not sure if it's wrong.

javiereguiluz:
The doc reorganization looks right to me and the syntax of the cross-section links looks right too.

xabbuh:
I would not make the number a part of the filename. This way, would have to rename them when new ones get added in the middle or when existing chapters will be removed. If you want the filenames to represent the structure, I suggest to move them to a subdirectory.

If I did not miss anything, you will also have to add toctree entries for the new files.

lsmith77:
@xabbuh yeah wasn't sure about the numbers in the file names either. was just following the current approach.

WouterJ:
I think you used numbers as the docs were rendered on GitHub. The only way to order them in such case is by numbering your pages. On symfony.com however, the toctree is what determines the order.

About the reorganization: Generally speaking, shorter pages, shorters sections and shorter sentences make things more easy to follow and read (of course, too short isn't good). I guess this change makes sense. (it's also easier to link to specific things and google can sent you to the direct listener immediately this way)

lsmith77:
I also wanted to prevent the old links from breaking, which is why I left the old listener overview page as is. This is also why I do not want to drop the numbering now. But I guess I can leave the numbering out of the new documents if you think its better.

xabbuh:
At least, it would help you keep URLs if you change the order in which the documents are referenced in the future.

lsmith77:
I have did the changes as suggested by @javiereguiluz. I also removed the numbers from the new files. good to merge?

Resources/doc/3-listener-support.rst
```diff
@@ -25,8 +25,8 @@ enable all listeners:
         view:
             view_response_listener: 'force'
 
-View Response listener
-----------------------
+3.1 View Response listener
```
xabbuh: For consistency, I would uppercase the L of listener.

Resources/doc/3-listener-support.rst
```diff
@@ -34,152 +34,10 @@ processed via the listener by the ``fos_rest.view_handler`` service.
...
-     *
-     * @View(populateDefaultVars=false)
-     */
-    public function getUserDetails(User $user)
-    {
-    }
-
-Body listener
--------------
+3.2 Body listener
```
xabbuh: same here

Resources/doc/3-listener-support.rst
```diff
@@ -187,312 +45,36 @@ example allows to receive data that normally would be sent via POST as
...
-        if (count($validationErrors) > 0) {
-            // Handle validation errors
-        }
-
-        // ...
-    }
-
-Format listener
----------------
+3.4 Format listener
```
xabbuh: and here

Resources/doc/3-listener-support.rst
```diff
@@ -187,312 +45,36 @@ example allows to receive data that normally would be sent via POST as
...
-
-The matched version is set as a Request attribute with the name ``version``,
-and when using JMS serializer it is also set as an exclusion strategy
-automatically in the ``ViewHandler``.
-
-See `this JMS Serializer article`_ for details.
-
-Mime type listener
-------------------
+3.5 Mime type listener
```
xabbuh: here too

Resources/doc/3-listener-support.rst
```diff
@@ -501,151 +83,16 @@ class. It works similar to the `mime type listener`_ proposed by Symfony.
         view:
             mime_types: {'jsonp': ['application/javascript+jsonp']}
 
-Param fetcher listener
-----------------------
+3.6 Param Fetcher listener
```
xabbuh: oh and here too, seems that all "listeners" except one were lowercased before

Resources/doc/body_listener.rst
```diff
@@ -0,0 +1,90 @@
+3.2 Body listener
+=================
+
+The Request body listener makes it possible to decode the contents of a request
+in order to populate the "request" parameter bag of the Request. This for
+example allows to receive data that normally would be sent via POST as
```
xabbuh: This, for example, allows [...]

PR #1034, opened 2015-05-12
Make it possible to override listener services, add extension point to body listener

lsmith77:
fixes https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1002
docs should get updated as well


DependencyInjection/Configuration.php
```diff
@@ -168,7 +205,11 @@ private function addBodyListenerSection(ArrayNodeDefinition $rootNode)
                     ->fixXmlConfig('decoder', 'decoders')
                     ->addDefaultsIfNotSet()
                     ->canBeUnset()
+                    ->treatFalseLike(array('enabled' => false))
```
lsmith77: I tried using ``canBeEnabled()`` here, but it didn't work as expected as it ended by defaulting ``enabled`` to ``false``, so I just added the code for ``canBeEnabled()`` for this, removed ``treatNullLike()`` and switched the default of ``enabled`` to ``true``.

stof: use ``canBeDisabled`` if you want it to be enabled by default

DependencyInjection/Configuration.php
```diff
@@ -42,18 +42,37 @@ public function getConfigTreeBuilder()
...
+                    ->children()
+                        ->scalarNode('service')->defaultNull()->end()
+                    ->end()
                 ->end()
                 ->scalarNode('cache_dir')->cannotBeEmpty()->defaultValue('%kernel.cache_dir%/fos_rest')->end()
-                ->scalarNode('allowed_methods_listener')->defaultFalse()->end()
+                ->arrayNode('allowed_methods_listener')
+                    ->canBeEnabled()
+                    ->children()
+                        ->scalarNode('service')->defaultNull()->end()
```
stof: looks like you only care about it being empty or no. So a boolean ``disable_listener`` might be better as you don't care about the actual id.

stof: and for this case, you don't even need anything IMO: given there is no specific config, it can just be disabled.

DependencyInjection/Configuration.php
```diff
@@ -137,10 +166,18 @@ private function addViewSection(ArrayNodeDefinition $rootNode)
...
                         ->end()
-                        ->scalarNode('view_response_listener')->defaultFalse()
+                        ->arrayNode('view_response_listener')
+                            ->beforeNormalization()
+                                ->ifString()->then(function($v) { return array('enabled' => true, 'mode' => $v); })
+                            ->end()
                             ->validate()
-                                ->ifNotInArray($this->forceOptionValues)
-                                ->thenInvalid('The view_response_listener option does not support %s. Please choose one of '.json_encode($this->forceOptionValues))
+                                ->ifTrue(function ($v) { return !in_array($v['mode'], $this->forceOptionValues); })
```
xabbuh: This does not work on PHP 5.3.

DependencyInjection/Configuration.php
lsmith77: I am trying to be forwards compatible and consistent across all listeners

lsmith77:
ok, think its good now. once its merged I will release 1.6.0-RC1

lsmith77:
anyone running into issues with this PR please try https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1038

PR #1036, opened 2015-05-13
Setting default format in RestRouteCollection.php only if not set already

Wirone:
Scenario:

`config.yml`:

    fos_rest:
        format_listener: true
        routing_loader:
            default_format: json

`ReportController.php` (simplified)

    /**
     * @Rest\Get("/report/export", defaults = {"_format": "xlsx"}, requirements = {"_format": "xlsx|csv|json|xml"})
     */
    public function exportAction(Report $report, HttpFoundation\Request $request) {
        return $this->get('report.exporter')->getReponse($report, $request->getRequestFormat('xlsx'));
    }

Annotation sets default format to `xlsx`, but `FOS\RestBundle\Routing\Loader\RestRouteLoader::load()` after reading all routes invokes `FOS\RestBundle\Routing\RestRouteCollection::setDefaultFormat()` which overrides it with global `default_format` from bundle config. When `/report/export` is accessed without format, global `default_format` is used even if there is custom default value in annotation.

This PR fixes it - default format is set only if it wasn't set already via annotation.

### DIFF
--- a/Routing/RestRouteCollection.php
+++ b/Routing/RestRouteCollection.php
@@ -62,7 +62,10 @@ class RestRouteCollection extends RouteCollection
     public function setDefaultFormat($format)
     {
         foreach (parent::all() as $route) {
-            $route->setDefault('_format', $format);
+            // Set default format only if not set already (could be defined in annotation)
+            if(!$route->getDefault('_format')) {
+                $route->setDefault('_format', $format);
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


Routing/RestRouteCollection.php
```diff
@@ -62,7 +62,10 @@ public function prependRouteControllersWithPrefix($prefix)
     public function setDefaultFormat($format)
     {
         foreach (parent::all() as $route) {
-            $route->setDefault('_format', $format);
+            // Set default format only if not set already (could be defined in annotation)
+            if(!$route->getDefault('_format')) {
```
lsmith77: missing space after ``if``

PR #1038, opened 2015-05-18
Fix #1037 - Error validating param_fetcher_listener configuration parameter

### LINKED ISSUE #1037: Error validating param_fetcher_listener configuration parameter
FOSRestBundle version: 1.6.0-RC1

I have the following configuration ...
```yaml
fos_rest:
    param_fetcher_listener: true
    ...
```
... and getting the following error:
```
[InvalidArgumentException]                                                                                                     
  The param_fetcher_listener option does not support {"enabled":true,"service":null}. Please choose one of [false,true,"force"]
```

kierate:
This allows the old (string) configuration to work alongside the new (array) configuration making this:

```
param_fetcher_listener: force
```

equivalent to this:

```
param_fetcher_listener:
    enabled: force
```


DependencyInjection/Configuration.php
```diff
@@ -57,12 +57,18 @@ public function getConfigTreeBuilder()
                 ->end()
                 ->scalarNode('unauthorized_challenge')->defaultNull()->end()
                 ->arrayNode('param_fetcher_listener')
-                    ->validate()
-                        ->ifNotInArray(Configuration::$forceOptionValues)
-                        ->thenInvalid('The param_fetcher_listener option does not support %s. Please choose one of '.json_encode(Configuration::$forceOptionValues))
+                    ->beforeNormalization()
+                        ->ifString()
+                        ->then(function ($v) { return array('enabled' => $v); })
```
xabbuh: I am not sure if that really is the best solution. Imho `enabled` should only be `true` or `false`. What do you think about adding a boolean child node `force` which could be set accordingly?

xabbuh:
By the way, the [extension class](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/DependencyInjection/FOSRestExtension.php#L195) needs to be fixed too (we should probably check if something else exists for the other options too).

kierate:
I fixed the extension class to work with force in the enabled key (last commit).

With regards to moving force to a separate key how would you see this? With a true|false for the force key? i.e.:

```yaml
param_fetcher_listener:
    enabled: true|false
    force: true|false    # this would be ignored if enabled=false
```

xabbuh:
@kierate This sounds like the best solution to me (IIRC `force` was the only possible value besides `false` and `true` before).

kierate:
I have separated it out, but just looking at the code - the `view_response_listener` also uses the true|false|force values (Configuration::$forceOptionValues). It would probably make sense to refactor that one as well at some point.

lsmith77:
see https://github.com/FriendsOfSymfony/FOSRestBundle/commit/74d28f309550a1d9905b938fff7a3f28170ac3f0 for the fix for view_response_listener .. also added some testes

Issue #1029, opened 2015-05-07
Name Prefix Ignored in Yaml


jtherin:
I have a small naming problem with name_prefix parameter.

Here is the configuration used :

**app/config/routing.yml**
```
_company_api:
    resource: routing/api.yml
    prefix:   /api
    type:     rest
```
**app/config/routing/api.yml**
```
_company_api_v1:
    type:     rest
    resource: api_v1.yml
    prefix:   /v1
    name_prefix: api_v1_
```
**app/config/routing/api_v1.yml**
```
company_api_v1_somethingbundle:
    type:     rest
    resource: "@CompanyApiV1SomethingBundle/Resources/config/routing.yml"
```
**src/Company/Api/V1/SomethingBundle/Resources/config/routing.yml**
```
company_api_v1_somethingbundle_first:
    type:     rest
    resource: Company\Api\V1\SomethingBundle\Controller\FirstController

company_api_v1_somethingbundle_second:
    type:     rest
    resource: Company\Api\V1\SomethingBundle\Controller\SecondController
```

With an `app/console router:debug`, I get :

```
 get_first                         GET      ANY    ANY  /api/v1/first
 get_second                        GET      ANY    ANY  /api/v1/second
```

But I was expecting something like this :

```
 api_v1_get_first                  GET      ANY    ANY  /api/v1/first
 api_v1_get_second                 GET      ANY    ANY  /api/v1/second
```

The `name_prefix: api_v1_` parameter seems to be ignored :(.
How can I fix this?

lsmith77:
i think I seem to remember having seen this noted before .. best would be to investigate the code inside the Routing namespace in this Bundle to try and come up with a fix.

tim96:
Hello @jtherin , @lsmith77 .
I suggest this solution #1047 .

PR #1047, opened 2015-05-30
Fixed error when name prefix ignored in yaml. Issue #1029

tim96:
Fixed error when name prefix ignored in yaml. Issue #1029.

I check the example from Issue #1029, and I think the problem is inheritance class don't save parent name_prefix. I have the same problem, and my solution in this pull request.
With an app/console router:debug, we have:
```
 get_first                         GET      ANY    ANY  /api/v1/first
 get_second                        GET      ANY    ANY  /api/v1/second
```
If we save parent name prefix the result will be:
```
 api_v1_get_first                         GET      ANY    ANY  /api/v1/first
 api_v1_get_second                        GET      ANY    ANY  /api/v1/second
```


lsmith77:
looks good

PR #1053, opened 2015-06-02
Fix old use of mime_types configuration

ktounet:
Wrong format. "enabled" value format due to new structure of configuration.


lsmith77:
good catch .. can you fix the test as well?

ktounet:
Oops, it's done.

Issue #1051, opened 2015-06-01
Change `_method` to `methods`


Padam87:
Hi,

As of symfony 2.7, every route defined with the rest bundle annotations creates a deprecation warning.

We have about 65 routes, generating 90% of our warnings. Should I create a pr replacing `_method` everywhere? Since the bundle requires 2.3, it should be fine IMO.

What do you think?

stof:
Yes, please send a PR doing the upgrade

lsmith77:
will be fixed in 1.7.0

PR #1054, opened 2015-06-02
Fixed deprecated _method errors

Padam87:
| Q             | A
| ------------- | ---
| Bug fix?      | yes
| New feature?  | no
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #1051 
| License       | MIT

### DIFF
--- a/Controller/Annotations/Route.php
+++ b/Controller/Annotations/Route.php
@@ -22,9 +22,8 @@ class Route extends BaseRoute
     public function __construct(array $data)
     {
         parent::__construct($data);
-        $requirements = $this->getRequirements();
-        $requirements['_method'] = $this->getMethod();
-        $this->setRequirements($requirements);
+
+        $this->setMethods((array) $this->getMethod());
     }
 
     /**
--- a/Routing/Loader/Reader/RestActionReader.php
+++ b/Routing/Loader/Reader/RestActionReader.php
@@ -183,7 +183,7 @@ class RestActionReader
         $routeName    = strtolower($routeName);
         $path         = implode('/', $urlParts);
         $defaults     = array('_controller' => $method->getName());
-        $requirements = array('_method' => strtoupper($httpMethod));
+        $requirements = array();
         $options      = array();
         $host         = '';
         $condition    = null;
@@ -200,12 +200,11 @@ class RestActionReader
                 $condition    = null;
 
                 $annoRequirements = $annotation->getRequirements();
+                $annoMethods = $annotation->getMethods();
 
-                if (isset($annoRequirements['_method'])) {
-                    $methods = explode('|', strtoupper($annoRequirements['_method']));
+                if (!empty($annoMethods)) {
+                    $methods = $annoMethods;
                 }
-
-                unset($annoRequirements['_method']);
 
                 $path         = $annotation->getPath() !== null ? $this->routePrefix.$annotation->getPath() : $path;
                 $requirements = array_merge($requirements, $annoRequirements);
@@ -241,12 +240,7 @@ class RestActionReader
                 }
             }
 
-            if (isset($requirements['_method'])) {
-                $methods = explode('|', $requirements['_method']);
-                unset($requirements['_method']);
-            } else {
-                $methods = array();
-            }
+            $methods = explode('|', strtoupper($httpMethod));
 
             // add route to collection
             $route = new Route(
--- a/Tests/Fixtures/Etalon/annotated_conditional_controller.yml
+++ b/Tests/Fixtures/Etalon/annotated_conditional_controller.yml
@@ -1,86 +1,91 @@
 get_users:
   path:      /users.{_format}
   controller:   ::getUsersAction
-  requirements: {_method: GET}
+  method: GET
 
 get_user:
   path:      /users/{slug}.{_format}
   controller:   ::getUserAction
-  requirements: {_method: GET, slug: '[a-z]+'}
+  requirements: {slug: '[a-z]+'}
+  method: GET
 
 patch_users:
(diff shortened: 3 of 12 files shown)


Routing/Loader/Reader/RestActionReader.php
```diff
@@ -200,13 +200,12 @@ public function read(RestRouteCollection $collection, \ReflectionMethod $method,
                 $condition    = null;
 
                 $annoRequirements = $annotation->getRequirements();
+                $annoMethods = $annotation->getMethods();
 
-                if (isset($annoRequirements['_method'])) {
```
stof: there is another place using ``_method`` in this file

Tests/Fixtures/Etalon/annotated_conditional_controller.yml
```diff
@@ -1,86 +1,91 @@
...
   path:      /users/{slug}/bget.{_format}
   controller:   ::bgetUserAction
-  requirements: {_method: GET}
+  method: GET
 
 bpost_user:
   path:      /users/{slug}/bpost.{_format}
   controller:   ::bpostUserAction
-  requirements: {_method: POST}
+  method: POST
```
stof: this should be ``methods``, not ``method``

Tests/Routing/Loader/RestRouteLoaderTest.php
```diff
@@ -101,10 +101,13 @@ public function testAnnotatedUsersFixture()
         foreach ($etalonRoutes as $name => $params) {
             $route = $collection->get($name);
 
-            if (!array_key_exists('_method', $route->getRequirements())) {
-                unset($params['requirements']['_method']);
+            if (!array_key_exists('requirements', $params)) {
+                $params['requirements'] = array();
```
stof: not needed actually

Tests/Routing/Loader/RestRouteLoaderTest.php
```diff
@@ -101,10 +101,13 @@ public function testAnnotatedUsersFixture()
...
             }
 
+            // Symfony sets _method to keep BC, should be removed in 3.0
+            $params['requirements']['_method'] = $params['method'];
```
stof: It should ``methods``, not ``method``, to match the Symfony naming

Tests/Fixtures/Etalon/annotated_conditional_controller.yml
Padam87: Yep, I know, but the other test fixtures use it like this. I think fixing that should be another PR. Or should I change it here?

Routing/Loader/Reader/RestActionReader.php
Padam87: fixed

Routing/Loader/Reader/RestActionReader.php
```diff
@@ -241,16 +239,11 @@ public function read(RestRouteCollection $collection, \ReflectionMethod $method,
...
-                unset($requirements['_method']);
-            } else {
-                $methods = array();
-            }
+            $methods = explode('|', strtoupper($httpMethod));
 
             // add route to collection
             $route = new Route(
-                $path, $defaults, $requirements, $options, $host, array(), $methods, $condition
+                $path, $defaults, array(), $options, $host, array(), $methods, $condition
```
stof: this is wrong. Requirements should still be set in the route

Padam87: At this point the `$requirements` only contained this:

https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1054/files#diff-4f168cae49074c5f30b4abcccc85c281L186

stof: not always. There is code setting the ``_format`` requirement a few lines above

Padam87:
Also fixed the `method` -> `methods` problem.

Padam87:
bump

Padam87:
Awesome, thank you!

Controller/Annotations/Route.php
```diff
@@ -22,9 +22,8 @@ class Route extends BaseRoute
     public function __construct(array $data)
     {
         parent::__construct($data);
-        $requirements = $this->getRequirements();
-        $requirements['_method'] = $this->getMethod();
-        $this->setRequirements($requirements);
+
+        $this->setMethods((array) $this->getMethod());
```
lsmith77: I guess this change is problematic, since we override the method, without checking if something was set before .. ?

lsmith77:
```
14:53 <Zerrvox> lsmith: FosRestBundle 1.7 has a BC break https://github.com/FriendsOfSymfony/FOSRestBundle/blob/bd233aca63b0965171d60fe4c539afa10d45633e/Controller/Annotations/Route.php
14:53 <lsmith> Zerrvox: what breaks?
14:55 <Zerrvox> If you have a route annotation in a FOSRest Controller the method of the generated route ends up being null
14:55 <Zerrvox> which cannot be loaded
```

Padam87: It shouldn't be a BC break.
`(array) null` ends up being an empty `[]`, which is is the default value for the methods property, and this code is in the constructor.

Althoug it might be a problem, tested it with both 1.6 and 1.7, and the methods part is definitely overwritten somewhere.

Padam87: https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1067

lsmith77: fixed with https://github.com/FriendsOfSymfony/FOSRestBundle/releases/tag/1.7.1

Issue #1063, opened 2015-06-09
Update DisableCSRFExtension


peshi:
Update for 2.7+

SecurityContextInterface, setDefaultOptions(OptionsResolverInterface).

https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Form/Extension/DisableCSRFExtension.php#L28

https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Form/Extension/DisableCSRFExtension.php#L34

lsmith77:
can you work on this?

peshi:
@lsmith77 There's no BC layer

PR #1066, opened 2015-06-10
Updated DisableCSRFExtension for 2.7+

peshi:
https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1063

### DIFF
--- a/DependencyInjection/Compiler/CsrfExtensionPass.php
+++ b/DependencyInjection/Compiler/CsrfExtensionPass.php
@@ -1,0 +1,34 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\RestBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+
+class CsrfExtensionPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        if($container->hasDefinition('fos_rest.form.extension.csrf_disable')) {
+            $definition = $container->getDefinition('fos_rest.form.extension.csrf_disable');
+
+            if (interface_exists('Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface')) {
+                $tokenStorageReference = new Reference('security.token_storage');
+                $definition->addArgument(new Reference('security.authorization_checker'));
+            } else {
+                $tokenStorageReference = new Reference('security.context');
+            }
+            $definition->replaceArgument(0, $tokenStorageReference);
+        }
+    }
+}
--- a/Form/Extension/DisableCSRFExtension.php
+++ b/Form/Extension/DisableCSRFExtension.php
@@ -12,8 +12,11 @@
 namespace FOS\RestBundle\Form\Extension;
 
 use Symfony\Component\Form\AbstractTypeExtension;
+use Symfony\Component\OptionsResolver\OptionsResolver;
 use Symfony\Component\OptionsResolver\OptionsResolverInterface;
 use Symfony\Component\Security\Core\SecurityContextInterface;
+use Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface;
+use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
 
 /**
  * Class DisableCSRFExtension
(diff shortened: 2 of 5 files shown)


lsmith77:
ah .. we will still need to support 2.3+ .. so this will require some conditional logic, along the lines of what I have done here:
https://github.com/liip/LiipTranslationBundle/pull/39/files#diff-c2c20f7d6267a9f0628c1dda9a00437fR1
https://github.com/liip/LiipTranslationBundle/pull/39/files#diff-a250e7565daed7b88f994541e152e67cR2

peshi:
morpheus

peshi:
@lsmith77 Have a look/try before i submit a pr.

https://github.com/hyperunknown/FOSRestBundle/tree/csrf-ext-patch-2

lsmith77:
@peshi easiest is to just push to the same branch to update this PR

Form/Extension/DisableCSRFExtension.php
```diff
@@ -22,22 +22,18 @@
...
-    public function __construct(SecurityContextInterface $securityContext, $role)
+    public function __construct(AuthorizationChecker $authorizationChecker, $role)
     {
-        $this->securityContext = $securityContext;
+        $this->authorizationChecker = $authorizationChecker;
         $this->role = $role;
     }
 
-    public function setDefaultOptions(OptionsResolverInterface $resolver)
+    public function configureOptions(OptionsResolver $resolver)
```
stof: please keep the ``setDefaultOptions(OptionsResolverInterface $resolver)`` calling the new ``configureOptions(OptionsResolver $resolver)`` as the forward compatibility layer is only implemented in AbstractTypeExtension in 2.7+, not in older versions, so you have to implement it yourselves here to keep compatibility with older versions

Form/Extension/DisableCSRFExtension.php
```diff
@@ -22,22 +22,18 @@
...
     {
-        if (!$this->securityContext->getToken()) {
-            return;
-        }
-
-        if (!$this->securityContext->isGranted($this->role)) {
+        if (!$this->authorizationChecker->isGranted($this->role)) {
```
stof: this will throw an exception in case there is no token (i.e. outside of a firewall), which is why there was the check on the token you removed

Form/Extension/DisableCSRFExtension.php
```diff
@@ -22,22 +22,18 @@
  */
 class DisableCSRFExtension extends AbstractTypeExtension
 {
-    private $securityContext;
+    private $authorizationChecker;
     private $role;
 
-    public function __construct(SecurityContextInterface $securityContext, $role)
+    public function __construct(AuthorizationChecker $authorizationChecker, $role)
```
stof: please never typehint the implementation but the interface (otherwise, you would not be compatible with any bundle extending the service through composition).

And you should accept both the old and the new APIs to be compatible with 2.3+ (this means removing the typehint and replacing it with 2 instanceof check in the constructor)

Resources/config/forms.xml
```diff
@@ -13,7 +13,7 @@
     <services>
         <service id="fos_rest.form.extension.csrf_disable" class="%fos_rest.form.extension.csrf_disable.class%">
             <tag name="form.type_extension" alias="form" />
-            <argument type="service" id="security.context" />
+            <argument type="service" id="security.authorization_checker" />
```
stof: you would need a compiler pass changing this argument to ``security.context`` for older versions

Form/Extension/DisableCSRFExtension.php
```diff
@@ -51,3 +47,4 @@ public function getExtendedType()
         return 'form';
     }
 }
+
```
stof: this extra line should be reverted

peshi:
Is there anything I have missed?

DependencyInjection/Compiler/CsrfExtensionPass.php
```diff
@@ -0,0 +1,32 @@
...
+
+use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+
+class CsrfExtensionPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        $definition = $container->getDefinition('fos_rest.form.extension.csrf_disable');
```
xabbuh: The `fos_rest.form.extension.csrf_disable` service is only loaded if `disable_csrf_role` is configured. So you need to check for its existence first.

DependencyInjection/Compiler/CsrfExtensionPass.php
```diff
@@ -0,0 +1,32 @@
...
+
+        if (interface_exists('Symfony\Component\Security\Core\Authentication\Token\Storage\TokenStorageInterface')) {
+            $tokenStorageReference = new Reference('security.token_storage');
+            $definition->addArgument(new Reference('security.authorization_checker'));
+        } else {
+            $tokenStorageReference = new Reference('security.context');
+        }
+        $definition->replaceArgument(0, $tokenStorageReference);
+    }
+}
```
xabbuh: missing newline

Form/Extension/DisableCSRFExtension.php
```diff
@@ -46,8 +68,14 @@ public function setDefaultOptions(OptionsResolverInterface $resolver)
...
+    {
+        $this->configureOptions($resolver);
+    }
+
     public function getExtendedType()
     {
         return 'form';
     }
-}
+}
```
xabbuh: missing newline

DependencyInjection/Compiler/CsrfExtensionPass.php
```diff
@@ -0,0 +1,35 @@
...
+use Symfony\Component\DependencyInjection\Reference;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+
+class CsrfExtensionPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        if($container->hasParameter('fos_rest.disable_csrf_role'))
+        {
```
xabbuh:
```php
if ($container->hasDefinition('fos_rest.form.extension.csrf_disable') {
```

PR #1072, opened 2015-06-22
Allow to use Symfony's serializer even with JMS Serializer enabled

sroze:
Right now, if we're using the Symfony serializer and there's the JMS's one registered too for any reason, we can't due to a serializer normalizer not removed:

```
  [Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]                                               
  The service "serializer" has a dependency on a non-existent service "fos_rest.serializer.exception_wrapper_normalizer".  
```

This PR fixes the problem.


lsmith77:
can you fix the tests?

sroze:
@lsmith77 tests are fixed, Travis just have network issues..

Issue #842, opened 2014-08-20
Detailed form errors


lucascourot:
At the moment the form errors look something like:

```javascript
{
  "code": 400,
  "message": "Validation Failed",
  "errors": {
    "children": {
      "username": {
        "errors": [
          "This value should not be blank."
        ]
      }
    }
  }
}
```

I would like to have both error message and error code for each validated form field, so the json would look like this example below http://www.vinaysahni.com/best-practices-for-a-pragmatic-restful-api#errors

```javascript
{
  "code" : 1024,
  "message" : "Validation Failed",
  "errors" : [
    {
      "code" : 5432,
      "field" : "first_name",
      "message" : "First name cannot have fancy characters"
    },
    {
       "code" : 5622,
       "field" : "password",
       "message" : "Password cannot be blank"
    }
  ]
}
```

Is it possible to obtain this with Symfony? Is it possible to associate an error code to a validation constraint?  Or do you see any other solution?

broncha:
+1 for consistency between hash and array

renatomefidf:
@lsmith77 How can I return errors from a different kind of exception? Which is not form errors.
I can see all the exceptions being turned into a FlattenException, which means my custom exceptions are not available at ExceptionHandler/Wrapper. How can I achieve custom errors?

Thank you!

eXtreme:
@renatomefidf I'm struggling with the same thing. I tried writing my ExceptionWrapperHandler but there is no original exception available to retrieve additional data from it. Has anyone managed to do this? I have a custom exception which carries command validation errors and I can't find any way to show them within fos rest.

Ener-Getick:
@lsmith77 what do you think about using a custom exception listener with an exception normalizer (something like [the symfony normalizers](https://github.com/symfony/symfony/tree/2.8/src/Symfony/Component/Serializer/Normalizer)) which would convert the exception to an array ?

renatomefidf:
@eXtreme sadly I have no solution for our case, you may need to do custom returns in each place you need it

eXtreme:
@renatomefidf the easiest thing to do (and the thing I did, cuz it just works) is to write a `kernel.exception` listener and return `JsonResponse` with custom format, but it is outside fosrest.

hsz:
@eXtreme 

I had similar issue and solved it with `ExceptionWrapperHandler`:

```class ExceptionWrapperHandler implements ExceptionWrapperHandlerInterface
{

    /**
[...]

PR #1074, opened 2015-06-23
Supports the 2.7 symfony serializer

Ener-Getick:
This PR add the support of the 2.7 symfony serializer.

For that, I created a standard context class that makes the link between the different context format.
To not create bc breaks, the users can still deliver a jms context class to the view but also the new standard class.

In the case that user delivers a jms context to the view, the context is transformed to a standard context.

But I see two problems :
- If the user gets directly the context in the view, he could get a jms context or a standard context. So should we transform the context directly in the view ? And in all the cases he will not have the same context as before...
- If we transform the context outside of the view, we must transform the context in each class that needs the (de)serialization context ...

This PR is not ready yet, I await your feedbacks !


Context/AbstractContext.php
```diff
@@ -0,0 +1,153 @@
...
+ */
+
+namespace FOS\RestBundle\Context;
+
+/**
+ * {@inheritdoc}
+ *
+ * @author Ener-Getick <egetick@gmail.com>
+ */
+abstract class AbstractContext implements ContextInterface
```
lsmith77: I think we might want to make a simpler ``ContextInterface`` maybe leaving out things like version, maxdepth and groups. and that for these 3 things we create additional interfaces that extend from ``ContextInterface``. this way we are more flexible for a possible 3rd serializer that might not provide these features. also afaik the core serializer does not yet support version or max depth (yet).

Context/ContextInterface.php
```diff
@@ -0,0 +1,112 @@
...
+ */
+
+namespace FOS\RestBundle\Context;
+
+/**
+ * Stores the serialization or deserialization context (groups, version, ...).
+ *
+ * @author Ener-Getick <egetick@gmail.com>
+ */
+interface ContextInterface
```
lsmith77: I think we might want to make a simpler ContextInterface maybe leaving out things like version, maxdepth and groups. and that for these 3 things we create additional interfaces that extend from ContextInterface. this way we are more flexible for a possible 3rd serializer that might not provide these features. also afaik the core serializer does not yet support version or max depth (yet).

lsmith77:
generally I like the direction of this PR but I fear you cannot just drop an typehint from a method:
http://3v4l.org/SOhKd

Context/DeserializationContext.php
```diff
@@ -0,0 +1,28 @@
...
+ * @author Ener-Getick <egetick@gmail.com>
+ */
+class DeserializationContext extends AbstractContext implements DeserializationContextInterface
+{
+    /**
+     * Constructor.
+     */
+    public function __construct()
+    {
+        parent::__construct();
```
stof: this does not make sense

View/ViewHandler.php
```diff
@@ -254,18 +258,26 @@ protected function getSerializationContext(View $view)
     {
         $context = $view->getSerializationContext();
 
-        if ($context->attributes->get('groups')->isEmpty() && $this->exclusionStrategyGroups) {
+        if($context instanceof Context) {
+            $JMSAdapter = new JMSContextAdapter();
+            $context = $JMSAdapter->transform($context);
+        }
+
+        if (empty($context->getGroups()) && $this->exclusionStrategyGroups) {
```
stof: passing the return value of a function to ``empty()`` works only on PHP 5.5+. AFAIK, this bundle still supports older versions

lsmith77:
in the end .. maybe we just need to do FOSRestBundle 2.0 ..
there are a few tickets planned there already .. maybe that its enough:
https://github.com/FriendsOfSymfony/FOSRestBundle/milestones/2.0

Context/ContextInterface.php
Ener-Getick: But if we create 3 interfaces, we have to create 3 context classes ? 
So it's not still standard.

Or do you think it's better to test in each class which context it is ?

Context/DeserializationContext.php
Ener-Getick: I think that we can remove this class and create a single context that supports deserialization and serialization. 
I split this class to support the JMSSerializer but we can do it otherwise.

View/ViewHandler.php
Ener-Getick: I didn't work on the compatibility yet.
I wanted before to submit an idea to resolve the problem of the 2.7 serializer.

Do you agree globally with the new classes ?

Ener-Getick:
@lsmith77 if you want to do the 2.0, I think that we should rewrite the bundle to remove everything linked to JMSSerializer.

About this PR, I don't know if we should not just use an array rather than a class.

lsmith77:
in 2.0 I would still want to support JMS serializer ... but we would not have to do complicated BC hacks.

Ener-Getick:
I mean that the bundle should support the jms serializer but not directly. I think it should be done via something included in the bundle and the users should not use directly the jms classes.

Context/ContextInterface.php
Ener-Getick: Does https://github.com/Ener-Getick/FOSRestBundle/commit/6fa9d371452e4b61d9f450a4bab3313bc6a42c27 fix what you wanted ?

Context/DeserializationContext.php
Ener-Getick: I changed that in https://github.com/Ener-Getick/FOSRestBundle/commit/6fa9d371452e4b61d9f450a4bab3313bc6a42c27. I use constants instead.

Context/ContextInterface.php
lsmith77: yeah ...

Ener-Getick:
So do you still want to work on this PR or do you want to work on something more generic for the 2.0 ?

stof:
I agree that in 2.0, the View classes should not expose any JMS class at all. The JMS integration should be completely done internally based on the exposed abstraction

lsmith77:
yeah ..

Ener-Getick:
There are several problems I don't know how to solve. 
It's impossible to not create bc breaks with a feature like that. This PR minimizes the bc breaks but there are several things illogical: user could get a jms context or standard context with ``View::getSerializationContext()`` and if he gets a standard context, there is a risk he don't know how to process it.

Do you think this bc breaks are acceptable or should we report this PR to the 2.0 ?

lsmith77:
yeah .. lets screw BC and make this a 2.0 PR

Ener-Getick:
Ok so I will remove the JMS direct support.

Ener-Getick:
In this case, I think it's easier to use an array instead of a class.

Ener-Getick:
Do you prefers classes or an array to support a serialization context ?

lsmith77:
I would prefer a class

Ener-Getick:
Ok so I revert the last commits

Ener-Getick:
Should I create a chain class for adapters and define it in the symfony container or should I process the adapters with conditions ?

Ener-Getick:
I don't know how to rewrite the request converters tests ... If you have an idea...

Ener-Getick:
The problem of using classes is that it's more complicated and less user-friendly :/

lsmith77:
I think a class is much more user friendly since it makes the structures explicit. What are you struggeling with specifically?

Ener-Getick:
The problem is that I use the container in ``AbstractRequestBodyParamConverter`` (I don't know if it is the good way) to get a context adapter. So the tests are more complicated but I will do fix the bugs.
I will maybe rewrite ``AbstractRequestBodyParamConverter`` to have less duplicated code.

Ener-Getick:
@lsmith77 do you see something to change ?

lsmith77:
i will try to find time this afternoon to have a look. sorry that I have been so unresponsive in terms of detailed feedback.

Context/Adapter/ContextAdapterInterface.php
```diff
@@ -0,0 +1,46 @@
...
+      *
+      * @param ContextInterface $context
+      * @param int            $direction
+      * @param mixed            $serializer
+      *
+      * @return mixed
+      *
+      * @throws \LogicException if the specialization is not supported
+      */
+     public function specialize(ContextInterface $context, $direction, $serializer);
```
lsmith77: not entirely happy with this method name .. maybe ``configure()`` ?

lsmith77: also not sure about ``$direction`` .. but don't have a good idea there either

lsmith77: another idea for the method name ``convert()`` .. for ``$direction`` maybe ``$type`` .. but not so great ..

lsmith77: to get rid of ``$direction`` we could also split up this method.

right now the adapters are stateless .. but maybe it would be better to require the serializer to be set on the adapter. this way we can use typehints to ensure the right serializer is passed in.

lsmith77: so maybe something like:

```
namespace FOS\RestBundle\Context\Adapter;

use FOS\RestBundle\Context\ContextInterface;

interface ContextAdapterInterface
{
      public function convertSerializationContext(ContextInterface $context);
      public function convertDeserializationContext(ContextInterface $context);
      public function supportsSerialization(ContextInterface $context);
      public function supportsDeserialization(ContextInterface $context);
}
```

we could even take it a step further:

```
namespace FOS\RestBundle\Context\Adapter;

use FOS\RestBundle\Context\ContextInterface;

interface SerializationContextAdapterInterface
{
      public function convertSerializationContext(ContextInterface $context);
      public function supportsSerialization(ContextInterface $context);
}

interface DeserializationContextAdapterInterface
{
      public function convertDeserializationContext(ContextInterface $context);
      public function supportsDeserialization(ContextInterface $context);
}
```
```

File: Request/AbstractRequestBodyParamConverter.php [hunk 6]
```diff
@@ -151,25 +157,33 @@ protected function execute(Request $request, ParamConverter $configuration)
     }
 
     /**
-     * @return DeserializationContext
+     * @return ContextInterface
+     */
+    protected function getContextAdapter()
+    {
+        return $this->container->get('fos_rest.context.adapter.chain_context_adapter');
```
lsmith77: why not inject this service?

Context/Adapter/ContextAdapterInterface.php
Ener-Getick: I didn't set the serializer on the adapter because ``ViewHandler::getSerializer`` allows to change dynamically the serializer by extending the class, no ?

Request/AbstractRequestBodyParamConverter.php
Ener-Getick: That's a good question ... I will create a call action.

And why isn't the serializer injected ?

Context/Adapter/ContextAdapterInterface.php
lsmith77: we can set the serializer inside the ViewHandler I guess before doing the conversion.

Request/AbstractRequestBodyParamConverter.php
lsmith77: what do you mean? the serializer is injected ..

Context/Adapter/ContextAdapterInterface.php
Ener-Getick: I prefers to split the adapter interface in two interfaces. If someone wants to change an adapter comportment, it's easier like that.

Ener-Getick: Yes, I thought you wanted to inject it by the dependency container.

Tests/Request/RequestBodyParamConverter20Test.php
```diff
@@ -182,8 +190,8 @@ public function testApplyWithSerializerContextOptionsForSymfonySerializer()
     {
         $this->serializer = $this->getMock('Symfony\Component\Serializer\SerializerInterface', array('serialize', 'deserialize'));
         $this->converter = new RequestBodyParamConverter20($this->serializer);
+        $this->fillContainer($this->converter);
         $requestBody = '{"name": "Post 1", "body": "This is a blog post"}';
-
         $options = array(
```
Ener-Getick: Is this important ? Because the new context class supports it just with attributes.

Ener-Getick: I can update the ``AbstractRequestBodyConverter::configureDeserializationContext()`` method like this :
```php
    /**
     * @param ContextInterface $context
     * @param array            $options
     *
     * @return ContextInterface
     */
    protected function configureDeserializationContext(ContextInterface $context, array $options)
    {
        foreach($options as $key => $value) {
            if ($key == 'group' && $context instanceof GroupableContextInterface) {
                $context->addGroups($options['groups']);
            } elseif ($key == 'version' && $context instanceof VersionableContextInterface) {
                $context->setVersion($options['version']);
            } elseif ($key == 'maxDepth' && $context instanceof MaxDepthContextInterface) {
                $context->setMaxDepth($options['maxDepth']);
            } elseif ($key == 'serializeNull' && $context instanceof SerializeNullContextInterface) {
                $context->setMaxDepth($options['maxDepth']);
            } else {
                $context->setAttribute($key, $value);
            }
        }

        return $context;
    }
```
Is it ok for you ?

lsmith77: looks fine to me

EventListener/ViewResponseListener.php
```diff
@@ -87,14 +87,10 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
...
-                $context = $view->getSerializationContext() ?: new SerializationContext();
-                $context->setGroups($configuration->getSerializerGroups());
-                $view->setSerializationContext($context);
+                $context->addGroups($configuration->getSerializerGroups());
             }
             if ($configuration->getSerializerEnableMaxDepthChecks()) {
-                $context = $view->getSerializationContext() ?: new SerializationContext();
-                $context->enableMaxDepthChecks();
-                $view->setSerializationContext($context);
+                $context->setMaxDepth(0);
```
lsmith77: is this really equivalent? ``enableMaxDepthChecks()`` just enables checking for max depth, where the max depth is then actually "dynamic", ie. defined on the relevant mapping data in the relevant entities

Tests/Context/Adapter/ArrayContextAdapterTest.php
```diff
@@ -0,0 +1,59 @@
...
+        $serializationContext = $this->adapter->convertSerializationContext($context);
+        $this->assertTrue(is_array($serializationContext));
+        $this->assertEquals(array('groups' => array('a', 'b', 'c'), 'foo' => 'bar', 'version' => 1.3, 'maxDepth' => 10, 'serializeNull' => false), $serializationContext);
+
+        $deserializationContext = $this->adapter->convertDeserializationContext($context);
+        $this->assertTrue(is_array($deserializationContext));
+        $this->assertEquals(array('groups' => array('a', 'b', 'c'), 'foo' => 'bar', 'version' => 1.3, 'maxDepth' => 10), $deserializationContext);
+    }
+
+    public function testSupports() {
```
lsmith77: ``{`` needs to be on the next line

Request/AbstractRequestBodyParamConverter.php
```diff
@@ -106,9 +127,11 @@ protected function execute(Request $request, ParamConverter $configuration)
             $context = $this->context;
         }
 
-        if ($this->serializer instanceof SerializerInterface) {
-            $context = $this->configureDeserializationContext($this->getDeserializationContext(), $context);
+        $context = $this->configureDeserializationContext($this->getDeserializationContext(), $context);
+        if($this->contextAdapter instanceof SerializerAwareInterface) {
```
lsmith77: missing space ... needs to be ``if (``

View/ViewHandler.php
```diff
@@ -445,12 +466,13 @@ private function initResponse(View $view, $format)
...
             $data = $this->getDataFromView($view);
             $serializer = $this->getSerializer($view);
-            if ($serializer instanceof SerializerInterface) {
-                $context = $this->getSerializationContext($view);
-                $content = $serializer->serialize($data, $format, $context);
-            } else {
-                $content = $serializer->serialize($data, $format);
+
+            $standardContext = $this->getSerializationContext($view);
+            if($this->contextAdapter instanceof SerializerAwareInterface) {
```
lsmith77: missing space ... needs to be ``if (``

Request/AbstractRequestBodyParamConverter.php
```diff
@@ -151,26 +174,33 @@ protected function execute(Request $request, ParamConverter $configuration)
...
      */
-    protected function configureDeserializationContext(DeserializationContext $context, array $options)
+    protected function configureDeserializationContext(ContextInterface $context, array $options)
     {
-        if (isset($options['groups'])) {
-            $context->setGroups($options['groups']);
-        }
-        if (isset($options['version'])) {
-            $context->setVersion($options['version']);
+        foreach($options as $key => $value) {
```
lsmith77: missing space ... needs to be ``foreach (``

lsmith77:
ok this is starting to look really good! I think ideally we will have people playing with this in some real applications both with using the advanced features of the JMS and Symfony core serializers.

EventListener/ViewResponseListener.php
Ener-Getick: Yes, for a serialization context it does the same thing :
```php
        if ($context instanceof MaxDepthContextInterface && null !== $context->getMaxDepth()) {
            $newContext->enableMaxDepthChecks();
        }
```

But for a deserialization context it does that in addition :
```php
        if ($context instanceof MaxDepthContextInterface && null !== $context->getMaxDepth()) {
            for ($i = 0; $i < $context->getMaxDepth(); $i++) {
                $newContext->increaseDepth();
            }
        }
```

I have to check if this is really what we want.

lsmith77: ok thx!

Ener-Getick:
We can even imagine something to define the serializer to use.

Ener-Getick:
I will need at least 2 weeks to rewrite the tests because there there is a lot of code not covered... (I'm in vacations this week)

lsmith77:
OK no worries .. in the mean time I will try and find time to test this on some apps and get others involved in the feedback. thank you so much for your efforts .. enjoy your vacation!

Ener-Getick:
Ok, I hope you will like to use it 😀 thx!

Ener-Getick:
I do not understand why the request converter tests (``FOS\RestBundle\Tests\Request``) are copied in the two sub classes... 
We can group them in the parent class ``FOS\RestBundle\Tests\Request\AbstractRequestBodyParamConverterTest`` no ?

lsmith77:
@Ener-Getick the reason is that one of the tests is just for a BC hack that we will remove in 2.0 .. if its causing issues for you just remove https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Tests/Request/RequestBodyParamConverter20Test.php and https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Request/RequestBodyParamConverter20.php since we can drop SensioFrameworkExtraBundle 2.x compatibility in FOSRestBundle 2.0

Ener-Getick:
I rewrote this tests and now they are grouped. I used the Reflection to test ``AbstractRequestBodyParamConverter::execute()`` directly.
I didn't test the dependencies output because I think that we must write tests for each component.
Instead, I tested their parameters.

Context/Adapter/ArrayContextAdapter.php
```diff
@@ -0,0 +1,94 @@
...
+    protected function convertContext(ContextInterface $context)
+    {
+        $newContext = array();
+        foreach ($context->getAttributes() as $key => $value) {
+            $newContext[$key] = $value;
+        }
+
+        if ($context instanceof GroupableContextInterface) {
+            $newContext['groups'] = $context->getGroups();
+        }
```
Ener-Getick: Should we create this, if the groups attribute is empty ?
Because if the user doesn't define context groups, he will have an empty output with the symfony serializer...

Ener-Getick:
I found a problem. ``ConstraintViolationListInterface`` is serialized but not its ``ConstraintViolation``s:
```json
[
  [],
  [],
  []
]
```
I don't know how to solve it ... Maybe a new custom normalizer ?

lsmith77:
yeah .. I think this would make sense .. @dunglas ?

dunglas:
Yes the custom normalizer sounds like the appropriate solution.
I've done it in API Platform for Hydra serialization, maybe can you just adapt it: https://github.com/dunglas/DunglasApiBundle/blob/master/Hydra/Serializer/ConstraintViolationListNormalizer.php

Ener-Getick:
Should I do the same as the current forms errors normalization ?
```json
{
  "code": 400,
  "message": "Validation Failed",
  "errors": {
    "children": {
      "username": {
        "errors": [
          "This value should not be blank."
        ]
      }
    }
  }
}
```

Or something more like https://github.com/FriendsOfSymfony/FOSRestBundle/issues/842 ?

lsmith77:
good question .. I would say what you prefer or rather we may want to end up making this configurable in another PR to fix #842

Ener-Getick:
I think it would be better to fix this in another PR. 
Moreover, I don't really have the time for the moment ...

lsmith77:
ok. i will try to find some time tomorrow for a final review. thank you for your efforts!

Ener-Getick:
Ok, cool ☺

lsmith77:
puh .. totally dropped the ball here .. putting it on my immediate todo list ..

Ener-Getick:
Don't worry, you have the time, there are many things to do before deploying the 2.0.

Context/Adapter/ArrayContextAdapter.php
lsmith77: @douglass ping

Context/Adapter/ChainContextAdapter.php
```diff
@@ -0,0 +1,125 @@
...
+     * @var mixed
+     */
+    protected $serializer;
+
+    /**
+     * Constructor.
+     *
+     * @param array $adapters
+     */
+    public function __construct(array $adapters)
```
lsmith77: should we maybe split the adapters into serialization and deserialization adapters?

Request/RequestBodyParamConverter20.php
```diff
@@ -16,7 +16,7 @@
 
 /**
  * This code is needed for SensioFrameworkExtraBundle 2.x compatibility
- * https://github.com/FriendsOfSymfony/FOSRestBundle/issues/622
+ * https://github.com/FriendsOfSymfony/FOSRestBundle/issues/622.
```
lsmith77: I guess this can be removed in 2.0 but we can leave that to another PR .. but then we should revert this change.

Context/Adapter/ChainContextAdapter.php
Ener-Getick: I don't know, an adapter can be both at the same time ...
Do you mean to split them in the constructor arguments or to create two class parameters  to simplify the deserialize/serialize functions ?

Request/RequestBodyParamConverter20.php
Ener-Getick: Oups, I didn't see this (Php-cs-fixer changes everything).
I think it's better to leave this to another PR, to discuss if we should support sensio 2.x still.
Unless you are sure that we can remove that ?

lsmith77: yeah .. we will keep sensio 2.x support for now which is why I was saying please remove the above change.

Context/Adapter/ChainContextAdapter.php
lsmith77: I think we should split them inside the constructor foreach loop .. so that at run time we do not have to do the ``instanceof`` check anymore

Request/RequestBodyParamConverter20.php
Ener-Getick: I removed the changes made by php-cs-fixer that are not linked to this PR.

Ener-Getick:
Is there a problem with travis ci ?

Context/Adapter/ArrayContextAdapter.php
```diff
@@ -0,0 +1,94 @@
...
+     */
+    public function convertSerializationContext(ContextInterface $context)
+    {
+        if (!$this->supportsSerialization($context)) {
+            throw new \LogicException(sprintf('%s can\'t convert this serialization context.', get_class($this)));
+        }
+
+        $newContext = $this->convertContext($context);
+        if ($context instanceof SerializeNullContextInterface) {
+            $newContext['serializeNull'] = $context->getSerializeNull();
```
amenophis: Why not use setter here ?

Context/Adapter/ArrayContextAdapter.php
amenophis: Why not use setter here ?

Context/Adapter/ArrayContextAdapter.php
```diff
@@ -0,0 +1,94 @@
...
+     * {@inheritdoc}
+     */
+    public function convertSerializationContext(ContextInterface $context)
+    {
+        if (!$this->supportsSerialization($context)) {
+            throw new \LogicException(sprintf('%s can\'t convert this serialization context.', get_class($this)));
+        }
+
+        $newContext = $this->convertContext($context);
+        if ($context instanceof SerializeNullContextInterface) {
```
amenophis: Why this is not done in convertContext directly ?

Ener-Getick: Because I think this option should not be present in a deserialization context.

Context/Adapter/ArrayContextAdapter.php
Ener-Getick: Should we add a new config parameter to define that ?

Ener-Getick:
Should we create documentations for this pr?

Context/Adapter/ArrayContextAdapter.php
lsmith77: we currently have a setting for default groups https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/DependencyInjection/Configuration.php#L111

lsmith77:
yeah .. docs would be good .. also needs a rebase

Ener-Getick:
I added a paragraph explaining how to use the ``Context`` class.

lsmith77:
I guess all open discussion points are taken care of here .. right?

would still be good to get the blessing from @dunglas

dunglas:
I've looked quickly but it looks good to me! Great work @Ener-Getick

Ener-Getick:
Yes, everything is done.

amenophis:
hi @lsmith77,
any idea when you this PR will get merged ?

lsmith77:
done :)

Issue #1078, opened 2015-06-29
Using setTemplateData on annotation actions


gtrias:
I'm trying to get my controllers format agnostic and I really would like to be able to use the @View annotations in conjunction with the method setTemplateData in order to pass variables to template not to serialize.

As in documentation but for the @View annotation actions:

```
        $view = $this->view($products, 200)
            ->setTemplate("MyBundle:Category:show.html.twig")
            ->setTemplateVar('products')
            ->setTemplateData($templateData)
        ;
```

Is there any way to accomplish this? I cannot found anything on documentation either stackoverflow, etc...

lsmith77:
this should already work .. what behavior are you experiencing that makes you believe it doesn't?

gtrias:
Well, I would like to have my actions something like this:

```
    /**
     * @Rest\View()
     */
    public function indexAction()
    {
        ...
        $dataNotToSerialize = array();
        $this->setTemplateData($dataNotToSerialize); // This is the functionality that I missing
        return $data;
    }
```

Instead of this: 

```
    public function indexAction()
    {
        ...
        $view = $this->view($entities, 200)
            ->setTemplate("ACSACSPanelBundle:Domain:index.html.twig")
            ->setTemplateVar("entities")
            ->setTemplateData($dataNotToSerialize)
        ;
        return $this->handleView($view);
    }
```
I would like to omit the setTemplate and setTemplateVar as it could be handled by the View annotation.. 
Is it possible to be achieved what I want?

lsmith77:
yes from memory it should be possible to do ``@Rest\View(template="..",templateVar="..")``
if this works feel free to send a PR to illustrate this in the docs

gtrias:
Ok I see, 

So this works well.

```
    /**
     * @Rest\View()
     */
    public function indexAction()
    {
        ...
        $view = $this->view($entities, 200)
            ->setTemplateData($dataNotToSerialize)
        ;
        return $this->handleView($view);
    }
```

I would like to have a way to return directly the data to be serialized as you can do with the View annotation but keeping the possibility to set data not to be serialized. But this approach is OK for me.

Should I create a PR to illustrate this scenario in the docs?

lsmith77:
well you cannot return 2 pieces of data, the serialized data and the template data, without a structure of sorts and this structure is a View instance.

[...]

PR #1081, opened 2015-07-06
Adding details how to use setTemplateData when @View() annotation #1078

gtrias:
Added instructions about how to set a controller action setting template data not for serialization.

### DIFF
--- a/Resources/doc/view_response_listener.rst
+++ b/Resources/doc/view_response_listener.rst
@@ -71,14 +71,14 @@ explicit configuration. So make sure to remove or disable the following setting:
             return $data;
         }
     }
-    
+
 If you need to pass additional data to template, not for serialization,
-you should return a $view object with the data seted by ``setTemplateData``.
+you should return a $view object with the data set by ``setTemplateData``.
 
 .. code-block:: php
 
     <?php
-    
+
     /**
      * @View()
      */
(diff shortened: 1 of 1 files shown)


Resources/doc/view_response_listener.rst
```diff
@@ -72,6 +72,25 @@ explicit configuration. So make sure to remove or disable the following setting:
...
+
+    <?php
+
+    /**
+     * @View()
+     */
+    public function getUsersAction()
+    {
+        // ...
+        $view = $this->view($data, 200)
```
lsmith77: hmm actually it might be clearer to use ``new View`` here as ``$this->view`` isnt available by default

Resources/doc/view_response_listener.rst
```diff
@@ -72,6 +72,25 @@ explicit configuration. So make sure to remove or disable the following setting:
         }
     }
 
+If you need to pass additional data to template, not for serialization,
+you should return a $view object with the data set by ``setTemplateData``.
```
xabbuh: I would enclose `$view` with double backticks too.

Resources/doc/view_response_listener.rst
```diff
@@ -72,6 +72,25 @@ explicit configuration. So make sure to remove or disable the following setting:
         }
     }
 
+If you need to pass additional data to template, not for serialization,
```
xabbuh: [...] to a template [...] or [...] to the template [...]

Resources/doc/view_response_listener.rst
gtrias: @lsmith77 let me try if it still works creating a new View instance

PR #1105, opened 2015-08-29
prepare for 2.0

lsmith77:
Compatibility with Symfony <2.7, JMS Serializer/SerializerBundle <1.0 and SensioFrameworkExtraBundle <3.0 was dropped


.travis.yml
```diff
@@ -22,22 +21,10 @@ branches:
...
-    - php: 5.5
-      env: SYMFONY_VERSION='2.3.* sensio/framework-extra-bundle:2.*'
-    - php: 5.5
-      env: SYMFONY_VERSION=2.4.*
-    - php: 5.5
-      env: SYMFONY_VERSION=2.6.*
-    - php: 5.5
-      env: SYMFONY_VERSION='2.7.*'
+    - php: 5.4
+      env: SYMFONY_VERSION=2.7.* COMPOSER_FLAGS="--prefer-lowest"
```
stof: ``COMPOSER_FLAGS="--prefer-lowest"`` must be done on a build which does not change the Symfony requirements at all, otherwise you are not actually testing the lowest bounds (as you modify them)

DependencyInjection/Configuration.php
```diff
@@ -186,7 +186,6 @@ private function addViewSection(ArrayNodeDefinition $rootNode)
                             ->canBeUnset()
                             ->children()
                                 ->scalarNode('callback_param')->defaultValue('callback')->end()
-                                ->scalarNode('callback_filter')->defaultValue('/(^[a-z0-9_]+$)|(^YUI\.Env\.JSONP\._[0-9]+$)/i')->end()
```
stof: you should first add a deprecation warning in 1.x versions when configuring this, to warn people *before* they upgrade (Symfony does not log deprecations triggered that early for now, but there is a pending PR for that)

stof:
actually, you should make a release triggering deprecation warnings in many places before going to 2.0, to make the upgrade path easier

Form/Extension/DisableCSRFExtension.php
```diff
@@ -26,41 +26,33 @@
...
+     * @var AuthorizationCheckerInterface
+     */
     private $authorizationChecker;
+    /**
+     * @var string
+     */
+    private $role;
 
-    public function __construct($tokenStorage, $role, $authorizationChecker = null)
+    public function __construct(TokenStorageInterface $tokenStorage, AuthorizationCheckerInterface $authorizationChecker, $role)
```
stof: you should avoid changing the order of arguments. It is a useless BC break

lsmith77: I doubt that anyone has extended this class ..

Resources/config/request_body_param_converter.xml
```diff
@@ -4,6 +4,10 @@
     xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
     xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
+    <parameters>
+        <parameter key="fos_rest.converter.request_body.class">FOS\RestBundle\Request\RequestBodyParamConverter</parameter>
```
stof: I suggest removing all ``*.class`` parameters in the new major version as this practice is discouraged in Symfony (we need a major version to stop using parameters, but we should not wait until the next one)

Form/Extension/DisableCSRFExtension.php
lsmith77: anyway .. I will revert the order change

PR #1095, opened 2015-08-17
Reorganize Util classes (BC layers)

Ener-Getick:
I applied #985 directives in this PR.

>I think there are a lot of classes in the ``Util`` namespace that could be located in more specific namespaces:
>
> - ``FormatNegotiator``, ``MediaTypeNegotiatorInterface``, ``FormatNegotiatorInterface`` could be in a ``Negotiation`` namespace.
>
> - ``ViolationFormatter``, ``ViolationFormatterInterface`` could go to a ``Validator`` namespace.
>
> - classes located in ``Util/Inflector`` could go to ``Inflector``

PR #1104, opened 2015-08-28
[insight] Fix "The request service should never be injected"

Minishlink:
https://insight.sensiolabs.com/what-we-analyse/symfony.inject_request_service

Inject request stack instead.

It is compatible with Symfony >= 2.4, thus the travis build failing on 2.3 configurations. It is worth noting that the last LTS is now 2.7.3.

PR #1107, opened 2015-08-29
Reorganize Util classes (with BC breaks)

Ener-Getick:
In this PR, I rewrote the ``FormatNegotiator`` to use willdurant/negotiation FormatNegotiatorInterface and the requestStack.
I also remove the class/methods deprecated in #1095.

Only the last commit is relevant; the others are from #1095.


lsmith77:
please rebase on master

lsmith77:
though you might also want to wait on #1104

Ener-Getick:
As you want, it doesn't change anything.

EventListener/FormatListener.php
```diff
@@ -51,15 +50,15 @@ public function onKernelRequest(GetResponseEvent $event)
...
             $format = $request->getRequestFormat(null);
             if (null === $format) {
-                if ($this->formatNegotiator instanceof MediaTypeNegotiatorInterface) {
-                    $mediaType = $this->formatNegotiator->getBestMediaType($request);
-                    if ($mediaType) {
-                        $request->attributes->set('media_type', $mediaType);
-                        $format = $request->getFormat($mediaType);
-                    }
-                } else {
+                // if ($this->formatNegotiator instanceof MediaTypeNegotiatorInterface) {
```
lsmith77: just removed outdated code

lsmith77:
yeah I also just noticed that #1104 only focuses on ParamFetcher ..

anyways .. looks like this needs a rebase again ..

EventListener/FormatListener.php
Ener-Getick: I have to adapt this code but I don't know how work the version listener...

Ener-Getick:
We must change the minimum version of willdurand/negotiation if we want to use the ``FormatNegotiationInterface``. This interface only appear with the 1.3.0 ...

EventListener/FormatListener.php
lsmith77: yeah .. it tries to parse a version string from the mime type

Ener-Getick:
@willdurand said that willdurand/negotiation@2.0 will be released as stable in a few weeks (https://github.com/willdurand/Negotiation/issues/56#issuecomment-127066087).

I think it could be a good idea to update this dependency version, no ?

Controller/ExceptionController.php
```diff
@@ -195,7 +195,10 @@ protected function getFormat(Request $request, $format)
     {
         try {
             $formatNegotiator = $this->container->get('fos_rest.exception_format_negotiator');
-            $format = $formatNegotiator->getBestFormat($request) ?: $format;
+            $accept = $formatNegotiator->getBest('', array());
```
amenophis: You could put this line in the if below ?

Ener-Getick: Is that a joke?

amenophis: not really, why ?

Ener-Getick: This if checks $accept value, so if I move this line which defines $accept, how could the condition be true?...

lsmith77:
needs a rebase

Ener-Getick:
done.

Controller/ExceptionController.php
```diff
@@ -195,7 +195,10 @@ protected function getFormat(Request $request, $format)
...
+            if($accept) {
```
lsmith77: space missing

EventListener/FormatListener.php
```diff
@@ -51,14 +50,12 @@ public function onKernelRequest(GetResponseEvent $event)
...
-                if ($this->formatNegotiator instanceof MediaTypeNegotiatorInterface) {
-                    $mediaType = $this->formatNegotiator->getBestMediaType($request);
-                    if (null !== $mediaType) {
-                        $request->attributes->set('media_type', $mediaType);
-                        $format = $request->getFormat($mediaType);
-                    }
-                } else {
-                    $format = $this->formatNegotiator->getBestFormat($request);
+                $accept = $this->formatNegotiator->getBest('', array());
+                if (null !== $accept && 0.0 < $accept->getQuality() &&
```
lsmith77: I think the above code is a bit hard to read .. maybe this is better ..

```
if (null !== $accept && 0.0 < $accept->getQuality()) {
    $format = $request->getFormat($accept->getType();
    if (null !== $format) {
        $request->attributes->set('media_type', $accept->getValue());
    }
}

File: EventListener/MimeTypeListener.php [hunk 5]
```diff
@@ -50,7 +46,6 @@ public function onKernelRequest(GetResponseEvent $event)
         if (HttpKernelInterface::MASTER_REQUEST === $event->getRequestType()) {
             foreach ($this->mimeTypes['formats'] as $format => $mimeType) {
                 $request->setFormat($format, $mimeType);
-                $this->formatNegotiator->registerFormat($format, (array) $mimeType, true);
```
lsmith77: this code was necessary to explicitly register supported mime types .. is this really no longer necessary?

Ener-Getick: This is no longer necessary because I now use the ``Request`` to find the format corresponding to a mime type.

lsmith77:
awesome!

Negotiation/FormatNegotiator.php
```diff
@@ -0,0 +1,113 @@
...
+                    if ($header && $extensionHeader) {
+                        $extensionHeader = ','.$extensionHeader;
+                    }
+                }
+                if ($extensionHeader) {
+                    $header .= $extensionHeader.'; q='.$options['prefer_extension'];
+                }
+            }
+
+            $mimeTypes = $this->normalizePriorities(empty($priorities) ? $options['priorities'] : $priorities);
```
lsmith77: this method no longer exists in 2.0

lsmith77: @willdurand I am not sure how we can revive this anyway .. the array with formats to mime types mapping is protected in the ``Request`` class .. so there is no way to nicely "explode" a format to an array of enabled mime types, so I guess users will just need to specify mime types from now on? or we just map a format to the "primary" mime type

lsmith77: @Ener-Getick it seems like we have no tests for this code path :(

Ener-Getick: Hum, the solution may be to copy the ``Request::$formats``array... Or as you said, we can also force the users to only provide mime types.

Ener-Getick: I created a new PR https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1149.

Issue #1108, opened 2015-09-07
Empty content-type for DELETE requests


ktounet:
Hi,

There is a the following function in the BodyListener :
```php
private function isNotAnEmptyDeleteRequestWithNoSetContentType($method, $content, $contentType)
{
    return false === ('DELETE' === $method && empty($content) && null === $contentType);
}
```

When sending a DELETE request with no content-type, It works with apache server but not with nginx. In the latter case, the content-type is equal to an empty string, not a null value.

Does the condition should be changed to ```empty($contentType)``` for compatibility with nginx ?

PS : I checked several times that the content-type header was not sent, and even tried clearing the header with nginx module "HttpHeadersMoreModule". So I don't think it's a bug from nginx, maybe caused by Symfony ?

Thanks.

lsmith77:
yeah, I think this could be done .. please submit a PR for the 1.7 branch

PR #1113, opened 2015-09-14
Use empty function instead of equal to null for DELETE request (nginx compatibility)

ktounet:
https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1108

### DIFF
--- a/EventListener/BodyListener.php
+++ b/EventListener/BodyListener.php
@@ -144,7 +144,7 @@ class BodyListener
      */
     private function isNotAnEmptyDeleteRequestWithNoSetContentType($method, $content, $contentType)
     {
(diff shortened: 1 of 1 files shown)


EventListener/BodyListener.php
```diff
@@ -144,7 +144,7 @@ public function onKernelRequest(GetResponseEvent $event)
      */
     private function isNotAnEmptyDeleteRequestWithNoSetContentType($method, $content, $contentType)
     {
-        return false === ('DELETE' === $method && empty($content) && null === $contentType);
+        return false === ('DELETE' === $method && empty($content) && empty($contentType));
```
xabbuh: I would add a test to avoid future regressions. It might not be obvious at a later time why `empty()` is used here.

PR #1118, opened 2015-09-15
Remove a deprecated call to ContainerInterface::get('request')

### LINKED ISSUE #824: Inject RequestStack instead of Request
Instead of injecting the Request itself, it is advised to instead inject the RequestStack as this won't lead to issues with wrong/widened scopes

Ener-Getick:
This PR fixes the last deprecated call to ``ContainerInterface::get('request')``.

It fixes #824 at the same time.


lsmith77:
BTW .. we also still lack tests for this controller :-/

PR #1097, opened 2015-08-18
Add file uploads support

Ener-Getick:
This PR add file uploads support. (#793)
Currently, there are no requirements supported.

 ```php
<?php

use FOS\RestBundle\Controller\Annotations\QueryParam;

/**
 * @FileParam(
 *   name="foo_file",
 *   key=null,
 *   incompatibles={},
 *   default=null,
 *   description="",
 *   strict=false,
 *   nullable=false,
 *   requirements={},
 *   image=false
 * )
 */
```

I moved the scalar constraints (notBlank, array) in another abstract class ``ScalarParam``, I don't know if this is a really good name ...

The ``requirements`` parameter of ``@FileParam`` must be a constraint or an array that can contain all the fields of http://symfony.com/doc/current/reference/constraints/File.html or http://symfony.com/doc/current/reference/constraints/Image.html. It depends of the ``image`` parameter.

PR #1123, opened 2015-09-23
Fix a bug when the AnnotationReader try to read a Param

Ener-Getick:
This PR fixes a bug reported here: https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1097#issuecomment-142413212.

I also added a test to prevent this to happen again.


Controller/Annotations/Param.php
```diff
@@ -1,46 +0,0 @@
-<?php
```
lsmith77: did we forget to remove this class earlier?

lsmith77:
there seems to be a style issue that needs fixing

Controller/Annotations/Param.php
Ener-Getick: Yes, maybe after a conflict ...

Ener-Getick:
I don't know, it's not related to this PR. Should I fix it anyway ?

PR #1128, opened 2015-09-26
Force a @dev stability

Ener-Getick:
During that willdurand/negotiation is not stable, I think we should force a ``@dev`` stability


stof:
this will only force it for our own testsuite, so is it really needed ?

Ener-Getick:
Hum, sorry, I understood that it would also force it for the projects using FOSRestBundle ...

PR #1134, opened 2015-10-01
use stable release of willdurand/negotiation

xabbuh:
The stable release [is now a thing](https://github.com/willdurand/Negotiation/releases/tag/v2.0.0).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
         "symfony/framework-bundle":       "^2.7",
         "symfony/routing":                "^2.7",
         "doctrine/inflector":             "^1.0",
-        "willdurand/negotiation":         "^2.0.0-alpha1",
+        "willdurand/negotiation":         "^2.0",
         "willdurand/jsonp-callback-validator": "^1.0"
     },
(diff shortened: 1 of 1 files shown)


willdurand:
:+1:

composer.json
```diff
@@ -26,7 +26,7 @@
         "symfony/framework-bundle":       "^2.7",
         "symfony/routing":                "^2.7",
         "doctrine/inflector":             "^1.0",
-        "willdurand/negotiation":         "^2.0.0-alpha1",
+        "willdurand/negotiation":         "^2.0.0",
```
Soullivaneuh: Should be `^2.0`.

xabbuh: we can do that but it does not make a difference

Soullivaneuh: Indeed, forgot the caret rule.

But it's still better to see. :-)

lsmith77: yeah can you fix this minor detail to improve readability?

xabbuh: done

PR #1136, opened 2015-10-03
Use symfony/phpunit-bridge

Ener-Getick:
This PR use ``symfony/phpunit-bridge``. We don't have to manage the annotations registry anymore for the tests and the deprecations are supposed to be reported https://github.com/symfony/phpunit-bridge.


Tests/Functional/Bundle/TestBundle/Controller/SerializerErrorController.php
```diff
@@ -32,6 +32,7 @@ public function exceptionAction()
 
     /**
      * @View
+     * @group legacy
```
stof: this is wrong here. This method is not a test

Tests/bootstrap.php
```diff
@@ -1,31 +0,0 @@
...
-$file = __DIR__.'/../vendor/autoload.php';
-if (!file_exists($file)) {
-    throw new RuntimeException('Install dependencies to run test suite.');
-}
-
-$autoload = require_once $file;
-
-use Doctrine\Common\Annotations\AnnotationRegistry;
-
-AnnotationRegistry::registerLoader(function ($class) {
```
stof: The annotations still need to be registered (even though the code can be simplified)

Tests/Functional/Bundle/TestBundle/Controller/SerializerErrorController.php
stof: and ``@group legacy`` should be added only on test covering your BC layers. This is not used in a test covering BC layers, so it should not be flagged as such (it would hide the fact that the testsuite will break on Symfony 3.0)

Tests/bootstrap.php
Ener-Getick: Symfony/phpunit-bridge registers them: https://github.com/symfony/phpunit-bridge/blob/2.7/bootstrap.php#L29.

Tests/Functional/Bundle/TestBundle/Controller/SerializerErrorController.php
Ener-Getick: The problem is that this tests use a deprecated API, so if I don't flag this tests as legacy, there is an error.

PR #1137, opened 2015-10-04
removed FOS\RestBundle\Util\Codes in favor of Symfony\Component\HttpFoundation\Response

lsmith77:


xabbuh:
@lsmith77 Imho using the constants from this class should be deprecated in FOSRestBundle 1.x in some way. For example, OroCRM refers to the `Codes` class which would probably lead to a lot of applications to break when upgrading.

PR #1139, opened 2015-10-04
Introduced the zone concept

lsmith77:
supersedes https://github.com/FriendsOfSymfony/FOSRestBundle/pull/955

This is an attempt to fix https://github.com/FriendsOfSymfony/FOSRestBundle/issues/760 by using a more global approach than https://github.com/FriendsOfSymfony/FOSRestBundle/pull/780.

To resume the issue is that FOSRest's listeners are registered to all requests and may conflict with other parts of the application (non-REST).

### Feature

This PR introduces a new configuration value allowing to specify request matchers to determine if the listeners will be enabled or not. By default, all listeners are still enabled to all requests.

```yaml
# app/config/config.yml
fos_rest:
    zone:
        - { path: ^/api }
```

In this example, FOSRest listeners will only be enabled for requests whose path matches `^/api`. It is also possible to specify an array of zones:

```yaml
# app/config/config.yml
fos_rest:
    zone:
        - { path: ^/api }
        - { path: /test, ips: 127.0.0.1 }
```

### Inner working

It works with an early request listener matching requests and storing whether the request matches a configured zone as request attribute. This request attribute can be checked by succeeding listeners to determine if they must be enabled.


lsmith77:
What I wonder if we can optimize things a bit more .. potentially unregistering the listeners instead of disabling them via the request attribute.

lsmith77:
or maybe the better approach would be to dynamically register the event listeners. in fact this could allow enabling different combinations listeners for different parts of the application.

florianv:
@lsmith77 Sorry for my absence, I had a lot of work.

Indeed, ideally the listeners configuration should be a child of the `zone` and we could add them at runtime once a zone is matched (it is possible as the EvenDispatcher is mutable)

lsmith77:
what I am thinking of is the following:

```
fos_rest:
    zones:
        api:
            view:
                force_redirects:
                    html: true
                formats:
                    json: true
                    xml: true
                templating_formats:
                    html: false
            allowed_methods_listener: true
            access_denied_listener:
                json: true
            format_listener:
                priorities: [ json, xml ]
                fallback_format: json
                prefer_extension: false
        default:
            view:
                force_redirects:
                    html: true
                templating_formats:
                    html: true
            format_listener:
                priorities: [ html, css ]
                fallback_format: html
                prefer_extension: true
    zone_rules:
        - { path: ^/api, zone: api }
        - { path: ^/, zone: default }
```

as a fall back we can have similar behavior we have on other places if there is only a single zone, then the following two configs would be equivalent:

```
fos_rest:
    zones:
        default:
            view:
                force_redirects:
                    html: true
                templating_formats:
                    html: true
            format_listener:
                priorities: [ html, css ]
                fallback_format: html
                prefer_extension: true
    zone_rules:
        - { path: ^/, zone: default }
```

is equivalent to

```
fos_rest:
    view:
        force_redirects:
            html: true
        formats:
            json: true
            xml: true
        templating_formats:
            html: true
    allowed_methods_listener: true
    access_denied_listener:
        json: true
    format_listener:
        priorities: [ html, json, xml, css ]
        fallback_format: html
        prefer_extension: false
    zone_rules:
        - { path: ^/ }
```

this could be implemented by having a very early listener that dynamically registers all relevant listeners for the given zone that matches.

florianv:
@lsmith77 looks nice :)

ahilles107:
:+1:

lsmith77:
ok .. when no one is defined we no longer register the ZoneListener
furthermore I added a few tests to check what happens of the zone support is enabled but didn't match.

lsmith77:
also allowed 3.0 to fail due to JMSSerializerBundle not supporting 3.0 atm

lsmith77:
@stof would love to get some feedback from you before merging .. but for now I do not see how we can make it faster without a lot of additional work.

.travis.yml
```diff
@@ -24,6 +24,8 @@ branches:
 
 matrix:
   fast_finish: true
+  allow_failures:
+    - env: SYMFONY_VERSION='3.0.*@dev'
```
lsmith77: should be removed once https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1188 is merged

Resources/config/format_listener.xml
```diff
@@ -22,7 +22,7 @@
             <argument type="collection" /> <!-- mime types -->
         </service>
 
-        <service id="fos_rest.request_matcher" class="Symfony\Component\HttpFoundation\RequestMatcher" public="false" />
+        <service id="fos_rest.format_request_matcher" class="Symfony\Component\HttpFoundation\RequestMatcher" public="false" />
```
lsmith77: this is just a cosmetic change as we now have 2 different places where we use a RequestMatcher

AlexandreKilian:
this needs to be added to the docs! I've been looking for ages for that!

xabbuh:
@AlexandreKilian For the records, the last paragraph at http://symfony.com/doc/current/bundles/FOSRestBundle/3-listener-support.html#security-exception-listener describes this, but I completely agree that this should be way more visible and discoverable.

PR #1145, opened 2015-10-05
fix symfony/security constraint

xabbuh:


lsmith77:
heh .. fun little typo!

stof:
would it be possible to depend on the security subcomponents instead of the full security component ?

lsmith77:
@stof sure .. someone just needs to figure out which one exactly we need :)

stof:
@lsmith77 look for any usage of the security classes in the bundle, and look at which namespace they come from :smile:

xabbuh:
If you want to I can look for the actual dependencies later on today. Otherwise, if found this to be done fast, feel free to close here and do the split in a separate PR. :)

stof:
From a quick github search, only ``security-core`` is needed: https://github.com/FriendsOfSymfony/FOSRestBundle/search?utf8=%E2%9C%93&q=Security

PR #1147, opened 2015-10-05
require only needed Symfony security components

xabbuh:
FOSRestBundle does not rely on the full Symfony Security component, but
only uses a few classes from the `symfony/security-core` package.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,9 +28,9 @@ matrix:
     - php: 5.4
       env: COMPOSER_FLAGS="--prefer-lowest"
     - php: 5.5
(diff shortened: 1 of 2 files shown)


lsmith77:
I wonder if we can also get rid of the acl dependency in the travis setup:
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/.travis.yml#L31

xabbuh:
Let's see if Travis agrees.

stof:
@lsmith77 this dependency is needed only because of the dev stability (Symfony 2.8 requires the ACL component for BC reasons)

stof:
ah, it is not needed anymore: Symfony 2.8 now allows to use security-acl 2.7, which is stable.

PR #1149, opened 2015-10-06
Test the format negotiator and reorganize it

Ener-Getick:
This PR add a test for the ``FormatNegotiator``. Several mistakes are fixed.

### DIFF
--- a/Negotiation/FormatNegotiator.php
+++ b/Negotiation/FormatNegotiator.php
@@ -54,14 +54,11 @@ class FormatNegotiator extends BaseNegotiator
     public function getBest($header, array $priorities = [])
     {
         $request = $this->requestStack->getCurrentRequest();
-
-        if (empty($header)) {
-            $header = $request->headers->get('Accept');
-        }
+        $header = $header ?: $request->headers->get('Accept');
 
         foreach ($this->map as $elements) {
             // Check if the current RequestMatcherInterface matches the current request
-            if (null === $elements[0] || !$elements[0]->matches($request)) {
+            if (!$elements[0]->matches($request)) {
                 continue;
             }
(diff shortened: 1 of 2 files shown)


Negotiation/FormatNegotiator.php
```diff
@@ -73,31 +72,27 @@ public function getBest($header, array $priorities = [])
...
+                        $header .= ',';
                     }
-                }
-                if ($extensionHeader) {
                     $header .= $extensionHeader.'; q='.$options['prefer_extension'];
                 }
             }
 
-            $mimeTypes = $this->normalizePriorities(empty($priorities) ? $options['priorities'] : $priorities);
+            $mimeTypes = empty($priorities) ? $options['priorities'] : $priorities;
```
lsmith77: does this mean that priorities now need to be mime types and we no longer support formats?

Ener-Getick: Yes, for the moment I just removed it. I have to think how to replace this ...

lsmith77: ok .. we should then create a follow up ticket for this. its annoying that the Request class does not give access to the mappings in ``Request::formats``.

/cc @Tobion

Ener-Getick: We should maybe create an issue / a pr for symfony 2.8 or 3.0 to add a new method ``Request::getMimeTypes()`` and see what the symfony staff think about this.

Ener-Getick: I created https://github.com/symfony/symfony/pull/17318

Ener-Getick: https://github.com/symfony/symfony/pull/17113 has been merged so we need a new pr on FOSRestBundle :-)

PR #1152, opened 2015-10-06
[WIP] implement flexible zone concept with lazy loading listeners

lsmith77:
relates to https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1139

idea would be to optimize performance and flexibility:

```
fos_rest:
    zones:
        api:
            view:
                formats:
                    json: true
                    xml: true
                templating_formats:
                    html: false
            allowed_methods_listener: true
            access_denied_listener:
                json: true
            format_listener:
                priorities: [ json, xml ]
                fallback_format: json
                prefer_extension: false
        default:
            view: true
            format_listener:
                priorities: [ html, js, css ]
    zone_rules:
        - { path: ^/api, zone: api }
        - { path: ^/, zone: default }
```

as a fall back we can have similar behavior we have on other places if there is only a single zone, then the following two configs would be equivalent:

```
fos_rest:
    zones:
        default:
            view:
                force_redirects:
                    html: true
                templating_formats:
                    html: true
            format_listener:
                priorities: [ html, css ]
                fallback_format: html
                prefer_extension: true
    zone_rules:
        - { path: ^/, zone: default }
```

is equivalent to

```
fos_rest:
    view:
        force_redirects:
            html: true
        formats:
            json: true
            xml: true
        templating_formats:
            html: true
    allowed_methods_listener: true
    access_denied_listener:
        json: true
    format_listener:
        priorities: [ html, json, xml, css ]
        fallback_format: html
        prefer_extension: false
    zone_rules:
        - { path: ^/ }
```

this could be implemented by having a very early listener that dynamically registers all relevant listeners for the given zone that matches.


lsmith77:
I do not think I will have time soon to continue work on this ..

PR #1154, opened 2015-10-06
added support for a default_version setting

lsmith77:


DependencyInjection/FOSRestExtension.php
```diff
@@ -171,6 +171,10 @@ private function loadFormatListener(array $config, XmlFileLoader $loader, Contai
                     'fos_rest.format_listener.media_type.version_regex',
                     $config['format_listener']['media_type']['version_regex']
                 );
+                $container->setParameter(
+                    'fos_rest.format_listener.media_type.default_version',
+                    $config['format_listener']['media_type']['default_version']
+                );
```
xabbuh: I would do something like the following instead:

```php
$container->getDefinition('fos_rest.version_listener')->addArgument($config['format_listener']['media_type']['default_version']);
```

This way you don't have to worry about maintaining container parameters in the future (people might use them).

lsmith77: I guess this applies to tons of other parameters as well ..

xabbuh: Probably, though I think 2.0 would be good release to clean them up (another advantage is that you reduce the size of the compiled container).

lsmith77: sure .. I am open to doing this.

xabbuh: alright, see #1158

Issue #1151, opened 2015-10-06
re-add support for formats as priorities


lsmith77:
right now we just support mime types.

relates to https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1149

Ener-Getick:
I tried to use the current ``Request::getMimeType()`` function. It supports the main mime types but not the aliases...

lsmith77:
exactly .. @Tobion what do you think are the chances to add a method to read ``Request::format`` from the outside?

lsmith77:
@Ener-Getick right now the only option I see is forcing the user to configure the entire mapping again

Ener-Getick:
Another solution is to copy the ``Request::format`` at least temporarily. We can also search if there is no component that contains this mime types map.

lsmith77:
the Negotiation lib used to provide such a mapping in version 1.x ..

Ener-Getick:
If this is too complicated we can always revert to willdurand/negotiation 1.x...

PR #1153, opened 2015-10-06
Re-add support for formats as priorities

Ener-Getick:
This PR re-add a basic support for formats as priorities.
For the moment only one mime type is supported per format.

Relates to #1151 and #1149

PR #1156, opened 2015-10-06
support format priorities

lsmith77:
supersedes https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1153
fixes https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1151

### DIFF
--- a/Negotiation/FormatNegotiator.php
+++ b/Negotiation/FormatNegotiator.php
@@ -86,7 +86,9 @@ class FormatNegotiator extends BaseNegotiator
                 }
             }
 
-            $mimeTypes = empty($priorities) ? $options['priorities'] : $priorities;
+            $mimeTypes = $this->normalizePriorities(
+                empty($priorities) ? $options['priorities'] : $priorities
+            );
             $mimeType = parent::getBest($header, $mimeTypes);
             if ($mimeType !== null) {
                 return $mimeType;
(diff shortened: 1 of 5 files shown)


Ener-Getick:
+1, the only problem remaining is that the aliases are not configured by default. 
I think we can let this bc break.

lsmith77:
yeah. i will add a note in the docs about this change

Negotiation/FormatNegotiator.php
```diff
@@ -103,4 +113,47 @@ public function getBest($header, array $priorities = [])
             }
         }
     }
+
+    /**
+     * @param array $values
+     *
+     * @return array
+     */
+    protected function sanitize(array $values)
```
xabbuh: Shouldn't these methods be private to not expose additionally extension points?

Negotiation/FormatNegotiator.php
```diff
@@ -103,4 +113,47 @@ public function getBest($header, array $priorities = [])
...
+
+    /**
+     * Transform the format (json, html, ...) to their mimeType form (application/json, text/html, ...).
+     *
+     * @param Request $request
+     * @param array   $priorities
+     *
+     * @return array formated priorities
+     */
+    protected function normalizePriorities(Request $request, $priorities)
```
xabbuh: What about adding an `array` type hint for `$priorities`?

lsmith77: +1

Negotiation/FormatNegotiator.php
lsmith77: +1

xabbuh: see #1162

PR #1158, opened 2015-10-06
replace unneeded DI container parameters

xabbuh:


Resources/config/body_listener.xml
```diff
@@ -15,7 +15,7 @@
         <service id="fos_rest.decoder.xml" class="FOS\RestBundle\Decoder\XmlDecoder" />
 
         <service id="fos_rest.decoder_provider" class="FOS\RestBundle\Decoder\ContainerDecoderProvider">
-            <argument>%fos_rest.decoders%</argument>
+            <argument /> <!-- decoders -->
```
stof: I would use ``<argument type="collection">`` here, so that the non-replaced argument is already an array

lsmith77: +1

xabbuh: good idea

lsmith77:
There is an error now:
https://travis-ci.org/FriendsOfSymfony/FOSRestBundle/jobs/84038805#L301

This is caused by https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1156 which now uses the ``fos_rest.mime_types`` in two more places.

xabbuh:
I rebased and fixed the two parts.

Resources/config/request_body_param_converter.xml
```diff
@@ -8,10 +8,10 @@
 
         <service id="fos_rest.converter.request_body" class="FOS\RestBundle\Request\RequestBodyParamConverter">
             <argument type="service" id="fos_rest.serializer"/>
-            <argument>%fos_rest.serializer.exclusion_strategy.groups%</argument>
-            <argument>%fos_rest.serializer.exclusion_strategy.version%</argument>
+            <argument /> <!-- serializer exclusion strategy groups -->
```
stof: shouldn't this be a collection rather than an empty string when loading the file ?

xabbuh: You are right. I also found some more places where such a change was needed.

PR #1161, opened 2015-10-07
Fix several scrutinizer issues

Ener-Getick:
This PR fix several scrutinizer issues.

### DIFF
--- a/Controller/ExceptionController.php
+++ b/Controller/ExceptionController.php
@@ -126,7 +126,7 @@ class ExceptionController extends ContainerAware
      * @param FlattenException $exception
      * @param array            $exceptionMap
      *
-     * @return int|false
+     * @return int|bool
      */
     protected function isSubclassOf($exception, $exceptionMap)
     {
@@ -141,7 +141,7 @@ class ExceptionController extends ContainerAware
                 }
             }
         } catch (\ReflectionException $re) {
-            return 'FOSUserBundle: Invalid class in  fos_res.exception.messages: '
+            return 'FOSUserBundle: Invalid class in fos_res.exception.messages: '
                     .$re->getMessage();
         }
 
--- a/Request/ParamFetcher.php
+++ b/Request/ParamFetcher.php
@@ -14,6 +14,7 @@ namespace FOS\RestBundle\Request;
 use FOS\RestBundle\Controller\Annotations\ParamInterface;
 use FOS\RestBundle\Validator\ViolationFormatterInterface;
 use Symfony\Component\DependencyInjection\ContainerAware;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestStack;
 use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
 use Symfony\Component\Validator\Constraints\File;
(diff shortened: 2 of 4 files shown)


Controller/ExceptionController.php
```diff
@@ -126,7 +126,7 @@ protected function getAndCleanOutputBuffering()
      * @param FlattenException $exception
      * @param array            $exceptionMap
      *
-     * @return int|false
+     * @return mixed False, or the $exceptionMap corresponding value
```
stof: this is wrong. The ``$exceptionMap`` array is meant to contain only integers as values

Request/ParamFetcher.php
```diff
@@ -95,6 +96,8 @@ public function getParams()
 
     /**
      * {@inheritdoc}
+     *
+     * @throws \InvalidArgumentException
```
stof: I don't think this is needed here. the case where it happens is not a case where you would want to catch it (except on a top-level error handler)

Controller/ExceptionController.php
```diff
@@ -141,7 +141,7 @@ protected function isSubclassOf($exception, $exceptionMap)
                 }
             }
         } catch (\ReflectionException $re) {
-            return 'FOSUserBundle: Invalid class in  fos_res.exception.messages: '
+            return 'FOSUserBundle: Invalid class in fos_res.exception.messages: '
```
Ener-Getick: Why the exception is not thrown here ?

Request/ParamFetcher.php
```diff
@@ -188,6 +191,21 @@ public function all($strict = null)
...
     /**
+     * @throws \InvalidArgumentException
+     *
+     * @return Request
+     */
+    private function getRequest()
+    {
+        $request = $this->requestStack->getCurrentRequest();
+        if ($request === null) {
+            throw new \InvalidArgumentException('There is no current request.');
```
stof: a RuntimeException would be better. This is not about validating an argument

Request/ParamFetcher.php
```diff
@@ -188,6 +191,21 @@ public function all($strict = null)
     }
 
     /**
+     * @throws \InvalidArgumentException
+     *
+     * @return Request
+     */
+    private function getRequest()
```
stof: private methods should be after protected ones

Controller/ExceptionController.php
Ener-Getick: No, the ``$exceptionMap`` can also contain boolean.
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/DependencyInjection/Configuration.php#L327
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Controller/ExceptionController.php#L161

Request/ParamFetcher.php
Ener-Getick: Yes, we should probably fix all this mistakes in another PR.

Controller/ExceptionController.php
Ener-Getick: I replace ``false`` by ``bool`` but the class can send a string, do you know why ?

Request/ParamFetcher.php
Ener-Getick: You're right, I changed it.

Controller/ExceptionController.php
lsmith77: since we are inside the exception handler, we would otherwise risk a loop.

xabbuh: Then we need to add `string` as a possible return type too.

PR #1162, opened 2015-10-07
some code style tweaks for the FormatNegotiator

xabbuh:


Negotiation/FormatNegotiator.php
```diff
@@ -18,26 +18,15 @@
...
 {
-    /**
-     * @var array
-     */
     private $map = [];
-
-    /**
-     * @var RequestStack
-     */
     private $requestStack;
```
Ener-Getick: Why did you remove this php docs ?

xabbuh: They did not provide any additional value (the type is already derived by the type hint of the constructor.

Ener-Getick: I don't know most of the packages always document the properties...
@lsmith77 @stof what do you think ?

stof: In Symfony (and in my own code), we only document it when the constructor does not allow inferring it. This avoids having to maintain the phpdoc for them (which could become outdated)

Ener-Getick: Ok, so it's better to remove them when they are useless. Thanks

PR #1165, opened 2015-10-08
send coverage report to Scrutinizer CI only once

### LINKED ISSUE #1163: scrutinizer code coverage statistics
```
Hi Lukas, 

I believe you are sending us code coverage information from multiple builds, however in your configuration, you only have set the ``runs`` to 1. So, my guess is that depending on which of your builds finishes first, we get different data and hence you see these changes in coverage depending on which of your builds sends in the data first.

An easy way to fix that would be to just send us code coverage from the same build or increase the number of runs to the total number of builds (the latter being a bit more maintenance intensive if you change the number of builds).

Let me know if that solves it.

Best

--
Thomas Müller from Scrutinizer
On Wed, Oct 7, 2015 at 08:35 PM, "Lukas Kahwe Smith" <smith@pooteeweet.org> wrote: 
the test coverage keeps jumping up and down by 17% on https://scrutinizer-ci.com/g/FriendsOfSymfony/FOSRestBundle/ ..

is there something I can do to prevent this?
```

@stof IIRC you initially set this up .. what do you think .. only send data for one build or increase the builds?

xabbuh:
With this change, the code coverage report should only be sent once to
the Scrutinizer CI service (after the build with PHP 5.6 has finished).

This fixes #1163.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,8 +45,7 @@ before_install:
 
 install: composer update $COMPOSER_FLAGS --prefer-dist
 
-script: phpunit --coverage-clover=coverage.clover
+script: if [ "$TRAVIS_PHP_VERSION" == "5.6" ]; then phpunit --coverage-clover=coverage.clover; else phpunit; fi
 
 after_script:
-  - if [ "$TRAVIS_PHP_VERSION" != "nightly" ]; then wget https://scrutinizer-ci.com/ocular.phar; fi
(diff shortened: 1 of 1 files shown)


stof:
I suggest changing the command to generate the code coverage only on 5.6 too, as you don't need it if you don't upload it

xabbuh:
indeed, good idea

PR #1166, opened 2015-10-09
Fix deprecated form.type_extension tag alias option usage

Soullivaneuh:
See: https://github.com/symfony/symfony/blob/2.8/UPGRADE-2.8.md#form

### DIFF
--- a/Form/Extension/DisableCSRFExtension.php
+++ b/Form/Extension/DisableCSRFExtension.php
@@ -76,6 +76,9 @@ class DisableCSRFExtension extends AbstractTypeExtension
 
     public function getExtendedType()
     {
-        return 'form';
+        return method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')
+            ? 'Symfony\Component\Form\Extension\Core\Type\FormType'
+            : 'form' // SF <2.8 BC
+        ;
     }
 }
--- a/Resources/config/forms.xml
+++ b/Resources/config/forms.xml
@@ -12,7 +12,8 @@
 
     <services>
         <service id="fos_rest.form.extension.csrf_disable" class="%fos_rest.form.extension.csrf_disable.class%">
-            <tag name="form.type_extension" alias="form" />
+            <!-- "alias" option for SF <2.8 -->
+            <tag name="form.type_extension" alias="form" extended-type="Symfony\Component\Form\Extension\Core\Type\FormType" />
             <argument /> <!-- security.token_storage or security.context for Symfony < 2.6 -->
             <argument>%fos_rest.disable_csrf_role%</argument>
         </service>


Soullivaneuh:
StyleCI is screaming a lot.

Maybe can I backport StyleCI config on 1.7 branch and fix CS?

Otherwise, you should deactivate StyleCI for this branch.

lsmith77:
right now style ci branch disabling is essentially useless. hopefully they will soon address this. as for back porting the CI fixes .. if you are motivated to do this .. feel free but we will then also need to ensure that we can still merge 1.7 into master afterwards.

DependencyInjection/Compiler/CsrfExtensionPass.php
```diff
@@ -29,6 +29,17 @@ public function process(ContainerBuilder $container)
                 $tokenStorageReference = new Reference('security.context');
             }
             $definition->replaceArgument(0, $tokenStorageReference);
+
+            // Symfony <2.8 BC
+            $typeExtensionTag = $definition->getTag('form.type_extension');
+            if (method_exists($definition, 'setShared')) {
```
stof: you should detect the version of the Form component, not of the DI component (they may be different).

Use ``method_exists(AbstractType::class, 'getBlockPrefix')`` (without using the PHP 5.5 feature if needed for compat).

stof: And ``getExtendedType`` needs to make the same check btw

Soullivaneuh: I was searching for a detection with form without success.

I did't see for `getBlockPrefix`, thanks for the tip. :+1:

Soullivaneuh: > And `getExtendedType` needs to make the same check btw

What do you mean? Related to this PR?

Soullivaneuh:
> feel free but we will then also need to ensure that we can still merge 1.7 into master afterwards.

IMO, if the code is actually BC and cs configuration the same, the merge should not be a problem.

Except maybe for deleted code/class on master.

I suggest to merge 1.7 on master just after the CS PR and ignore modification (because CS is already fixed on master). Then you will able to continue merges later.


@stof Review corrected.

xabbuh:
The main problem will probably be the fact that the `master` branch uses the shortcut notation for arrays.

DependencyInjection/Compiler/CsrfExtensionPass.php
stof: The alias here must be the same thing that the return value of ``getExtendedType`` in the type extension (and Symfony 2.8 now validates it early to avoid hard-to-debug issues). So you need to update the type extension to be compatible with 2.8 without deprecations too (your current code is not compatible at all because of the mismatch)

Soullivaneuh:
@xabbuh Please take a look: https://github.com/Soullivaneuh/FOSRestBundle/commit/93e87093d2289567db7b7fa41126b13c096a2f77

No `long_syntax_array` fixer applied. Should not be a problem IMO excepts for some inline array with modified content.

DependencyInjection/Compiler/CsrfExtensionPass.php
```diff
@@ -29,6 +29,17 @@ public function process(ContainerBuilder $container)
...
+            // Symfony <2.8 BC
+            $typeExtensionTag = $definition->getTag('form.type_extension');
+            if (method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')) {
+                $typeExtensionTag['extended-type'] = 'Symfony\Component\Form\Extension\Core\Type\FormType';
+            } else {
+                $typeExtensionTag['alias'] = 'form';
+            }
+            $definition->setTags(array_merge($definition->getTags(), array(
+                'form.type_extension' => $typeExtensionTag,
+            )));
```
stof: This is wrong. You can have multiple tags with the same name.

IMO, the simpler solution would be to remove the tag entirely from the XML file and add it here rather than trying to edit it here.

stof: hmm, even much simpler actually: set both ``alias`` and ``extended-type`` in the tag. the compiler pass in 2.8 will ignore ``alias`` if ``type-extension`` is set, and it will use the alias in 2.7

Soullivaneuh: > hmm, even much simpler actually: set both alias and extended-type in the tag. the compiler pass in 2.8 will ignore alias if type-extension is set, and it will use the alias in 2.7

Oh, indeed.

In this case, would not be better to set both of them directly under the xml config file?

stof: you still need to detect the Symfony version in ``FOS\RestBundle\Form\Extension\DisableCSRFExtension::getExtendedType`` though

Soullivaneuh: > you still need to detect the Symfony version in `FOS\RestBundle\Form\Extension\DisableCSRFExtension::getExtendedType` though

I'm sorry I don't understand. Detecting this for what?

stof: the return value of this method needs to be the new value in Symfony 2.8, based on the feature detection.

and yes, set both aliases in XML as it is possible to do it

Soullivaneuh: > the return value of this method needs to be the new value in Symfony 2.8, based on the feature detection.

I didn't find this one the upgrade note...

Soullivaneuh: Oh! Found it! Thanks, will update this PR.

Soullivaneuh: @stof Note: Add some sub-titles on upgrade note would be better for link reference. :+1:

xabbuh: @Soullivaneuh Second item at https://github.com/symfony/symfony/blob/2.8/UPGRADE-2.8.md:

> Type extension should return the fully-qualified class name of the extended type from `FormTypeExtensionInterface::getExtendedType()` now.

Soullivaneuh:
@stof @lsmith77 done.

stof:
:+1:

PR #1168, opened 2015-10-09
CS backport

Soullivaneuh:
As discussed on https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1166#issuecomment-146811495.

> I suggest to merge 1.7 on master just after the CS PR and ignore modification (because CS is already fixed on master). Then you will able to continue merges later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ phpunit.xml
 vendor
 composer.lock
 composer.phar
+.php_cs.cache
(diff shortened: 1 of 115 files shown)


lsmith77:
maybe I didn't understand the approach .. but it seems there are still CS violations left to fix.

Soullivaneuh:
Strange. Will take a look.

Soullivaneuh:
Seems StyleCI has extra check that PHP-CS-Fixer could not handle.

Will manually fix it.

Soullivaneuh:
@lsmith77 CS fixed.

PR #1169, opened 2015-10-09
Fix special charater bug in schema location

lennerd:
Bugfix to encode schema location pathes if the base path of the project contains special characters like blanks.

I'm not sure how to test this as we use `__DIR__` to locate the current library path.


xabbuh:
Do you only have blanks or also other "special characters" in your path?

lennerd:
Validation also wont work if you have other special characters than blanks. I tested this with the german characters `ä`, `ö` and  `ü`.

lsmith77:
I guess this should go into 1.7?

lsmith77:
closed via https://github.com/FriendsOfSymfony/FOSRestBundle/commit/a812c5ae367acf44a1b07f07d57ef413fb25f365

PR #1171, opened 2015-10-10
stop injecting the container when possible

xabbuh:
Injecting the container in your services is not a good practice. After this, there are only three classes left that rely on the container being injected: The `RestRouteLoader` class needs it to resolve controllers registered as services, the `ParamFetcher` class injects the container into container aware `ParamInterface` implementations, and the parent class of the `ViewResponseListener` uses the container.

### DIFF
--- a/Controller/ExceptionController.php
+++ b/Controller/ExceptionController.php
@@ -11,14 +11,14 @@
 
 namespace FOS\RestBundle\Controller;
 
+use FOS\RestBundle\Negotiation\FormatNegotiator;
 use FOS\RestBundle\Util\ExceptionWrapper;
 use FOS\RestBundle\Util\StopFormatListenerException;
 use FOS\RestBundle\View\ExceptionWrapperHandlerInterface;
 use FOS\RestBundle\View\View;
-use FOS\RestBundle\View\ViewHandler;
+use FOS\RestBundle\View\ViewHandlerInterface;
+use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
 use Symfony\Bundle\FrameworkBundle\Templating\TemplateReference;
-use Symfony\Component\DependencyInjection\ContainerAwareInterface;
-use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Debug\Exception\FlattenException;
@@ -27,9 +27,33 @@ use Symfony\Component\HttpKernel\Log\DebugLoggerInterface;
 /**
  * Custom ExceptionController that uses the view layer and supports HTTP response status code mapping.
  */
-class ExceptionController implements ContainerAwareInterface
+class ExceptionController
 {
-    use ContainerAwareTrait;
+    private $exceptionWrapperHandler;
+    private $formatNegotiator;
+    private $viewHandler;
+    private $templating;
+    private $exceptionCodes;
+    private $exceptionMessages;
+    private $showException;
+
+    public function __construct(
+        ExceptionWrapperHandlerInterface $exceptionWrapperHandler,
+        FormatNegotiator $formatNegotiator,
+        ViewHandlerInterface $viewHandler,
+        EngineInterface $templating,
+        array $exceptionCodes,
+        array $exceptionMessages,
+        $showException
+    ) {
+        $this->exceptionWrapperHandler = $exceptionWrapperHandler;
+        $this->formatNegotiator = $formatNegotiator;
+        $this->viewHandler = $viewHandler;
+        $this->templating = $templating;
+        $this->exceptionCodes = $exceptionCodes;
+        $this->exceptionMessages = $exceptionMessages;
+        $this->showException = $showException;
+    }
 
     /**
      * Creates a new ExceptionWrapper instance that can be overwritten by a custom
@@ -41,10 +65,7 @@ class ExceptionController implements ContainerAwareInterface
      */
     protected function createExceptionWrapper(array $parameters)
     {
-        /** @var ExceptionWrapperHandlerInterface $exceptionWrapperHandler */
-        $exceptionWrapperHandler = $this->container->get('fos_rest.exception_handler');
-
-        return $exceptionWrapperHandler->wrap($parameters);
+        return $this->exceptionWrapperHandler->wrap($parameters);
     }
 
     /**
@@ -76,23 +97,22 @@ class ExceptionController implements ContainerAwareInterface
             $request->headers->get('X-Php-Ob-Level', -1)
         );
         $code = $this->getStatusCode($exception);
-        /** @var ViewHandler $viewHandler */
-        $viewHandler = $this->container->get('fos_rest.view_handler');
-        $parameters = $this->getParameters($viewHandler, $currentContent, $code, $exception, $logger, $format);
-        $showException = $request->attributes->get('showException', $this->container->get('kernel')->isDebug());
+        $parameters = $this->getParameters($this->viewHandler, $currentContent, $code, $exception, $logger, $format);
+        $showException = $request->attributes->get('showException', $this->showException);
+
         try {
-            if (!$viewHandler->isFormatTemplating($format)) {
+            if (!$this->viewHandler->isFormatTemplating($format)) {
                 $parameters = $this->createExceptionWrapper($parameters);
             }
 
             $view = View::create($parameters, $code, $exception->getHeaders());
             $view->setFormat($format);
 
-            if ($viewHandler->isFormatTemplating($format)) {
+            if ($this->viewHandler->isFormatTemplating($format)) {
                 $view->setTemplate($this->findTemplate($request, $format, $code, $showException));
             }
 
-            $response = $viewHandler->handle($view);
+            $response = $this->viewHandler->handle($view);
         } catch (\Exception $e) {
             $message = 'An Exception was thrown while handling: ';
             $message .= $this->getExceptionMessage($exception);
@@ -173,10 +193,9 @@ class ExceptionController implements ContainerAwareInterface
      */
     protected function getExceptionMessage($exception)
     {
-        $exceptionMap = $this->container->getParameter('fos_rest.exception.messages');
-        $showExceptionMessage = $this->isSubclassOf($exception, $exceptionMap);
+        $showExceptionMessage = $this->isSubclassOf($exception, $this->exceptionMessages);
 
-        if ($showExceptionMessage || $this->container->get('kernel')->isDebug()) {
+        if ($showExceptionMessage || $this->showException) {
             return $exception->getMessage();
         }
 
@@ -194,8 +213,7 @@ class ExceptionController implements ContainerAwareInterface
      */
     protected function getStatusCode($exception)
     {
-        $exceptionMap = $this->container->getParameter('fos_rest.exception.codes');
-        $isExceptionMappedToStatusCode = $this->isSubclassOf($exception, $exceptionMap);
+        $isExceptionMappedToStatusCode = $this->isSubclassOf($exception, $this->exceptionCodes);
 
         return $isExceptionMappedToStatusCode ?: $exception->getStatusCode();
     }
@@ -211,8 +229,7 @@ class ExceptionController implements ContainerAwareInterface
     protected function getFormat(Request $request, $format)
     {
         try {
-            $formatNegotiator = $this->container->get('fos_rest.exception_format_negotiator');
-            $accept = $formatNegotiator->getBest('', []);
+            $accept = $this->formatNegotiator->getBest('', []);
             if ($accept) {
                 $format = $request->getFormat($accept->getType());
             }
@@ -230,7 +247,7 @@ class ExceptionController implements ContainerAwareInterface
      * Overwrite it in a custom ExceptionController class to add additionally parameters
      * that should be passed to the view layer.
      *
-     * @param ViewHandler          $viewHandler
+     * @param ViewHandlerInterface $viewHandler
      * @param string               $currentContent
      * @param int                  $code
      * @param FlattenException     $exception
@@ -239,7 +256,7 @@ class ExceptionController implements ContainerAwareInterface
      *
      * @return array
      */
-    protected function getParameters(ViewHandler $viewHandler, $currentContent, $code, $exception, DebugLoggerInterface $logger = null, $format = 'html')
+    protected function getParameters(ViewHandlerInterface $viewHandler, $currentContent, $code, $exception, DebugLoggerInterface $logger = null, $format = 'html')
     {
         $parameters = [
             'status' => 'error',
@@ -283,14 +300,14 @@ class ExceptionController implements ContainerAwareInterface
         // when not in debug, try to find a template for the specific HTTP status code and format
         if (!$showException) {
             $template = new TemplateReference('TwigBundle', 'Exception', $name.$statusCode, $format, 'twig');
-            if ($this->container->get('templating')->exists($template)) {
+            if ($this->templating->exists($template)) {
                 return $template;
             }
         }
 
         // try to find a template for the given format
         $template = new TemplateReference('TwigBundle', 'Exception', $name, $format, 'twig');
-        if ($this->container->get('templating')->exists($template)) {
+        if ($this->templating->exists($template)) {
             return $template;
         }
 
--- a/Decoder/ContainerDecoderProvider.php
+++ b/Decoder/ContainerDecoderProvider.php
@@ -11,27 +11,27 @@
 
 namespace FOS\RestBundle\Decoder;
 
-use Symfony\Component\DependencyInjection\ContainerAwareInterface;
-use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
  * Provides encoders through the Symfony2 DIC.
  *
  * @author Igor Wiedler <igor@wiedler.ch>
  */
-class ContainerDecoderProvider implements DecoderProviderInterface, ContainerAwareInterface
+class ContainerDecoderProvider implements DecoderProviderInterface
 {
-    use ContainerAwareTrait;
-
+    private $container;
     private $decoders;
 
     /**
      * Constructor.
      *
-     * @param array $decoders List of key (format) value (service ids) of decoders
+     * @param ContainerInterface $container The container from which the actual decoders are retrieved
+     * @param array              $decoders  List of key (format) value (service ids) of decoders
      */
-    public function __construct(array $decoders)
+    public function __construct(ContainerInterface $container, array $decoders)
     {
+        $this->container = $container;
         $this->decoders = $decoders;
     }
 
--- a/EventListener/ParamFetcherListener.php
+++ b/EventListener/ParamFetcherListener.php
@@ -13,7 +13,6 @@ namespace FOS\RestBundle\EventListener;
 
 use FOS\RestBundle\FOSRestBundle;
 use FOS\RestBundle\Request\ParamFetcherInterface;
-use Symfony\Component\DependencyInjection\ContainerInterface;
 use Symfony\Component\HttpKernel\Event\FilterControllerEvent;
 
 /**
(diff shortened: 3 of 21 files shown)


xabbuh:
This is ready to be reviewed.

EventListener/ParamFetcherListener.php
```diff
@@ -24,18 +24,18 @@
...
-    private $container;
+    private $paramFetcher;
     private $setParamsAsAttributes;
 
     /**
      * Constructor.
      *
-     * @param ContainerInterface $container
-     * @param bool               $setParamsAsAttributes
+     * @param ParamFetcher $paramFetcher
```
Ener-Getick: There is a ``ParamFetcherInterface``. I think it would be better to use it.

Ener-Getick:
+1

View/ViewHandler.php
```diff
@@ -315,7 +340,7 @@ protected function getTemplating()
     public function handle(View $view, Request $request = null)
     {
         if (null === $request) {
-            $request = $this->container->get('request_stack')->getCurrentRequest();
+            $request = $this->requestStack->getCurrentRequest();
```
Ener-Getick: ``RequestStack::getCurrentRequest()`` can also be null, maybe add a check ?

lsmith77:
IIRC we injected the container in the ExceptionController and ViewHandler mostly because depending on the code paths, not all dependencies were needed. These days there are of course proxy services which could be used here as well.

lsmith77:
@xabbuh so how to best address the performance concerns in regards to optional dependencies ..? I guess our options are 1) leave as is 2) your proposal 3) use proxy service. I guess the proper way to do this would be to try and benchmark this ..

we can of course look for some blog posts (f.e. https://blog.cloudtroopers.com/symfony-2-lazy-loading-services-and-performance-analysis) and try and extrapolate the relevance. maybe we can also look a bit more closely which services are actually expensive to create and which are actually fairly cheap.

xabbuh:
@lsmith77 Do you remember which of the injected services were expected to be slow to initialize? On a quick look it doesn't seem to me that it's actually expensive to construct them.

Decoder/DecoderProvider.php
```diff
@@ -11,21 +11,19 @@
...
 
-use Symfony\Component\DependencyInjection\ContainerAware;
-
 /**
  * Provides encoders through the Symfony2 DIC.
  *
  * @author Igor Wiedler <igor@wiedler.ch>
  */
-class ContainerDecoderProvider extends ContainerAware implements DecoderProviderInterface
+class DecoderProvider implements DecoderProviderInterface
```
lsmith77: this is a good example of a service where I think for performance reasons it would make sense to keep the container injection

Controller/ExceptionController.php
```diff
@@ -26,8 +27,34 @@
...
+    private $templating;
+    private $exceptionCodes;
+    private $exceptionMessages;
+    private $showException;
+
+    public function __construct(
+        ExceptionWrapperHandlerInterface $exceptionWrapperHandler,
+        FormatNegotiator $formatNegotiator,
+        ViewHandlerInterface $viewHandler,
+        EngineInterface $templating,
```
lsmith77: I would make this a lazy service, since it might never be needed for an API only setup

View/ViewHandler.php
```diff
@@ -104,17 +106,33 @@ class ViewHandler extends ContainerAware implements ConfigurableViewHandlerInter
...
+     * @param int                              $failedValidationCode    The HTTP response status code for a failed validation
+     * @param int                              $emptyContentCode        HTTP response status code when the view data is null
+     * @param bool                             $serializeNull           Whether or not to serialize null view data
+     * @param array                            $forceRedirects          If to force a redirect for the given key format, with value being the status code to use
+     * @param string                           $defaultEngine           default engine (twig, php ..)
      */
     public function __construct(
+        RouterInterface $router,
+        $serializer,
+        EngineInterface $templating,
```
lsmith77: I would make this a lazy service, since it might never be needed for an API only setup

View/ViewHandler.php
```diff
@@ -104,17 +106,33 @@ class ViewHandler extends ContainerAware implements ConfigurableViewHandlerInter
...
+     * @param array                            $formats                 the supported formats as keys and if the given formats uses templating is denoted by a true value
+     * @param int                              $failedValidationCode    The HTTP response status code for a failed validation
+     * @param int                              $emptyContentCode        HTTP response status code when the view data is null
+     * @param bool                             $serializeNull           Whether or not to serialize null view data
+     * @param array                            $forceRedirects          If to force a redirect for the given key format, with value being the status code to use
+     * @param string                           $defaultEngine           default engine (twig, php ..)
      */
     public function __construct(
+        RouterInterface $router,
+        $serializer,
```
lsmith77: I would make this a lazy service, since it might never be needed for an HTML request

View/ViewHandler.php
```diff
@@ -104,17 +106,33 @@ class ViewHandler extends ContainerAware implements ConfigurableViewHandlerInter
...
+        EngineInterface $templating,
+        RequestStack $requestStack,
+        ExceptionWrapperHandlerInterface $exceptionWrapperHandler,
```
lsmith77: not sure if this should also be a lazy service .. I guess its fairly cheap ..

lsmith77:
@xabbuh I commented inline

lsmith77:
ping

lsmith77:
need a rebase

Controller/ExceptionController.php
stof: making the service lazy would be hard (as you don't own the service, and it would add overhead to anyone using it all the time because of the proxy wrapper).

However, there is something else needed here: 

- it should be possible to build this service when there is no templating system in the project (assuming you don't use the HTML format)
- I would suggest using Twig directly instead of the templating component wrapping twig, as you only render Twig templates anyway

EventListener/ParamFetcherListener.php
stof: indeed

Controller/ExceptionController.php
lsmith77: good point .. this means we need to dynamically determine in the extension if to inject the service .. or alternatively we can use the syntax to optionally inject the twig service only if its available

xabbuh: Good idea, though which interface should we rely on then?

Decoder/DecoderProvider.php
xabbuh: Agreed, I reverted this change.

Tests/EventListener/ParamFetcherListenerTest.php
```diff
@@ -79,6 +74,7 @@ public function testSettingParamFetcherOnRequest()
 
     public function testParamFetcherOnRequestNoZone()
     {
+        $this->markTestSkipped();
```
xabbuh: I am wondering how the test passed before as the event does not return a callable controller.

xabbuh: Got it, I simply deleted too much from the `ParamFetcherListener` class.

Decoder/DecoderProvider.php
xabbuh: Well, maybe we should inject the container in the constructor as it is required anyway. What do you think?

View/ViewHandler.php
xabbuh: Not sure if that works `fos_rest.serializer` only is an alias. What about still injecting the container and retrieving the serializer from it? We then just have to think about what we do at runtime if the service doesn't exist.

Controller/ExceptionController.php
lsmith77: BTW same applies to the ViewHandler

lsmith77:
please rebase again to get the 3.0 build issue fixed

View/ViewHandler.php
lsmith77: imho its also worth consider using setter injection for those "optional" or non strongly typed dependencies and then using an internal getter that handles the case of a missing dependency

lsmith77: BTW for the serializer we could also consider creating an adapter so that for the ViewHandler we can assume a specific interface

Controller/ExceptionController.php
lsmith77: here is an attempt to solve the issue:
https://github.com/FriendsOfSymfony/FOSRestBundle/commit/93758654c0ac2ed29576c0dcbe7d42122d71bf37

maybe we should do something similar for the ViewHandler?

lsmith77:
@xabbuh FYI I have you write permissions on this repo .. as you know its still best to work through PRs and branches but this makes it easier to collaborate

xabbuh:
I did two more things here:

* Removed the `ContainerAwareInterface` usages where we already always inject the container through a setter. The only places left being "container aware" should now be the request params.
* Introduced adapter for the JMS serializer and Symfony's core serializer. Now we can use lazy services and type hints here. I just wonder if we now also could get rid the of context adapters.

Context/Adapter/JMSContextAdapter.php
```diff
@@ -74,7 +76,7 @@ public function convertDeserializationContext(ContextInterface $context)
      */
     public function supportsSerialization(ContextInterface $context)
     {
-        return $this->serializer instanceof JMSSerializerInterface;
+        return $this->serializer instanceof JmsSerializerAdapter;
```
Ener-Getick: May be better to use the same vendor syntax as in [``jms/serializer``](https://github.com/schmittjoh/serializer/tree/master/src/JMS/Serializer):
``JMSSerializerAdapter``.

Ener-Getick:
:+1: with this change, we could move the ``Context`` transformation in ``Serializer``.

lsmith77:
```
1) FOS\RestBundle\Tests\Functional\RequestBodyParamConverterTest::testRequestBodyIsDeserialized
Argument 1 passed to FOS\RestBundle\Tests\Functional\Bundle\TestBundle\Controller\RequestBodyParamConverterController::putPostAction() must be an instance of FOS\RestBundle\Tests\Functional\Bundle\TestBundle\Controller\Post, null given
```

DependencyInjection/Compiler/SerializerConfigurationPass.php
```diff
@@ -33,13 +33,26 @@ public function process(ContainerBuilder $container)
...
+            if ('%' === $class[0]) {
+                $class = $container->getParameter(substr($class, 1, -1));
+            }
+
+            if (is_subclass_of($class, 'Symfony\Component\Serializer\SerializerInterface')) {
+                $container->setAlias('fos_rest.serializer', 'fos_rest.serializer.symfony');
+            } elseif (is_subclass_of($class, 'JMS\Serializer\SerializerInterface')) {
+                $container->setAlias('fos_rest.serializer', 'fos_rest.serializer.jms');
+            } else {
+                throw new \LogicException(sprintf('Cannot handle serializer of type "%s"', $class));
```
lsmith77: I guess this is one of the draw backs of this .. ie. we now hardcode exactly which Serializer implementations we support

lsmith77:
I guess we should not overburden this PR .. so imho lets get this wrapped up and then do more things in other PRs.

I still think that we should also do something like https://github.com/FriendsOfSymfony/FOSRestBundle/commit/93758654c0ac2ed29576c0dcbe7d42122d71bf37 with all the places where we inject the templating engine.

DependencyInjection/Compiler/SerializerConfigurationPass.php
xabbuh: Well, the idea is to add a config option in the future that would make it possible to inject/create your own serializer service. The compiler pass then would exit earlier [here](https://github.com/xabbuh/FOSRestBundle/blob/service-container/DependencyInjection/Compiler/SerializerConfigurationPass.php#L27-L29) without throwing an exception.

xabbuh:
@lsmith77 I agree with you (93758654c0ac2ed29576c0dcbe7d42122d71bf37 actually is the thing I talked about I would like to include). On the other hand @Ener-Getick and me were talking about backporting certain features to a (potential) 1.8 release to be able to create true BC layers and to provide a smooth upgrade path (see https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1213#discussion_r45556284). What do you think about this?

lsmith77:
I must admit .. I personally do not think such back ports are worth the effort

Ener-Getick:
@xabbuh I merged the master branch and fixed the tests in [this branch](https://github.com/FriendsOfSymfony/FOSRestBundle/tree/CONTAINER) if it helps you ;)

DependencyInjection/Compiler/SerializerConfigurationPass.php
stof: well, to support everything, the compiler pass should also accept configuring a service implementing directly the FOSRestBundle interface

stof:
@lsmith77 if you don't provide a smooth upgrade path from 1.x to 2.x, it means you will need to provide a LTS for 1.x as people would not be able to upgrade all projects.
And it will also make it much harder for bundles depending on FOSRestBundle as they would have to either support only one version (splitting the community between people able to upgrade or now) or make crazy stuff to support both without a simple way to write compatible code.

Note that BC layers are mostly necessary for changes impacting the places using the bundle feature, for instance the annotations.
Bundles extending internal parts of the bundle may be left in front of the above difficulties if you want (but a smooth upgrade path for everyone would be easier)

Resources/config/serializer.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" ?>
+<container xmlns="http://symfony.com/schema/dic/services"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="fos_rest.serializer.jms" class="FOS\RestBundle\Serializer\JmsSerializerAdapter" public="false" lazy="true">
+            <argument type="service" id="jms_serializer.serializer" />
```
Ener-Getick: This will throw an error if the service doesn't exist, right ?

DependencyInjection/Compiler/SerializerConfigurationPass.php
xabbuh: I agree and that's what I mean with the early exit in the compiler pass when the `fos_rest.serializer` service is already set (which we may ease later by adding a config option for it). Or do you mean something else?

Resources/config/serializer.xml
xabbuh: I think in this case the service should already be removed as it is an unused private service then (IIRC we have functional tests that cover this case).

xabbuh:
@lsmith77 I agree with you that we cannot do that for each and every part. But I also agree with @stof that we should provide a clear and smooth upgrade path to ease the life for other bundles that depend on FOSRestBundle.

I believe defining first which parts of the bundle are internal and which expose our public API would be a big step in the right direction (everything not being part of the public API would not necessarily have to provide a smooth upgrade path). For that reason I suggest that we start marking all classes that are not meant to be used by consumer code as internal (see https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1213#issuecomment-158617633). What do you think?

xabbuh:
@Ener-Getick Ah that did the trick. What a stupid mistake. Thank you very much!

Serializer/Serializer.php
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * @author Christian Flothmann <christian.flothmann@xabbuh.de>
+ */
+interface Serializer
+{
+    /**
+     * @param mixed  $data
+     * @param string $format
+     * @param mixed  $context
```
stof: this should probably be more specific than ``mixed``, abstracting the differences

EventListener/ViewResponseListener.php
```diff
@@ -27,16 +28,22 @@
  */
 class ViewResponseListener extends TemplateListener
```
stof: this should not extend the TemplateListener anymore but just run earlier (and so no need for the container)

View/ViewHandler.php
stof: this is indeed needed to avoid fatal errors in weird cases

View/ViewHandler.php
```diff
@@ -107,17 +107,33 @@ class ViewHandler implements ConfigurableViewHandlerInterface, ContainerAwareInt
...
+     * @param ExceptionWrapperHandlerInterface $exceptionWrapperHandler An exception wrapper handler
+     * @param array                            $formats                 the supported formats as keys and if the given formats uses templating is denoted by a true value
+     * @param int                              $failedValidationCode    The HTTP response status code for a failed validation
+     * @param int                              $emptyContentCode        HTTP response status code when the view data is null
+     * @param bool                             $serializeNull           Whether or not to serialize null view data
+     * @param array                            $forceRedirects          If to force a redirect for the given key format, with value being the status code to use
+     * @param string                           $defaultEngine           default engine (twig, php ..)
      */
     public function __construct(
+        RouterInterface $router,
```
stof: We should only inject a UrlGeneratorInterface here (changing the ``getRouter``)

View/ViewHandler.php
```diff
@@ -107,17 +107,33 @@ class ViewHandler implements ConfigurableViewHandlerInterface, ContainerAwareInt
...
+        Serializer $serializer,
+        EngineInterface $templating,
```
stof: We should allow injecting Twig_Environment directly too.

stof: and should be optional

Resources/config/serializer.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" ?>
+<container xmlns="http://symfony.com/schema/dic/services"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+        <service id="fos_rest.serializer.jms" class="FOS\RestBundle\Serializer\JmsSerializerAdapter" public="false" lazy="true">
```
Ener-Getick: Must be changed to ``FOS\RestBundle\Serializer\JMSSerializerAdapter`` ;-)

Serializer/Serializer.php
Ener-Getick: As @xabbuh said [here](https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1171#issuecomment-158617224) IMO we should the context adapters behavior in the serializer adapters and change this type hint to ``ContextInterface``.
May be better to do this later in another PR as this one becomes really big.

Ener-Getick:
@xabbuh do you need help on this PR ?

xabbuh:
@Ener-Getick Rebasing was a nightmare. ;) I need to resolve some conflicts and move the serialization API to a new pull request that will go to the `1.8` branch.

Ener-Getick:
Already done in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1252 :P

View/ViewHandler.php
xabbuh: That's true but we need to fix this in previous versions.

xabbuh:
@Ener-Getick Ah right, I didn't think about that. I think we still need to simplify that a bit, but I will comment directly on your PR.

Let's see if all tests still pass and then I need to check all comments again and see if I really addressed all of your comments.

Decoder/ContainerDecoderProvider.php
```diff
@@ -11,27 +11,27 @@
...
-use Symfony\Component\DependencyInjection\ContainerAwareTrait;
+use Symfony\Component\DependencyInjection\ContainerInterface;
 
 /**
  * Provides encoders through the Symfony2 DIC.
  *
  * @author Igor Wiedler <igor@wiedler.ch>
  */
-class ContainerDecoderProvider implements DecoderProviderInterface, ContainerAwareInterface
+class ContainerDecoderProvider implements DecoderProviderInterface
```
Ener-Getick: Not sure this name is still relevant

Ener-Getick:
Ready to merge for me.
Do you want to include other changes @xabbuh?

Ener-Getick:
Status: reviewed

lsmith77:
needs a rebase

Ener-Getick:
See #1286

EventListener/ViewResponseListener.php
Ener-Getick: This is causing a bug see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1301.
Should we revert this change or force the user to always enable the ``TemplateListener`` ?

Ener-Getick: May be better to just revert this change or the ``ViewResponseListener`` will be dependant of the ``SensioExtraFrameworkBundle``

PR #1178, opened 2015-10-16
[master] csrf getExtendedType with BC for < 2.8

peshi:
This BC for getExtendedType has to be included if the master branch is going to support 2.7 and up.


stof:
you also need to add ``extended-type`` in the tag in the service definition, with the new name in it

PR #1180, opened 2015-10-18
Fix a bug

### LINKED ISSUE #1179: Non-existent parameter format_listener
Hi,

Since the update from 1.7.1 to dev-master version using Symfony 2.7, format listener is no longer working. Without any modifications of the config I now get the following error when I clear the cache :

```
[Symfony\Component\DependencyInjection\Exception\ParameterNotFoundException]
You have requested a non-existent parameter "fos_rest.format_listener.rules".
```

And when I go to the API :
```
InvalidArgumentException in ContainerBuilder.php line 868:
The service definition "fos_rest.exception_format_negotiator" does not exist.
```

Here is my config of the bundle :
```
fos_rest:
    param_fetcher_listener:    force
    body_listener:             true
    allowed_methods_listener:  true
    disable_csrf_role:         IS_AUTHENTICATED_ANONYMOUSLY
    routing_loader:
        default_format:  null
        include_format:  false
    format_listener:
        rules:
            - { path: '^/', priorities: [ 'json', 'xml' ], fallback_format: json, prefer_extension: false }
    view:
        view_response_listener: force
        formats:
            json: true
            xml:  true
    serializer:
        serialize_null:         true
```

Thank you.

Ener-Getick:
This PR fixes #1179 by adding a static variable to ``FormatListenerRulesPass`` and adds a new test to prevent this bug to happen again.


Ener-Getick:
If someone has an idea why the travis ci check fails ...

lsmith77:
@xabbuh hmm not sure if adding a static method here is really worth it .. I guess in theory we can set the parameter as we did before and remove it again here? then again not sure if this is less of a hack than a static method.

Ener-Getick:
It depends if we absolutely want to not expose the bundle internals.
I pushed another commit that uses the container, I don't know what's the best ...

lsmith77:
@stof what do we do in core in such cases when we need to pass some data to a compiler pass? use container parameters (and remove them in the pass?) or set them in the pass via a static call in the extension .. or another approach?

xabbuh:
I just had a quick look at the compiler pass and it seems to me that we could simply move this handling to the extension class.

lsmith77:
ah that would of course be best .. could be that thanks to various other refactorings its indeed no longer necessary to have a compiler pass all together

lsmith77:
actually its still needed for the auto rule for the profiler:
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/DependencyInjection/Compiler/FormatListenerRulesPass.php#L30

however maybe we could handle this via a PrependExtension approach?

xabbuh:
But the profiler rule does not depend on the parameter. So we can handle it differently.

Ener-Getick:
I tried to apply what you described but I doesn't understand how you want to use ``Extension::prepend`` @lsmith77 to move the rule for the profiler. Is this function executed after ``Extension::load()`` ?

lsmith77:
its called before the `load()`` method:
http://symfony.com/doc/current/cookbook/bundles/prepend_extension.html

this way we could check the config of the web profiler and store in a property if we need to define this rule or not.

xabbuh:
The issue with that solution is that in a prepend extension we can only prepend some config but we cannot check the configured values. Thus we should keep the handling for the profiler in the compiler pass but move our own handling to the extension.

lsmith77:
> The issue with that solution is that in a prepend extension we can only prepend some config but we cannot check the configured values. Thus we should keep the handling for the profiler in the compiler pass but move our own handling to the extension.
> 
in the prepend extension you have access to all configs of all Bundles.

but yeah it might be possible to keep the logic for that rule the compiler pass .. might be easier

xabbuh:
Are you sure? Looking at the code I don't see how the bundle's configuration should already have been processed when the prepend extensions are processed.

lsmith77:
you get the raw config indeed but you can process it then.

Ener-Getick:
It has to be tested but this should work.

DependencyInjection/FOSRestExtension.php
```diff
@@ -163,9 +164,34 @@ private function loadFormatListener(array $config, XmlFileLoader $loader, Contai
...
                 if (!isset($rule['exception_fallback_format'])) {
-                    $config['format_listener']['rules'][$key]['exception_fallback_format'] = $rule['fallback_format'];
+                    $rule['exception_fallback_format'] = $rule['fallback_format'];
+                }
+                $this->addFormatListenerRule($rule, $config, $container);
+            }
+
+            $bundles = $container->getParameter('kernel.bundles');
+            if (isset($bundles['WebProfilerBundle'])) {
+                $profilerConfig = $this->processConfiguration(new ProfilerConfiguration(), $configs);
```
xabbuh: I am not sure how this is supposed to work. At least, there should be a test for it.

lsmith77: basically this duplicates the configuration logic from ``WebProfilerBundle`` to be able to determine in the FOSRestBundle if the toolbar will be enabled or not.

xabbuh: Well, but we never get the web profiler bundle's configuration here, do we?

Ener-Getick: It seems that we can only access the ``WebProfilerBundle`` config in the ``Extension::prepend`` function. I will move this logic.

Ener-Getick:
This way, it works. I created a new functional test that checks that the profiler is accessible.

DependencyInjection/FOSRestExtension.php
```diff
@@ -37,6 +40,29 @@ public function prepend(ContainerBuilder $container)
         if ($config['view']['view_response_listener']['enabled']) {
             $container->prependExtensionConfig('sensio_framework_extra', ['view' => ['annotations' => false]]);
         }
+
+        $bundles = $container->getParameter('kernel.bundles');
+        if (isset($bundles['WebProfilerBundle'])) {
+            $profilerConfigs = $container->getExtensionConfig('web_profiler');
```
xabbuh: I am still not convinced that this is the best solution here. We would now depend on assuming how the web profiler config is processed internall which makes our code very fragile to break in case they make some changes there.

xabbuh: Imho we should split the handling or if you prefer to keep it in one place reintroduce the container parameter in FOSRestBundle.

Ener-Getick: That's right. I don't think this part of their code will change but in the case it does, it may be better to use a container parameter (https://github.com/FriendsOfSymfony/FOSRestBundle/compare/master...Ener-Getick:aafa3a7ce39e316e263fc5a5b88a344709b1238b).

We have to decide what's the best to do ...

DependencyInjection/FOSRestExtension.php
```diff
@@ -37,6 +40,29 @@ public function prepend(ContainerBuilder $container)
...
+            $profilerConfig = $this->processConfiguration(new ProfilerConfiguration(), $profilerConfigs);
+
+            if ($profilerConfig['toolbar'] || $profilerConfig['intercept_redirects']) {
+                $path = '_profiler';
+                if ($profilerConfig['toolbar']) {
+                    $path .= '|_wdt';
+                }
+
+                $this->formatListenerDefaultRules[] = [
```
lsmith77: hmm I actually wonder how this even worked before .. the order of rules matters .. so imho just appending the rule to the end is problematic as if someone as a ``^/`` rule defined, then this automatic rule will never be applied. so I guess we need to add this rule to the beginning of the array.

Ener-Getick: This is already like that https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1180/files#diff-ffb88a00253e285e14e0d32ef5423cd1R192.

lsmith77:
so which direction do we want to take? container parameter with compiler pass or prepend extension?

Ener-Getick:
I think the easiest and clearest is to use a container parameter with a compiler pass.

Ener-Getick:
Unfortunately, it seems that we can't remove a container parameter... 
So the ``FormatListenerRulesPass`` set ``fos_rest.format_listener.rules`` to ``null`` when it ends.

lsmith77:
@Ener-Getick ok thx .. merged it .. but as I stated in one of the above comments .. I think we actually have a bug here in that the compiler pass appends the rule to the end of the rule list and it should be pushed to the front of the rule list. this bug fix should however then go to 1.7.

Ener-Getick:
I do not understand, in the compiler pass the profiler role is added at the end but before all the other rules are added so it's the first rule...

lsmith77:
ah .. you are right .. all is well :)

PR #1181, opened 2015-10-18
Array validator - pass value with key

piotrantosik:
Hi!

I try to use this query param constraint validator with example filters:

```php
[
    'tags' => [1, 2],
    'name' => 'Some name'
]
```
From doc:
```@QueryParam(array=true, name="filters", requirements=@MyComplexConstraint, description="List of complex filters")```

But currently MyComplexConstraint to validator is passed only value - there is no possibility of checking a given key and value.
This PR allow to test key with value in constraint, so i have a option to test e.g. available tags or name in filters.

### DIFF
--- a/Controller/Annotations/Param.php
+++ b/Controller/Annotations/Param.php
@@ -34,6 +34,8 @@ abstract class Param
     /** @var bool */
     public $array = false;
     /** @var bool */
+    public $withKeys = false;
+    /** @var bool */
     public $nullable = false;
     /** @var bool */
     public $allowBlank = true;
--- a/Request/ParamFetcher.php
+++ b/Request/ParamFetcher.php
@@ -147,8 +147,13 @@ class ParamFetcher implements ParamFetcherInterface
             }
 
             $self = $this;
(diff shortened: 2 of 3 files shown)


lsmith77:
you need to handle the fact that 2.3 doesnt support the ``key`` option. also please add a note to the docs. thanks

Ener-Getick:
Shoud we really continue to add features to the 1.7 branch ?

piotrantosik:
I have not checked version 2.0. I will try to test as soon as possible if it is ok then this pr is pointless.

lsmith77:
@Ener-Getick you have a good point here .. better target 2.0 for this ..

lsmith77:
@piotrantosik can you add some docs, fix the 2.3 issue and open a new PR for master?

Ener-Getick:
I don't think this is needed for the master branch the code allows to pass the raw value to the constraints.

piotrantosik:
I will try to check this functionality on master branch at the weekend, sorry I did not have free time for last week

lsmith77:
ping

PR #1182, opened 2015-10-21
added version annotation

lsmith77:
Makes it possible to set a route condition checking the matched version:

```
use FOS\RestBundle\Controller\Annotations\Version;

/**
 * @Version("v2")
 */
class MyController
{
```

will add a condition to all routes, ie.:
```
my_route:
    ...
    condition: "request.attributes.get('version') == 'v2'"
```

### DIFF
--- a/Controller/Annotations/Version.php
+++ b/Controller/Annotations/Version.php
@@ -1,0 +1,24 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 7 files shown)


hacfi:
Very cool idea :+1:

Ener-Getick:
At the same time, we should probably refactor the version listener. Actually, it only supports the versioned mime-types.
It may be great to supports different version providers. (http://www.troyhunt.com/2014/02/your-api-versioning-is-wrong-which-is.html)

There are several versioning ways:
- the uri ``/v1/users``
- a query parameter ``/users?version=v1``
- a custom header ``X-Accept-Version: v1``
- a custom mime-type containing the version ``Accept: application/json.v2``

We can create a new ``VersionResolver`` which would try to resolve the version from this different channels.

lsmith77:
sure. sounds interesting as well

Issue #890, opened 2014-10-20
[META] API versioning 


lsmith77:
I am closing all previous tickets/PRs related to versioning that were open at the time this ticket was opened and simply linking them here for easier access:
https://github.com/FriendsOfSymfony/FOSRestBundle/pull/136
https://github.com/FriendsOfSymfony/FOSRestBundle/issues/329
https://github.com/FriendsOfSymfony/FOSRestBundle/pull/561
https://github.com/FriendsOfSymfony/FOSRestBundle/issues/610
https://github.com/FriendsOfSymfony/FOSRestBundle/pull/632
https://github.com/FriendsOfSymfony/FOSRestBundle/issues/638

Furthermore I would like to keep future discussions here as well.

smilesrg:
@lsmith77 Is there any progress on this? API versioning is a really must-have feature!

lsmith77:
the current state is this:
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/1.5.0-RC3/Resources/doc/3-listener-support.md#media-type-version-extraction

PR #1183, opened 2015-10-21
Support versioning from different channels

Ener-Getick:
This PR adds the support of the versioning from different channels:
- ~~the uri ``/v1/users``~~
- a query parameter ``/users?version=v1``
- a custom header ``X-Accept-Version: v1``
- an accept header ``Accept: application/vnd.app.v1+json``

Should we create a new section in the configuration ? Currently, there is only one way to use the api versioning: the mime-types (with the ``format_listener.media_type.version_regex``). If we had several ways, I think that we should change and move this part of the configuration ...

fixes https://github.com/FriendsOfSymfony/FOSRestBundle/issues/890


lsmith77:
I would say first write the docs and then we can see if it makes sense to move it to a separate section .. but I agree that the entire version stuff is currently too hidden.

Ener-Getick:
I tried to write a new part of the documentation. Can you tell me what you think about it ?

Resources/doc/version_listener.rst
```diff
@@ -0,0 +1,150 @@
...
+
+If you do not want to allow all the methods described above, you should choose which version resolver to enable:
+
+.. code-block:: yaml
+
+    # app/config/config.yml
+    fos_rest:
+        versioning:
+            enabled: true
+            resolvers:
```
lsmith77: i think it would also be good to define the order of the resolvers similar to here:
https://github.com/lunetics/LocaleBundle/blob/master/Resources/doc/index.markdown#guessers

lsmith77:
looks great!

Resources/doc/version_listener.rst
Ener-Getick: Yes, good idea ! Should we also use ``router`` instead of ``uri`` ?

Ener-Getick:
Just needs another review :)

Resources/doc/3-listener-support.rst
```diff
@@ -184,7 +196,7 @@ Listener                    Event                  Priority
 ==========================  =====================  ========
 ``MimeTypeListener``        ``kernel.request``     200
 ``FormatListener``          ``kernel.request``     34
-``VersionListener``         ``kernel.request``     33
+``VersionListener``         ``kernel.request``     31
```
lsmith77: this is a problem ..
the request listener is triggered at 32 and I understand why you now moved it to 31 because you need this for the other resolvers. however if we do this at 31, then it means that mime type based versioning will not be able to set the version as a request attribute before the routing, meaning that we loose the ability to define route conditions on the version.

as such we will need to trigger the listener twice .. once before and once after the request (ie. 32).

this however also raises the question if its even feasible to allow the configuration of multiple resolvers .. or maybe it only makes sense to allow the configuration of one resolver ... ?

Ener-Getick: Hum... Well, we can maybe remove the uri resolver, for most of the cases it only duplicates the corresponding route attribute.

Ener-Getick:
I think that we can extend the routing condition expression language (https://github.com/symfony/routing/blob/master/Matcher/UrlMatcher.php). 

I'll check this tomorrow.

lsmith77:
@Tobion might have some hints here as well .. especially as he plans to do some refactoring in core on this topic.

Ener-Getick:
I think that we can add a function to the routing expression language:
```yml
my_route:
    # ...
    condition: "matchVersion('1.0')"
```

lsmith77:
sure .. what would be nice .. so this function would then trigger the version matching earlier?

Ener-Getick:
Yes and we can even think about using a tool like ``composer/semver`` to use version constraints.

Ener-Getick:
But I don't know how we can compile this kind of function ... http://symfony.com/doc/current/components/expression_language/extending.html#registering-functions

Should we serialize the ``VersionResolverInterface`` ?

Ener-Getick:
If we remove the ``RequestAttributeResolver``, there is no problem with the ``Router`` right ?

Ener-Getick:
The last commits remove completely the uri support so only the ``Request`` attributes are used.

I also merged #1194 in this PR and add a little part in the doc.

Resources/doc/versioning.rst
```diff
@@ -0,0 +1,180 @@
+API versioning
+==============
+Why version your API ?
+----------------------
+If you want to introduce breaking changes in your API, you should think about version your API. Indeed, if you introduce them on your current API, your users won't necessary upgrade their application directly and their applications won't work.
```
lsmith77: ``think about versioning your API``

``if you introduce changes to your current API, your users might not upgrade their applications right away, so versioning is important to prevent existing applications to break when such changes are made.``

Resources/doc/versioning.rst
```diff
@@ -0,0 +1,180 @@
...
+How to version your API ?
+-------------------------
+There are several ways, none is standard :
+    - Use a query parameter ``/users?version=v1``
+    - Use a custom mime-type with an ``Accept`` header ``Accept: application/json; version=1.0``
+    - Use a custom header ``X-Accept-Version: v1``
+
+The ``FOSRestBundle`` allows you to use several of them at the same time or to choose one of them.
+
+URI API versioning
```
lsmith77: i think this is now a bit confusing .. I guess I would still mention URI versioning in the list above.

Resources/doc/versioning.rst
```diff
@@ -0,0 +1,180 @@
...
+                - custom_header
+                - media_type
+
+The matched version is set as a Request attribute with the name ``version``,
+and when using JMS serializer it is also set as an exclusion strategy
+automatically in the ``ViewHandler``.
+
+If you use the method using the ``Accept`` header, to make the version mechanism working:
+
+1 - The format listener must be enabled
```
lsmith77: @WouterJ how do we usually format such lists?

Resources/doc/versioning.rst
```diff
@@ -0,0 +1,180 @@
...
+            guessing_order:
+                - query
+                - custom_header
+                - media_type
+
+The matched version is set as a Request attribute with the name ``version``,
+and when using JMS serializer it is also set as an exclusion strategy
+automatically in the ``ViewHandler``.
+
+If you use the method using the ``Accept`` header, to make the version mechanism working:
```
lsmith77: If you want to version by ``Accept`` header, you will need to do the following"

Resources/doc/versioning.rst
WouterJ: Either:
```rst
1. Item 1
2. Item 2
3. Item 3
```

Or:
```
#. Item 1
#. Item 2
#. Item 3
```

Both will render as an ordered list. If possible, use the second example.

Resources/doc/versioning.rst
```diff
@@ -0,0 +1,180 @@
...
+
+See :doc:`Format Listener <format_listener>`
+
+2 - The client must pass the requested version in his header like this :
+
+.. code-block:: yaml
+
+    Accept:application/json;version=1.0
+
+3 - You must have declared the version value in your config, otherwise it won't be catched :
```
lsmith77: ``You must configure the possible mime types for all supported versions``

(makes me wonder if we could not support version ranges like composer supports to automatically expand the possible mime types ..)

Resources/doc/3-listener-support.rst
```diff
@@ -91,6 +92,17 @@ determine the request format, rather than a file extension (like foo.json).
...
 For details see :doc:`Format Listener <format_listener>`.
 
+Versioning
+----------
+
+This listener attemps to determine the current api version from different parameters of the ``Request``:
+    - the uri ``/{version}/users``
+    - a query parameter ``/users?version=v1``
+    - an ``Accept`` header ``Accept: appication/json; version=1.0``
+    - a custom header ``X-Accept-Version: v1``
```
WouterJ: you shouldn't indent the list and use `*` as the indicator char. Also, please put an empty line between the paragraph and the list.

Resources/doc/versioning.rst
```diff
@@ -0,0 +1,180 @@
+API versioning
+==============
+Why version your API ?
```
WouterJ: please add a new line before and after each headline (before line 3 and after line 4 in this example)

Resources/doc/versioning.rst
```diff
@@ -0,0 +1,180 @@
...
+    ));
+
+And then, import it from your config.yml file:
+
+.. code-block:: yaml
+
+    imports:
+        - { resource: assets_version.php }
+
+If you have to verify if the version is correctly catched you can use something like :
```
lsmith77: I think we can remove this and instead suggest to use the profiler

Resources/doc/versioning.rst
```diff
@@ -0,0 +1,180 @@
...
+
+How to match a specific version in my routing ?
+-----------------------------------------------
+You can use conditions on your request to check for the version that was determined:
+
+.. code-block:: yaml
+
+    my_route:
+        # ...
+        condition: "request.attributes.get('version') == 'v2'"
```
lsmith77: I guess here it might be nice syntax sugar to offer a custom expression that would simplify this code

Resources/doc/versioning.rst
```diff
@@ -0,0 +1,180 @@
...
+
+.. code-block:: php
+
+    use FOS\RestBundle\Controller\Annotations\Version;
+
+    /**
+     * @Version("v2")
+     *
+     * or if you support multiple versions in this controller
+     * @Version({"v1", "v2"})
```
lsmith77: I think we should also think about how one would deal with having to different controllers for the same thing, ie. I have one ProductController for v1 and one for v2. Should we also support some convention there? like ``MyBundle/Controller/v1/FooController`` and ``MyBundle/Controller/v2/FooController`` will automatically set the version contraints in the route and name the routes accordingly ..

Resources/doc/versioning.rst
Ener-Getick: +1, might be:
```yaml
condition: "matchVersion('v1')"
```

Resources/doc/versioning.rst
Ener-Getick: I didn't use the profiler a lot ... I gave you the rights on my repository if you want to change this part.

Resources/doc/versioning.rst
Ener-Getick: It may make the code harder to maintain for a small benefit :/ 
If we find a simple solution why not (besides IMO we should reorganize the routing files).

Resources/doc/versioning.rst
Ener-Getick: Thanks and sorry it's the first documentation I wrote, I didn't know all this standards...
I'll try to read the reStructuredText doc soon.

Resources/doc/versioning.rst
Ener-Getick: If we use a pre-defined list of versions, we could use something like:
```yaml
fos_rest:
    view:
        mime_types:
            json: ['application/json;version={^1.0}']
    versioning:
        versions: ['v1', 'v1.2', 'v2']
```
And using https://github.com/composer/semver we can transform this in 
```yaml
['application/json;version=v1', 'application/json;version=v1.2']
```
for example.

lsmith77:
needs a rebase

Resources/doc/versioning.rst
lsmith77: yeah .. I also think it will add confusion .. we have a solution in the docs to make it easier to define multiple mime types quickly .. so yeah lets not bother

Resources/doc/configuration-reference.rst
```diff
@@ -119,3 +119,21 @@ Full default configuration
...
+                    enabled: true
+                    parameter_name: version
+                custom_header:
+                    enabled: true
+                    header_name: X-Accept-Version
+                media_type:
+                    enabled: true
+                    regex: /(v|version)=(?P<version>[0-9\.]+)/
+            guessing_order:
+                - uri
```
lsmith77: I guess ``uri`` needs to be removed?

Resources/doc/versioning.rst
lsmith77: the profiler shows the request attributes its quite simple to use .. I would just remove this entire hint

Resources/doc/versioning.rst
lsmith77: I guess this could also be done in a followup PR

Resources/doc/configuration-reference.rst
Ener-Getick: oops... it's a mistake

Resources/doc/versioning.rst
Ener-Getick: Sure it's just a syntax sugar.

Resources/doc/versioning.rst
Ener-Getick: It's a misunderstanding, I use this part of the profiler ^_^ 
I'll change this.

lsmith77:
@Ener-Getick I guess we can merge it now?

Ener-Getick:
Yes, I'll make the ``ExpressionLanguage`` extension later.

Resources/doc/versioning.rst
lsmith77: created https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1202 as a follow up

luispabon:
This is great stuff, thank you for the hard work. Can I ask which bundle version will see this in? 2.0?

Ener-Getick:
Yes, it will be in 2.0. You can already use this with ``2.0.x-dev`` if you want but it may have bugs and several features might change.

luispabon:
:+1: I'm working on a new API build so it will be a while until I have to worry about versioning, but will definitely need this down the line

luispabon:
I will however integrate 2.0 + versioning as soon as it's available.

Ener-Getick:
Cool, do not hesitate to give us your feedback :-)

luispabon:
Will do. This project needs a 1.0 real quick so unfortunately I don't quite have the time to experiment with this just yet, otherwise I would.

Issue #1106, opened 2015-08-29
ensure Symfony 3.0 compatibility


lsmith77:

casparcrop94:
I still can't do composer update with the FOSRestBundle, because the bundle requires an 2.* http-kernel

stof:
@casparcrop94 you need to use the ``2.0.*@dev`` versions of the bundle, as the Symfony 3 support is not yet available in released versions of the bundle (and not in the 1.7 branch either).

@lsmith77 any plan to make 1.x compatible with Symfony 3 ?

Ener-Getick:
What do you mean? all the constraints are updated in 2.0 https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/composer.json

stof:
@Ener-Getick 2.0 is a dev version, not a released version, so this is not what people are using

lsmith77:
@stof this is somewhat undecided at this point .. originally I wanted to have 2.0 release by now already. a few things came up but imho we can release very quickly if we decide to.

In the mean time this idea of a BC layer came up which would go into a 1.8 release and there we could in theory also enable 3.0 but with deprecation notices. I guess we can have a meeting on Saturday to discuss this and maybe pull in @Ener-Getick via video chat?

Ener-Getick:
Sure sorry I didn't see your message... To support symfony 3.0, we would have to create bc layers. We should discuss about this, it depends too if the 2.0 will be released soon or not.

Ener-Getick:
Why not @lsmith77, but I do not understand English as well orally as writing :/

PR #1188, opened 2015-11-07
fix Symfony 3.0

lsmith77:
fixes https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1106

### DIFF
--- a/Controller/ExceptionController.php
+++ b/Controller/ExceptionController.php
@@ -17,7 +17,8 @@ use FOS\RestBundle\View\ExceptionWrapperHandlerInterface;
 use FOS\RestBundle\View\View;
 use FOS\RestBundle\View\ViewHandler;
 use Symfony\Bundle\FrameworkBundle\Templating\TemplateReference;
-use Symfony\Component\DependencyInjection\ContainerAware;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Debug\Exception\FlattenException;
@@ -26,8 +27,10 @@ use Symfony\Component\HttpKernel\Log\DebugLoggerInterface;
 /**
  * Custom ExceptionController that uses the view layer and supports HTTP response status code mapping.
  */
-class ExceptionController extends ContainerAware
+class ExceptionController implements ContainerAwareInterface
 {
+    use ContainerAwareTrait;
+
     /**
      * Creates a new ExceptionWrapper instance that can be overwritten by a custom
      * ExceptionController class.
--- a/Tests/Controller/Annotations/AbstractParamTest.php
+++ b/Tests/Controller/Annotations/AbstractParamTest.php
@@ -28,7 +28,7 @@ class AbstractParamTest extends \PHPUnit_Framework_TestCase
     public function testInterface()
     {
         $this->assertInstanceOf('FOS\RestBundle\Controller\Annotations\ParamInterface', $this->param);
-        $this->assertInstanceOf('Symfony\Component\DependencyInjection\ContainerAware', $this->param);
+        $this->assertInstanceOf('Symfony\Component\DependencyInjection\ContainerAwareInterface', $this->param);
     }
 
     public function testDefaultValues()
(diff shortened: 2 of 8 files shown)


Tests/Request/ParamFetcherTest.php
```diff
@@ -406,7 +406,7 @@ public function testParamContainerDefinition()
         $param2->expects($this->once())
             ->method('getName')
             ->willReturn('bar');
-        $param2->expects($this->once())
+        $param2->expects($this->any())
```
lsmith77: not really sure how PHPUnit deals with traits at this point .. maybe better just remove this ``expects()`` entirely now?

Tests/Controller/Annotations/AbstractParamTest.php
```diff
@@ -28,7 +28,6 @@ public function setUp()
     public function testInterface()
     {
         $this->assertInstanceOf('FOS\RestBundle\Controller\Annotations\ParamInterface', $this->param);
-        $this->assertInstanceOf('Symfony\Component\DependencyInjection\ContainerAware', $this->param);
```
lsmith77: is there an elegant way to check for the trait? I guess one could check for the method .. then again what was the value of this assertion to begin with?

composer.json
```diff
@@ -31,7 +38,7 @@
     },
 
     "require-dev": {
-        "sensio/framework-extra-bundle":  "^3.0",
+        "sensio/framework-extra-bundle":  "dev-master as 3.0",
```
lsmith77: needs to be reverted back once SensioFrameworkExtraBundle gets a new release

composer.json
```diff
@@ -45,8 +52,7 @@
         "symfony/expression-language":    "~2.7|^3.0",
         "symfony/css-selector":           "^2.7|^3.0",
         "phpoption/phpoption":            "^1.1",
-        "jms/serializer-bundle":          "^1.0",
-        "sllh/php-cs-fixer-styleci-bridge": "^1.3"
+        "jms/serializer-bundle":          "dev-symfony_3_0 as 1.0"
```
lsmith77: needs to be reverted back once JMSSerializerBundle gets a new release after https://github.com/schmittjoh/JMSSerializerBundle/pull/492 was merged

composer.json
```diff
@@ -54,7 +60,8 @@
         "jms/serializer-bundle":          "Add support for advanced serialization capabilities, recommended, requires ^1.0",
         "symfony/expression-language":    "Add support for using the expression language in the routing, requires ^2.7|^3.0",
         "symfony/serializer":             "Add support for basic serialization capabilities and xml decoding, requires ^2.7|^3.0",
-        "symfony/validator":              "Add support for validation capabilities in the ParamFetcher, requires ^2.7|^3.0"
+        "symfony/validator":              "Add support for validation capabilities in the ParamFetcher, requires ^2.7|^3.0",
+        "sllh/php-cs-fixer-styleci-bridge": "To auto configure PHP-CS-Fixer from StyleCI config file, requires ^1.3"
```
lsmith77: can be moved back to ``require-dev`` once https://github.com/Soullivaneuh/php-cs-fixer-styleci-bridge/pull/49 is merged

Controller/ExceptionController.php
```diff
@@ -26,8 +26,10 @@
 /**
  * Custom ExceptionController that uses the view layer and supports HTTP response status code mapping.
  */
-class ExceptionController extends ContainerAware
+class ExceptionController
```
Tobion: you need to implement `ContainerAwareInterface`. Otherwise I don't thikn, the container will be set at all.

lsmith77: ah .. right

Tests/Controller/Annotations/AbstractParamTest.php
nicolas-grekas: Imho, checking for a trait is checking for an implementation detail, should be avoided

composer.json
Soullivaneuh: Why not just use your VCS?

drgomesp:
Any timelines for this to be merged? Trying to integrate it with a Sf 3.0 app.

lsmith77:
it will be merged right away once SerializerBundle abd SensioFrameworkExtraBundle releases have been made. in the mean time you can use this branch.

drgomesp:
Thanks @lsmith77

composer.json
lsmith77: I do not want to block the release because of this.

PR #1190, opened 2015-11-08
Param withKeys

lsmith77:
supersedes https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1181/files


lsmith77:
damn .. not as simple as I had hoped

Ener-Getick:
We can think about renaming the ``array`` parameter  to ``map`` because this option only encapsulate the constraints in a super ``Constraint\All``. So if ``map=false`` the raw data is passed to the constraints (an array).

lsmith77:
sounds good to me.

PR #1191, opened 2015-11-08
Rename AbstractScalarParameter::$array to ::$map

Ener-Getick:
Relates to https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1190#issuecomment-154802257

### DIFF
--- a/Controller/Annotations/AbstractScalarParam.php
+++ b/Controller/Annotations/AbstractScalarParam.php
@@ -24,7 +24,7 @@ abstract class AbstractScalarParam extends AbstractParam
     /** @var mixed */
     public $requirements = null;
     /** @var bool */
-    public $array = false;
+    public $map = false;
     /** @var bool */
     public $allowBlank = true;
 
@@ -32,8 +32,6 @@ abstract class AbstractScalarParam extends AbstractParam
     public function getConstraints()
     {
         $constraints = parent::getConstraints();
-
-        $constraints[] = new Constraints\Type(array('type' => 'scalar'));
         $requirements = $this->resolve($this->requirements);
 
         if ($this->requirements instanceof Constraint) {
@@ -58,8 +56,8 @@ abstract class AbstractScalarParam extends AbstractParam
             $constraints[] = new Constraints\NotBlank();
         }
 
-        // If an array is expected apply the constraints to each element.
-        if ($this->array) {
+        // If the user wants to map the value
+        if ($this->map) {
             $constraints = array(
                 new Constraints\All($constraints),
             );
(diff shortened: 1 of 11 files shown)


lsmith77:
would be good to add a test case and mention this in the docs

Ener-Getick:
This is just a rename, there is already a test for this. And is this https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1191/files#diff-2bf0083cfe57bb37a1bf4800c70a2733R61 not enough ?

lsmith77:
well it would be good to also test with an associative array

lsmith77:
@Ener-Getick FYI I have you write permissions on this repo .. as you know its still best to work through PRs and branches but this makes it easier to collaborate

Ener-Getick:
Thanks, so from now I have to use a branch on the official repo for my PRs?

lsmith77:
@Ener-Getick yeah that would be best. And of course you can now add tags, close/edit tickets

Ener-Getick:
Hum, I just see that there is a type check here https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Controller/Annotations/AbstractScalarParam.php#L36. 
IMO, we should not create a new option, maybe create a new class containing several constants:
```php
<?php

class ValidationStrategy {
    const RAW = 0;
    const MAP = 1;
}
```

And then we would use it in our annotation: 
```php
/**
 * @RequestParameter(array=ValidationStrategy::MAP)
 */
```
If ``array === true``, the default value would be ``ValidationStrategy::map``.

Another solution would be to just remove this type check.

Ener-Getick:
I think that the best solution is to remove this type check as we don't know what the user wants...
We can in another PR add a ``type`` validation.

lsmith77:
the build error seems to be an issue with the PHP7 setup on travis:
https://travis-ci.org/FriendsOfSymfony/FOSRestBundle/jobs/91239000

Ener-Getick:
I'll try to relaunch the Travis build tomorrow it doesn't seem to be caused by our config...

lsmith77:
build is fixed again

Ener-Getick:
Ok, I will rebase this PR tonight. 

I do not remember why I added this type check, there were nothing similar before...

stof:
I suggest providing a BC layer for this, to make it easier to migrate to the new version of the bundle. It should not be hard to keep support for the old signature of the annotation for now

Ener-Getick:
I will create a pr with bc layers for the last bc breaks.

Ener-Getick:
See #1213

Ener-Getick:
For me everything is ok. 
Can I merge this PR @lsmith77 ?

PR #1192, opened 2015-11-08
Custom constraint - pass array instead of scalar

piotrantosik:
Replace #1181

@lsmith77 @Ener-Getick pls review - if is ok, i try write some tests & doc explaining this feature

### DIFF
--- a/Controller/Annotations/AbstractScalarParam.php
+++ b/Controller/Annotations/AbstractScalarParam.php
@@ -26,6 +26,8 @@ abstract class AbstractScalarParam extends AbstractParam
     /** @var bool */
     public $array = false;
     /** @var bool */
+    public $withKeys = false;
+    /** @var bool */
     public $allowBlank = true;
 
     /** {@inheritdoc} */
@@ -33,7 +35,12 @@ abstract class AbstractScalarParam extends AbstractParam
     {
         $constraints = parent::getConstraints();
 
-        $constraints[] = new Constraints\Type(array('type' => 'scalar'));
+        if ($this->array && $this->withKeys) {
+            $constraints[] = new Constraints\Type(array('type' => 'array'));
+        } else {
+            $constraints[] = new Constraints\Type(array('type' => 'scalar'));
+        }
+
         $requirements = $this->resolve($this->requirements);
 
         if ($this->requirements instanceof Constraint) {
@@ -59,7 +66,7 @@ abstract class AbstractScalarParam extends AbstractParam
         }
 
         // If an array is expected apply the constraints to each element.
-        if ($this->array) {
+        if ($this->array && !$this->withKeys) {
             $constraints = array(
                 new Constraints\All($constraints),
             );


Ener-Getick:
I don't know the ``array`` and ``withKeys`` parameters are not really different... 
I think that just renaming ``array`` to ``map`` is clearer and it will be easier to maintain (https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1191). 

The type check may be a good idea but with something more flexible (like a ``type`` option).

lsmith77:
@piotrantosik is https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1191 a viable alternative for you?

piotrantosik:
Sorry but no, i try to explain with simple example code

Controller:
```php
/**
 * @QueryParam(array=true, name="filters", requirements=@MyComplexConstraint, description="List of complex filters")
 */
public function getUsersAction(ParamFetcher $paramFetcher)
```

Validator - MyComplexConstraintValidator:
```php
class MyComplexConstraintValidator extends ConstraintValidator
{
    public function validate($value, Constraint $constraint)
    {
        dump($value);
    }
}
```
Request: ```?filters[name]=rob&filters[year]=2015```
Without PR dump result:
```php
MyComplexConstraintValidator.php on line 15:
"rob"
MyComplexConstraintValidator.php on line 15:
"2015"
```
In my opinion using array param without withKey is unpractical because we do not know key/value pair of being validated. Another example, from documentation:
Requesting ```?ids[]=1&ids[]=2``` - in this case same array/map option is corectly because it validates only the value - the key remains same

With this PR and request ```?filters[name]=rob&filters[year]=2015``` in validator whe have value and key:
```php
MyComplexConstraintValidator.php on line 15:
array:2 [
  "name" => "rob"
  "year" => "2015"
]
```
This allows me to check the filter ```name``` or ```year``` is valid.
Conclusion - can use multiple filters in one array is not a good solution? :smile: 
Or maybe we should accept it and rename ```withKeys``` to for example ```mapKeys``` like in #1191?

Ener-Getick:
But I do not understand you can set ``array``/``map`` to false to have the same result...

lsmith77:
@piotrantosik can you also push a test case into this PR in a separate commit? then @Ener-Getick can pull in the test case into his PR and show how things would work there.

Ener-Getick:
With your example it would be:

```php
namespace Controller;
class Foo {
    /**
     * No array=true
     * @QueryParam(name="filters", requirements=@MyComplexConstraint, description="List of complex filters")
     */
    public function getUsersAction(ParamFetcher $paramFetcher)
}

namespace Constraint;
class MyComplexConstraintValidator extends ConstraintValidator
{
    public function validate($value, Constraint $constraint)
    {
        dump($value);
    }
}
```

With ``?filters[name]=rob&filters[year]=2015`` the constraint will dump:
```php
MyComplexConstraintValidator.php on line 15:
array:2 [
  "name" => "rob"
  "year" => "2015"
]
```

The only difference with your code is that there is no type check, but I think that a ``type`` parameter must be better.

Ener-Getick:
Ping @piotrantosik

lsmith77:
closed in favor of https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1191

PR #1207, opened 2015-11-13
FOSRestController is now a trait and not longer a class extending from the core Controller class

lsmith77:


GromNaN:
:+1: As PHP 5.4 is required.

Ener-Getick:
I'm wondering if we wouldn't be better to suffix its name with ``Trait`` and to remove ``FOS``: ``RestControllerTrait``.

lsmith77:
yeah .. was pondering this as well

Ener-Getick:
And IMO we should create a last release for the 1.* branch with bc layers to help the migration to FOSRestBundle 2.0.

lsmith77:
not sure if its worth it .. I mean we have tons of open bugs/enhancement tickets .. at this point moving to 2.0 should not take all too long .. what does make sense is to maybe invest some more time into a migration guide

UPGRADING.md
```diff
@@ -16,6 +16,7 @@ This document will be updated to list important BC breaks and behavioral changes
  * Removed ``callback_filter`` configuration option for the jsonp_handler
  * ``exception_wrapper_handler`` is now the name of a service and not the name of a class
  * removed all ``.class`` parameters, instead overwriting services via explicit Bundle configuration is preferred
+ * ControllerTrait is now a trait and not longer a class extending from the core Controller class
```
Ener-Getick: This is not clear, imo include the new and the old name would be clearer.

Controller/ControllerTrait.php
```diff
@@ -12,17 +12,23 @@
...
 
 /**
- * Base Controller for Controllers using the View functionality of FOSRestBundle.
+ * Trait for Controllers using the View functionality of FOSRestBundle.
  *
  * @author Benjamin Eberlei <kontakt@beberlei.de>
+ * @author Lukas Kahwe Smith <smith@pooteeweet.org>
  */
-abstract class FOSRestController extends Controller
+trait ControllerTrait
```
Ener-Getick: I think that this trait must extend the ``ContainerAwareTrait`` as it can't work without it.

Ener-Getick: Another solution would be to add an abstract method ``get``.

lsmith77: well the trait falls back to using the magic ``get()`` method .. but it can work without it .. 
I am not quite sure how to best approach this .. maybe provide two traits .. ?

Ener-Getick: Oops look too fast... 
If there are two traits, the problem will be the same (moreover ``get`` belongs to ``Controller``) ... 
You can maybe check if the ``get`` method exists on ``$this`` or if ``$this`` is an instance of ``Controller``.

Controller/ControllerTrait.php
```diff
@@ -83,6 +89,30 @@ protected function routeRedirectView($route, array $parameters = [], $statusCode
...
+
+        return $this->viewhandler;
+    }
+
+    /**
+     * Get the ViewHandler.
+     *
+     * @@param View $view
+     */
+    protected function setViewHandler(ViewHandlerInterface $viewhandler)
```
Ener-Getick: Why did you create this function?

lsmith77: because i wanted this trait also to be useful when not using the ContainerAwareTrait. will send a PR in a few mins that splits that use case into two separate traits

Ener-Getick: Extending ``getViewHandler`` is already possible and I'm not sure that the setter will be used a lot.

lsmith77: its useful if you define controllers as services

Controller/ControllerTrait.php
lsmith77: i fixed it by using ``$this->container->get()``

Ener-Getick: :+1:

Controller/ControllerTrait.php
```diff
@@ -83,6 +89,30 @@ protected function routeRedirectView($route, array $parameters = [], $statusCode
...
+            throw new \RuntimeException('A "ViewHandlerInterface" instance must be set when usign the FOSRestBundle "ControllerTrait".');
+        }
+
+        return $this->viewhandler;
+    }
+
+    /**
+     * Get the ViewHandler.
+     *
+     * @@param View $view
```
lsmith77: there is a typo here that I need to fix

Resources/doc/2-the-view-layer.rst
```diff
@@ -88,10 +94,12 @@ or it is possible to use lazy-loading:
 
     <?php
 
-    use FOS\RestBundle\Controller\FOSRestController;
+    use Symfony\Bundle\FrameworkBundle\Controller\Controller;
 
-    class UsersController extends FOSRestController
+    class UsersController extends Controller
     {
+        use ContainerAwareControllerTrait;
```
lsmith77: we should actually have at least one test case using this trait

Controller/ControllerTrait.php
Ener-Getick: Finally, I'm wondering if it's not too complicated to explain if there are traits... 
Can't we just check if ``$this`` is an instance of ``ContainerAwareInterface`` or if ``$this->viewHandler`` is defined and otherwise throw an exception?

Resources/doc/2-the-view-layer.rst
lsmith77: ```Symfony\Component\DependencyInjection\ContainerAware and FOS\RestBundle\Controller\ContainerAwareControllerTrait define the same property ($container) in the composition of FOS\RestBundle\Tests\Fixtures\Controller\AnnotatedNonPluralizedArticleController. This might be incompatible, to improve maintainability consider using accessor methods in traits instead. Class was composed```

Ener-Getick: You should not use the ``ContainerAwareTrait`` apparently as it is already used by ``Controller``.

lsmith77:
To fix the tests one needs to do the following, which sucks big time as it basically means the trait isn't a logical unit anymore but requires other things that I cannot express with code :-/

```
diff --git a/Controller/ContainerAwareControllerTrait.php b/Controller/ContainerAwareControllerTrait.php
index b03ecd2..d677c1d 100644
--- a/Controller/ContainerAwareControllerTrait.php
+++ b/Controller/ContainerAwareControllerTrait.php
@@ -12,7 +12,6 @@
 namespace FOS\RestBundle\Controller;
 
 use FOS\RestBundle\View\ViewHandlerInterface;
-use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 
 /**
  * Trait for Controllers using the View functionality of FOSRestBundle.
@@ -22,7 +21,6 @@ use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 trait ContainerAwareControllerTrait
 {
     use ControllerTrait;
-    use ContainerAwareTrait;
 
     /**
      * Get the ViewHandler.
```

lsmith77:
Note even if we would change the core ``ContainerAware`` class to use the ``ContainerAwareTrait`` we would still run into the same problem as can be seen here https://3v4l.org/l1qkE

lsmith77:
ok changing to use ``getContainer()`` and defining such an abstract method within ``ContrainerAwareControllerTrait`` along with the following change fixes the issue:
```
diff --git a/ContainerAware.php b/ContainerAware.php
index af977fe..080f8cb 100644
--- a/ContainerAware.php
+++ b/ContainerAware.php
@@ -18,18 +18,5 @@ namespace Symfony\Component\DependencyInjection;
  */
 abstract class ContainerAware implements ContainerAwareInterface
 {
-    /**
-     * @var ContainerInterface
-     */
-    protected $container;
-
-    /**
-     * Sets the Container associated with this Controller.
-     *
-     * @param ContainerInterface $container A ContainerInterface instance
-     */
-    public function setContainer(ContainerInterface $container = null)
-    {
-        $this->container = $container;
-    }
+    use ContainerAwareTrait;
 }
diff --git a/ContainerAwareTrait.php b/ContainerAwareTrait.php
index 57280aa..caf222a 100644
--- a/ContainerAwareTrait.php
+++ b/ContainerAwareTrait.php
@@ -32,4 +32,14 @@ trait ContainerAwareTrait
     {
         $this->container = $container;
     }
+
+    /**
+     * gets the Container associated with this Controller.
+     *
+     * @return ContainerInterface
+     */
+    public function getContainer()
+    {
+        return $this->container;
+    }
 }
```

tl;dr: trait properties should always come with a getter method

Ener-Getick:
What about doing something like https://github.com/FriendsOfSymfony/FOSRestBundle/compare/master...Ener-Getick:CONTROLLER?expand=1 ?

lsmith77:
@Ener-Getick at the end of the day it still relies on a property that is no where defined in the trait or the contract defined by the trait. will see what happens with the ticket I opened on symfony core.

Controller/ContainerAwareControllerTrait.php
```diff
@@ -0,0 +1,46 @@
...
+
+use FOS\RestBundle\View\ViewHandlerInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Trait for Controllers using the View functionality of FOSRestBundle.
+ *
+ * @author Lukas Kahwe Smith <smith@pooteeweet.org>
+ */
+trait ContainerAwareControllerTrait
```
lsmith77: refactor this to be a base class

lsmith77:
ok .. keeping FOSRestController and just introducing ControllerTrait .. ready for final review.

Controller/ControllerTrait.php
```diff
@@ -0,0 +1,118 @@
...
+
+        return $this->viewhandler;
+    }
+
+    /**
+     * Get the ViewHandler.
+     *
+     * @param ViewHandlerInterface $viewhandler
+     */
+    protected function setViewHandler(ViewHandlerInterface $viewhandler)
```
Ener-Getick: IMO this should be public to allow setter injection

Ener-Getick:
Except my last comment, sounds good to me :+1:.

Controller/ControllerTrait.php
lsmith77: agreed

lsmith77:
@Ener-Getick fixed

PR #1087, opened 2015-08-03
Allow use of placeholders as requirements in Query/RequestParam

alekitto:
See #1039 
Note: contains a BC break as the container has been injected into the ParamFetcher as a required argument

PR #1213, opened 2015-11-16
Add BC layers

Ener-Getick:
This PR adds several BC layers.
It also includes the ``FileParam`` backward compatible support.

Classes marked as deprecated:
- ``FOS\RestBundle\Controller\Annotations\Param``
- ``FOS\RestBundle\Request\AbstractRequestBodyParamConverter``
- ``FOS\RestBundle\Request\RequestBodyParamConverter20``
- ``FOS\RestBundle\Util\Codes``
- ``FOS\RestBundle\Util\FormatNegotiator``
- ``FOS\RestBundle\Util\FormatNegotiatorInterface``
- ``FOS\RestBundle\Util\MediaTypeNegotiatorInterface``
- ``FOS\RestBundle\Util\ViolationFormatter``
- ``FOS\RestBundle\Util\ViolationFormatterInterface``

At the same time I'm wondering if we should not refactor the bundle annotations to stop using public properties and instead using getters/setters...


lsmith77:
not sure what the benefit would be of adding accessors to the annotations

lsmith77:
if we do this .. then maybe it makes more sense after all to do a 1.8 release

Controller/Annotations/AbstractParam.php
```diff
@@ -0,0 +1,178 @@
...
+use Symfony\Component\DependencyInjection\ContainerAware;
+
+/**
+ * {@inheritdoc}
+ *
+ * @author Jordi Boggiano <j.boggiano@seld.be>
+ * @author Boris Guéry <guery.b@gmail.com>
+ * @author Ener-Getick <egetick@gmail.com>
+ */
+abstract class AbstractParam extends ContainerAware implements ParamInterface
```
stof: a container-aware annotation ? WTF ?

Controller/Annotations/FileParam.php
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * Represents a file that must be present.
+ *
+ * @Annotation
+ * @Target("METHOD")
+ *
+ * @author Ener-Getick <egetick@gmail.com>
+ */
+class FileParam extends AbstractParam
```
stof: This is not a BC layer, it is a new feature. So it must not be in this PR

Util/Codes.php
```diff
@@ -22,6 +22,8 @@
  *
  * @author Jordi Boggiano <j.boggiano@seld.be>
  * @author Markus Lanthaler <markus.lanthaler@gmx.net>
+ *
+ * @deprecated since 1.7, to be removed in 2.0. Use {@link \Symfony\Component\HttpFoundation\Response} instead.
```
stof: you should add a ``@trigger_error`` on top

Ener-Getick:
We can't use ``trigger_error`` or throw an exception when someone gets the value directly without using accessors ...
Moreover we can't really check the type of the value.

Maybe can someone give use a hint ?

Util/ViolationFormatter.php
```diff
@@ -11,40 +11,11 @@
...
 
-use FOS\RestBundle\Controller\Annotations\Param;
-use FOS\RestBundle\Controller\Annotations\QueryParam;
-use Symfony\Component\Validator\ConstraintViolationInterface;
-use Symfony\Component\Validator\ConstraintViolationListInterface;
+use FOS\RestBundle\Validator\ViolationFormatter as BaseViolationFormatter;
 
-class ViolationFormatter implements ViolationFormatterInterface
+/**
+ * @deprecated since 1.7, to be removed in 2.0. Use {@link BaseViolationFormatter} instead.
```
stof: using the FQCN would make the message more understandable when being logged by the DebugClassLoader

stof:
@Ener-Getick usign public properties is the way the annotation parser works

Controller/Annotations/AbstractParam.php
Ener-Getick: The ``Container`` is used to resolve the parameter in the default value. We should maybe move this to the ``ParamFetcher`` ...

Ener-Getick:
@stof not necessary, for example the ``Route`` annotation uses private properties (see https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Routing/Annotation/Route.php).

Controller/Annotations/AbstractParam.php
Ener-Getick: I remember why it was not in the ``ParamFetcher`` ... The ``requirements`` are also resolved but this cannot be done in the ``ParamFetcher``.
So we have two solutions:
- remove the ``requirements`` resolution
- let the annotation extending ``ContainerAware``

stof: injecting the container in the annotation is likely to cause issues with the annotation parser cache. This is a very bad idea IMO

stof:
@Ener-Getick because there is 2 ways to work with annotations:

- the old way using the constructor
- the newest way using public properties directly, for which the annotation parser can validate what are the supported properties (and also which one are required and the type of values)

Controller/Annotations/AbstractParam.php
Ener-Getick: This behaviour was introduced by #1087 and I transformed it to introduce the ``ParamInterface``.

Should we keep this feature @lsmith77 ?

Ener-Getick: Firstly, I will remove this in this PR as this is a BC break ...
But we have to decide what to do for the master branch.

Ener-Getick:
Ok so we can leave the annotations as is.

Should I deprecate or add BC layers for something else ?

Ener-Getick:
``ParamFetcherInterface``, ``ParamReaderInterface``, ``ViewHandlerInterface`` and ``ConfigurableViewHandlerInterface`` are IMO not really useful and there are probably not used.
Should we not remove them to make the bundle clearer ?

lsmith77:
@Ener-Getick I assume you went through the UPGRADING guide?

aside from this, here you can see all changes between master and 1.7:
https://github.com/FriendsOfSymfony/FOSRestBundle/compare/1.7...master

Ener-Getick:
I didn't think to use the github comparator ^_^
Apparently it remains the context adapters. I will create a simple support (just the context classes not the symfony serializer support).

xabbuh:
IIRC in Symfony we made some properties `protected` in annotations and added magic getters and setters to be able to trigger deprecation warnings.

By the way, what do you think about adding `@internal` annotations to a lot of classes? This would make it easier for us to break BC in the future (many classes actually should never be used directly by consumers of the bundle)?

Request/AbstractRequestBodyParamConverter.php
```diff
@@ -151,21 +155,23 @@ protected function execute(Request $request, ParamConverter $configuration)
     }
 
     /**
-     * @return DeserializationContext
+     * @return DeserializationContext|Context
      */
     protected function getDeserializationContext()
```
Ener-Getick: What about renaming these functions to ``getContext`` and ``setContext`` ?

Context/GroupableContextInterface.php
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+namespace FOS\RestBundle\Context;
+
+/**
+ * Context with groups.
+ *
+ * @author Ener-Getick <egetick@gmail.com>
+ */
+interface GroupableContextInterface extends ContextInterface
```
Ener-Getick: I'm wondering if we should not remove the sub-context interfaces and focus on the most used features.

Ener-Getick:
:+1: @xabbuh, keeping BC is sometimes really hard because of this classes...

Request/AbstractRequestBodyParamConverter.php
xabbuh: Not sure if we need them at all, when we introduce the serializer API with #1171. What do you think?

Ener-Getick: These functions allow to define the (de)serialization context for a specific view. How would we do this with only the serializer api?

xabbuh: Ah yes, you're right. Maybe we should backport the serializer API to 1.8, add our own layer for the context and rely on that then.

Ener-Getick: Yeah probably better to import the 2.0 features to 1.8 when possible and then remove all the deprecations in 2.0.

Ener-Getick:
I'll split this PR into smallest PRs, it begins to be too complicated to manage...

PR #1222, opened 2015-11-19
Add support for simple serialization context on View

J-Ben87:
Add a simple serialization context on View class so that it is possible to use it with Symfony Serializer.


J-Ben87:
Oops, wrong branch...

PR #1223, opened 2015-11-19
Add support for simple serialization context on View

J-Ben87:
Add a simple serialization context on View class so that it is possible to use it with Symfony Serializer

### DIFF
--- a/View/View.php
+++ b/View/View.php
@@ -42,7 +42,7 @@ class View
     /**
      * @var array
      */
-    private $simpleSerializationContext = [];
+    private $simpleSerializationContext = array();
 
     /**
      * @var Response
--- a/View/View.php
+++ b/View/View.php
@@ -40,6 +40,11 @@ class View
     private $serializationContext;
 
     /**
+     * @var array
+     */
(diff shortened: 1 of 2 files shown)


Ener-Getick:
I'm not sure that this is a good solution, this class would have two methods to set the serialization context and this new method would be deleted in 2.0 so instantly deprecated... 

I'm actually creating BC layers in #1213 to help people to migrate to 2.0. The 2.0 will probably be released soon so except if you can't I suggest you to wait.

lsmith77:
yeah .. I do not think it makes sense for us to add this to 1.7 at this point. That being said, until we release 2.0 stable it might make sense for some users to adopt this.

J-Ben87:
@Ener-Getick I completely understand your point, but the solution I implemented is made not to break anything, simply allow users from 1.7 to use the context through a little workaround.
The decision is up to you anyway ;)

lsmith77:
closing in favor of https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1245

PR #1229, opened 2015-11-24
Ignoring PSR7 request typehint in route loader

eXtreme:
I'm playing with symfony psr7 bridge and using `ServerRequestInterface` typehint instead of Symfony's `Request` in actions caused incorrect route generation. Collection routes get suffixed with `/{request}` param, item routes as well, when the `$request` param was the first.

This PR fixes that by adding `Psr\Http\Message\ServerRequestInterface` to ignored classed in `RestActionReader`. I also added a simple test case for PSR's and Symfony's request.

### DIFF
--- a/Routing/Loader/Reader/RestActionReader.php
+++ b/Routing/Loader/Reader/RestActionReader.php
@@ -414,6 +414,7 @@ class RestActionReader
             'FOS\RestBundle\Request\ParamFetcherInterface',
             'Symfony\Component\Validator\ConstraintViolationListInterface',
             'Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter',
+            'Psr\Http\Message\MessageInterface',
         ];
 
         $arguments = [];
--- a/Tests/Fixtures/Controller/TypeHintedController.php
+++ b/Tests/Fixtures/Controller/TypeHintedController.php
@@ -1,0 +1,41 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\RestBundle\Tests\Fixtures\Controller;
+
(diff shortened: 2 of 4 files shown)


eXtreme:
Updated PR to support also `Psr\Http\Message\MessageInterface` and `Psr\Http\Message\RequestInterface`

lsmith77:
+1 .. that being said .. I keep wondering if we should not generally ignore any type hinted parameter?

Ener-Getick:
A solution might be to ignore any type hinted parameter except those coupled with a ``ParamConverter`` annotation.

eXtreme:
What about automatic paramconverters?

_pushed the commit once again with removed forgotten debug line_

Ener-Getick:
Sure but I think this is the best compromise we can do ... :/

eXtreme:
What about extending the list via configuration? Like

```yml
fos_rest:
    routing_loader:
        ignored_typehints:
            - Foo\Bar
```

Ener-Getick:
@lsmith77 I think we can merge this PR as long as we have not a better solution.

lsmith77:
yes :)

robhunt3r:
Please take a look at https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1232

claudiobizzotto:
I don't know the status of this issue and why this PR was reverted, but my routes come out like this (note the extra `/{request}` under POST):

```
 --------------- -------- -------- ------ ------------------------------- 
  Name            Method   Scheme   Host   Path                           
 --------------- -------- -------- ------ ------------------------------- 
  cget_thing     GET      ANY      ANY    /thing.{_format}            
  get_thing      GET      ANY      ANY    /thing/{id}.{_format}       
  post_thing     POST     ANY      ANY    /thing/{request}.{_format}  
  put_thing      PUT      ANY      ANY    /thing/{id}.{_format}       
 --------------- -------- -------- ------ -------------------------------
```

Collection routes look good when I add `\Psr\Http\Message\MessageInterface::class` to `ignoreClasses[]`. Would be nice to work/play with  PSR-7 Bridge and FOSRestBundle *together*.

xabbuh:
I think we can solve this by checking first if the interface exists (see #1597).

PR #1230, opened 2015-11-24
[WIP] Refactor the exception management

Ener-Getick:
This PR attempts to refactor the exception management to allow the user to change the serialized output.

### DIFF
--- a/EventListener/ExceptionListener.php
+++ b/EventListener/ExceptionListener.php
@@ -1,0 +1,32 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\RestBundle\EventListener;
+
+use Symfony\Component\HttpKernel\EventListener\ExceptionListener as BaseExceptionListener;
+
+class ExceptionListener extends BaseExceptionListener
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function duplicateRequest(\Exception $exception, Request $request)
+    {
+        $attributes = array(
+           '_controller' => $this->controller,
+           'exception' => $exception,
+        );
+        $request = $request->duplicate(null, null, $attributes);
+        $request->setMethod('GET');
+
+        return $request;
+    }
+}
--- a/Normalizer/ExceptionNormalizerInterface.php
+++ b/Normalizer/ExceptionNormalizerInterface.php
@@ -1,0 +1,40 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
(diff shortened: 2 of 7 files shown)


Ener-Getick:
Another solution would be to add normalizers on top of the serializers (and not only for the exceptions) to avoid having different output for common classes...

Normalizer/ExceptionNormalizerInterface.php
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Normalizes an exception.
+ *
+ * @author Ener-Getick <egetick@gmail.com>
+ */
+interface ExceptionNormalizerInterface
+{
+    /**
+     * Normalizes an exception into a set of arrays/scalars.
+     public function normalize($exception);
```
stloyd: Missing `*`.

Normalizer/SimpleExceptionNormalizer.php
```diff
@@ -0,0 +1,140 @@
...
+     *
+     * @param object $exception
+     *
+     * @return int
+     */
+    private function getExceptionStatusCode($exception)
+    {
+        if ($code = $this->isSubclassOf($exception, $this->codesMap)) {
+            return $code;
+        } elseif ($exception instanceof HttpExceptionInterface) {
```
stloyd: Should be just `if` as you call `return` in `if` above.

Normalizer/SimpleExceptionNormalizer.php
```diff
@@ -0,0 +1,140 @@
...
+            return $exception->getStatusCode();
+        } else {
```
stloyd: `else` is useless as you call `return` in `if`.

Normalizer/SimpleExceptionNormalizer.php
```diff
@@ -0,0 +1,140 @@
...
+     *
+     * @param object $exception
+     *
+     * @return array
+     */
+    private function getExceptionHeaders($exception)
+    {
+        if ($exception instanceof HttpExceptionInterface) {
+            return $exception->getHeaders();
+        } else {
```
stloyd: Same here.

EventListener/ExceptionListener.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * {@inheritdoc}
+     */
+    protected function duplicateRequest(\Exception $exception, Request $request)
+    {
+        $attributes = array(
+           '_controller' => $this->controller,
+           'exception' => $exception,
+       );
```
stloyd: Wrong indent.

Ener-Getick:
Thanks @stloyd but this is not a definitive api. I'm wondering if we should not manage this differently on top of the new serializer adapters introduced by @xabbuh in 651714c94a7c2c61dcc83cff7e7d0410714253fa.

lsmith77:
can you also look into incorporating the changes I did here https://github.com/FriendsOfSymfony/FOSRestBundle/commit/93758654c0ac2ed29576c0dcbe7d42122d71bf37 ?

Ener-Getick:
I'm not sure about this PR.... might be better to provide different normalizers for each serializer. 
I'll search if we can provide a standard normalizer which will be adapted for each serializer then.

Issue #1197, opened 2015-11-10
Make tests fail on deprecation notices ?


Ener-Getick:
Actually, the deprecations notices don't make the Travis build fail.
(https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/.travis.yml#L5)

To avoid as much as possible to have deprecations in the bundle, IMO we should remove this env parameter.

/cc @lsmith77 @stof

stof:
this has been added temporarily because Symfony stable versions currently trigger deprecation warnings of Twig 1.23

Ener-Getick:
Ok... Then we will have to remove this and to set this variable for the legacy build (https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/.travis.yml#L28) (see https://github.com/KnpLabs/KnpMenuBundle/blob/master/.travis.yml#L19 for an example).

PR #1231, opened 2015-11-25
Make tests fail on deprecation notices

Ener-Getick:
Relates to #1197

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,5 @@
 language: php
 
-env:
-    global:
-        - SYMFONY_DEPRECATIONS_HELPER: weak
-
 php:
   - 5.4
   - 5.5
@@ -26,7 +22,7 @@ matrix:
   fast_finish: true
   include:
     - php: 5.4
-      env: COMPOSER_FLAGS="--prefer-lowest"
+      env: COMPOSER_FLAGS="--prefer-lowest" SYMFONY_DEPRECATIONS_HELPER=weak
     - php: 5.6
       env: SYMFONY_VERSION='2.8.*@dev'
     - php: 5.6


Ener-Getick:
It seems that there is no longer deprecation warnings with symfony 2.7.

lsmith77:
uhm .. why is travis not running the tests?

Ener-Getick:
Probably because of [this](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/.travis.yml#L20 it was included in 07d02e5e1282475ba8d090868f7dbe406ae3c458). 

@stof do you know why travis is not running the tests for this PR ?

Ener-Getick:
The build is run twice when I remove the ``branches`` section ...

.travis.yml
```diff
@@ -17,11 +17,6 @@ cache:
   directories:
     - $HOME/.composer/cache/files
 
-branches:
-  only:
-    - master
-    - /^\d+\.\d+$/
```
stof: This is totally unrelated to the title of the PR, and it must not be removed

stof:
@Ener-Getick PR builds are not triggered when the PR is conflicting. This may have been the issue

Ener-Getick:
I removed the last commit and rebased but Travis is still not launched ...

stof:
hmm, this makes me think that Travis broke something on their end

.travis.yml
```diff
@@ -26,7 +22,7 @@ matrix:
   fast_finish: true
   include:
     - php: 5.4
-      env: COMPOSER_FLAGS="--prefer-lowest"
+      env: COMPOSER_FLAGS="--prefer-lowest" SYMFONY_DEPRECATIONS_HELPER: weak
```
stof: your Yaml is invalid (because of the ``:`` here). This is why there is no Travis build

Ener-Getick: Right stupid error....

Ener-Getick:
That's better thanks @stof

PR #1234, opened 2015-11-27
add forward compatibility and deprecation notices

Ener-Getick:
This PR adds BC layers for #1107.


Inflector/DoctrineInflector.php
```diff
@@ -0,0 +1,34 @@
...
+use Doctrine\Common\Inflector\Inflector;
+
+/**
+ * Inflector object using the Doctrine/Inflector.
+ *
+ * @author Mark Kazemier <Markkaz>
+ */
+class DoctrineInflector implements InflectorInterface
+{
+    public function __construct()
```
sstok: Is this empty constructor really needed?

Ener-Getick: Not yet I forgot to extend the old class which triggers a deprecation in its constructor.

Tests/Negotiation/FormatNegotiatorTest.php
```diff
@@ -0,0 +1,30 @@
...
+
+namespace FOS\RestBundle\Tests\Negotiation;
+
+class FormatNegotiatorTest extends \PHPUnit_Framework_TestCase
+{
+    const FORMAT_NEGOTIATOR_CLASS = 'FOS\RestBundle\Negotiation\FormatNegotiator';
+
+    protected $negotiator;
+
+    public function setUp()
```
sstok: FYI setUp() is actually defined protected in the parent class ;)

Ener-Getick:
@xabbuh @lsmith77 we should decide if we create bc layers for 1.* or if we leave it as is.
As @stof said [here](https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1171#issuecomment-159972323) IMO it's easier for everyone to upgrade a project to 2.0 with BC layers but that's also true that we lose time by creating them and that we could use this time to create new features...

IMO we should at least create BC layers for the features the most used (context, annotation properties, ...).

lsmith77:
should go to the 1.8 branch /cc @stof

lsmith77:
@stof can you modify the target branch? would like to see this merged quickly .. @Ener-Getick alternatively you can also just open a new PR

Ener-Getick:
I opened #1254

Issue #1217, opened 2015-11-18
Remove Parameter::$requirements and ::$default resolution


Ener-Getick:
Relates to https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1213#discussion_r45054184.

The ``Container`` must be injected in the ``AbstractParameter`` to resolve the default value and the ``requirements`` (see [here](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Controller/Annotations/AbstractParam.php#L98))

We should move this behaviour to the ``ParamFetcher`` but it's not possible for the ``requirements`` so IMO we should not support this anymore ...

Ener-Getick:
Another solution is to extend the constraints which needs the ``Container`` and to make them ``ContainerAware``. 
Then the ``ParamFetcher`` will inject the ``Container`` into these constraints and the ``AbstractParameter`` won't be ``ContainerAware``. 

Is this a better solution @stof ?

PR #1240, opened 2015-12-03
Resolve the container parameter in the ParamFetcher instead of in the parameters

Ener-Getick:
This is an attempt to solve #1217

### DIFF
--- a/Controller/Annotations/AbstractScalarParam.php
+++ b/Controller/Annotations/AbstractScalarParam.php
@@ -11,8 +11,10 @@
 
 namespace FOS\RestBundle\Controller\Annotations;
 
+use FOS\RestBundle\Validator\Constraints\Regex;
 use Symfony\Component\Validator\Constraint;
-use Symfony\Component\Validator\Constraints;
+use Symfony\Component\Validator\Constraints\NotBlank;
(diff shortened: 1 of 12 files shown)


Validator/Constraints/Regex.php
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+namespace FOS\RestBundle\Validator\Constraints;
+
+use FOS\RestBundle\Util\ResolverTrait;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Symfony\Component\Validator\Constraints\Regex as BaseRegex;
+
+/**
+ * @author Ener-Getick <egetick@gmail.com>
```
stof: missing ``@Annotation`` to make the class usable as annotation

Controller/Annotations/AbstractScalarParam.php
```diff
@@ -32,12 +34,12 @@
     public function getConstraints()
     {
         $constraints = parent::getConstraints();
-        $requirements = $this->resolve($this->requirements);
+        $requirements = &$this->requirements;
```
stof: does it actually need to be a reference ? This will cause reference mismatches

PR #1241, opened 2015-12-04
Solved a problem with the documentation

patxi1980:
It does not show correctly the part of the documentation of versioning, that does not show as block code


Ener-Getick:
Thank you @patxi1980

PR #1247, opened 2015-12-04
save some jobs when running builds on Travis

xabbuh:


Ener-Getick:
IMO you can also remove the tests on symfony 2.6 and 3.0 (3.0 is now the default version of symfony installed)

xabbuh:
Yeah, makes sense to remove the 3.0 build. Though I would rather like to keep the 2.6 build as that version still keeps getting security patches.

Ener-Getick:
As you want but IMO we already test the bundle with symfony 2.3 and 2.8 so it should work with 2.6

PR #1249, opened 2015-12-05
Tweak travis

lsmith77:


lsmith77:
@xabbuh / @Ener-Getick good to merge?

scrutinzer seems to not have received any info to build .. but lets not worry about that

Ener-Getick:
:+1:

.travis.yml
```diff
@@ -2,6 +2,7 @@ language: php
 
 php:
   - 5.4
+  - 5.5
   - 5.6
   - 7.0
```
xabbuh: We can remove this line to save one job.

lsmith77: but don't we then miss running php 7.0 with symfony 3.0 ?

xabbuh: We would. But is that necessary? I mean we already run Symfony 3.0 tests with other PHP versions.

lsmith77: imho its cleaner to do this ..

PR #1250, opened 2015-12-05
Fix issue with redirect

lsmith77:
not sure yet if to commit this into 1.7
also still missing a test ..


lsmith77:
@xabbuh / @Ener-Getick what do you think about this change? I think its a bug fix .. but maybe "safer" to put it into 1.8/2.0 rather than 1.7? obviously also still needs a test case.

xabbuh:
@lsmith77 I am not sure if I understand the underlying issue. Can you give an example that does not work with the current code?

EventListener/ViewResponseListener.php
```diff
@@ -111,7 +117,7 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
             }
 
             $template = $request->attributes->get('_template');
-            if ($template) {
+            if ($template && !$view->getTemplate()) {
```
lsmith77: this here just prevents overriding a manually set template

EventListener/ViewResponseListener.php
```diff
@@ -97,11 +98,16 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
             $vars = $request->attributes->get('_template_default_vars');
         }
 
+        /** @var ViewHandlerInterface $viewHandler */
         $viewHandler = $this->container->get('fos_rest.view_handler');
 
-        if ($viewHandler->isFormatTemplating($view->getFormat())) {
+        if ($viewHandler->isFormatTemplating($view->getFormat())
```
lsmith77: if we do not do this change it means that in the case of a redirect, we still set template variables, which in turn then means that instead of a redirect we trigger this code (introduced in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/451):
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/View/ViewHandler.php#L352

this issue is illustrated in the LiipHelloBundle:
https://github.com/liip/LiipHelloBundle/blob/master/Controller/ArticleController.php#L149

xabbuh:
I think we should merge this into `1.7`. We can then decide at some point if we really want to release a new `1.7` patch version or if we just include it in the new minor version (depends on how fast we progress with new versions imo).

Ener-Getick:
I agree with @xabbuh IMO this should be merged in ``1.7`` as ``1.8`` is clearly not ready yet.

lsmith77:
ok .. will work on a test case

lsmith77:
I pushed a test case

View/ViewHandler.php
```diff
@@ -430,7 +431,7 @@ public function createResponse(View $view, Request $request, $format)
     {
         $route = $view->getRoute();
         $location = $route
-            ? $this->getRouter()->generate($route, (array) $view->getRouteParameters(), true)
+            ? $this->getRouter()->generate($route, (array) $view->getRouteParameters(), UrlGeneratorInterface::ABSOLUTE_URL)
```
lsmith77: this fixes an issue since in newer versions one actually needs to pass ``0`` to get absolute urls

EventListener/ViewResponseListener.php
lsmith77: this is the only change that isn't clearly a bug fix.

Tests/Functional/Bundle/TestBundle/Resources/config/routing.yml
```diff
@@ -1,11 +1,19 @@
 request_body_param_converter:
     path:     /body-converter
-    defaults: { _controller: TestBundle:RequestBodyParamConverter:putPost }
+    defaults: { _controller: "TestBundle:RequestBodyParamConverter:putPost" }
```
Ener-Getick: why did you escape this?

Tests/Functional/Bundle/TestBundle/Resources/config/routing.yml
```diff
@@ -1,11 +1,19 @@
...
-    defaults: { _controller: TestBundle:SerializerError:exception }
+    defaults: { _controller: "TestBundle:SerializerError:exception" }
 
 test_serializer_error_invalid_form:
     path:     /serializer-error/invalid-form.{_format}
-    defaults: { _controller: TestBundle:SerializerError:invalidForm }
+    defaults: { _controller: "TestBundle:SerializerError:invalidForm" }
+
+test_view_response_listener:
+    resource: "FOS\\RestBundle\\Tests\\Functional\\Bundle\\TestBundle\\Controller\\ArticleController"
```
Ener-Getick: I personnaly prefer the shorter syntax:
```yml
resource: FOS\RestBundle\Tests\Functional\Bundle\TestBundle\Controller\ArticleController
```

Ener-Getick:
I don't know if this PR is a bug fix or an enhancement...
After all, may be safer to merge in ``1.8``.

Tests/Functional/Bundle/TestBundle/Resources/config/routing.yml
lsmith77: I got some errors .. but I think in the end it was probably something else .. so I guess this could be reverted

lsmith77:
the change in the ViewHandler is clearly a bug fix and cannot have any unwanted side effects. the change with not setting a template if one was set manually is imho also a clear bug fix and its unlikely to break any work arounds. the only questionable change is the one about not setting parameters on a redirect. so I could move that one over to a new PR on the 1.8 branch.

Ener-Getick:
I think this change has to be merged in ``1.8``.
For the rest :+1:

lsmith77:
agreed

EventListener/ViewResponseListener.php
xabbuh: @lsmith77 Should still we do this change?

PR #1251, opened 2015-12-05
Bump minimum php version to 5.5

Ener-Getick:
This PR change the minimum version of php required to use FOSRestBundle.
It also updates many class paths to the new syntax sugar ``Foo::class``

### DIFF
--- a/DependencyInjection/Compiler/ExceptionWrapperHandlerPass.php
+++ b/DependencyInjection/Compiler/ExceptionWrapperHandlerPass.php
@@ -11,6 +11,7 @@
 
 namespace FOS\RestBundle\DependencyInjection\Compiler;
 
+use JMS\Serializer\Handler\SubscribingHandlerInterface;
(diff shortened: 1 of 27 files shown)


DependencyInjection/Compiler/ExceptionWrapperHandlerPass.php
```diff
@@ -27,7 +28,7 @@ public function process(ContainerBuilder $container)
             return;
         }
 
-        if (interface_exists('JMS\Serializer\Handler\SubscribingHandlerInterface')) {
+        if (interface_exists(SubscribingHandlerInterface::class)) {
```
xabbuh: The issue with introducing changes like this right now might be that merges between branches might become trickier so we should be careful with doing them.

Ener-Getick: yeah :/ we'll see if the next merges are complicated and if they are we can still revert this changes.

PR #1252, opened 2015-12-05
Create a serializer adapter

### LINKED ISSUE #1245: create a serialization context adapter
include https://github.com/xabbuh/FOSRestBundle/commit/d9818e842ce2a8ec5dd8960d3e007e963d8240d0

Ener-Getick:
Closes #1245

### DIFF
--- a/Context/Adapter/JMSContextAdapter.php
+++ b/Context/Adapter/JMSContextAdapter.php
@@ -1,0 +1,82 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\RestBundle\Context\Adapter;
+
+use FOS\RestBundle\Context\Context;
+use JMS\Serializer\Context as JMSContext;
+use JMS\Serializer\DeserializationContext as JMSDeserializationContext;
+use JMS\Serializer\SerializationContext as JMSSerializationContext;
+
+/**
+ * BC FOSRestBundle < 2.0.
+ *
+ * @author Ener-Getick <egetick@gmail.com>
+ *
+ * @internal
+ *
+ * @todo Remove this in 2.0
+ */
+final class JMSContextAdapter
+{
+    public static function convertSerializationContext(Context $context)
+    {
+        $newContext = JMSSerializationContext::create();
+        self::fillContext($context, $newContext);
+
+        return $newContext;
+    }
+
+    public static function convertDeserializationContext(Context $context)
+    {
+        $newContext = JMSDeserializationContext::create();
+        if (null !== $context->getMaxDepth()) {
+            for ($i = 0; $i < $context->getMaxDepth(); ++$i) {
+                $newContext->increaseDepth();
+            }
+        }
+
+        self::fillContext($context, $newContext);
+
+        return $newContext;
+    }
+
+    /**
+     * Fill a jms context.
+     *
+     * @param Context    $context
+     * @param JMSContext $newContext
+     *
+     * @return JMSContext
+     */
+    private static function fillContext(Context $context, JMSContext $newContext)
+    {
+        foreach ($context->getAttributes() as $key => $value) {
+            $newContext->attributes->set($key, $value);
+        }
+
+        if (null !== $context->getVersion()) {
+            $newContext->setVersion($context->getVersion());
+        }
+        $groups = $context->getGroups();
+        if (!empty($groups)) {
+            $newContext->setGroups($context->getGroups());
+        }
+        if (null !== $context->getMaxDepth()) {
+            $newContext->enableMaxDepthChecks();
+        }
+        if (null !== $context->getSerializeNull()) {
+            $newContext->setSerializeNull($context->getSerializeNull());
+        }
+
+        return $newContext;
+    }
+}
--- a/Context/Context.php
+++ b/Context/Context.php
@@ -1,0 +1,206 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\RestBundle\Context;
+
+/**
+ * Stores the serialization or deserialization context (groups, version, ...).
+ *
+ * @author Ener-Getick <egetick@gmail.com>
+ */
+final class Context
+{
+    /**
+     * @var array
+     */
+    private $attributes = array();
+    /**
+     * @var int|null
+     */
+    private $version;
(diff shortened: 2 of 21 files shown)


Context/Adapter/JMSContextAdapter.php
```diff
@@ -0,0 +1,84 @@
...
+use JMS\Serializer\Context as JMSContext;
+use JMS\Serializer\DeserializationContext as JMSDeserializationContext;
+use JMS\Serializer\SerializationContext as JMSSerializationContext;
+
+/**
+ * @author Ener-Getick <egetick@gmail.com>
+ *
+ * @internal
+ */
+final class JMSContextAdapter
```
Ener-Getick: This class is just there for BC and must be removed in 2.0.

Ener-Getick:
@xabbuh @lsmith77 can you review this PR please?

Context/Context.php
```diff
@@ -0,0 +1,175 @@
...
+ */
+
+namespace FOS\RestBundle\Context;
+
+/**
+ * {@inheritdoc}
+ *
+ * @author Ener-Getick <egetick@gmail.com>
+ */
+class Context implements ContextInterface, GroupableContextInterface, VersionableContextInterface, MaxDepthContextInterface, SerializeNullContextInterface
```
xabbuh: I am not sure if we really need an interface for the context. Imo we can rather make this class even final. Or do you see a use case for a custom implementation of the context?

Context/Context.php
```diff
@@ -0,0 +1,175 @@
...
+        return $this->version;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function addGroup($group)
+    {
+        if (!is_string($group)) {
+            throw new \InvalidArgumentException('A normalization group must be a string.');
```
xabbuh: I don't think we should do types check at runtime. Having proper docblock contracts should be enough.

Context/Context.php
Ener-Getick: The goal was [to make it more flexible](https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1074#discussion_r33039525) for a potential 3rd serializer but I agree that we can just not consider the unsupported fields and have a single class.

Context/Context.php
Ener-Getick: :+1:

DependencyInjection/Compiler/SerializerConfigurationPass.php
```diff
@@ -35,13 +35,26 @@ public function process(ContainerBuilder $container)
...
+            if ('%' === $class[0]) {
+                $class = $container->getParameter(substr($class, 1, -1));
+            }
+
+            if (is_subclass_of($class, 'Symfony\Component\Serializer\SerializerInterface')) {
+                $container->setAlias('fos_rest.serializer', 'fos_rest.serializer.symfony');
+            } elseif (is_subclass_of($class, 'JMS\Serializer\SerializerInterface')) {
+                $container->setAlias('fos_rest.serializer', 'fos_rest.serializer.jms');
+            } else {
+                throw new \LogicException(sprintf('Cannot handle serializer of type "%s"', $class));
```
xabbuh: This will now break applications in which users managed to somehow inject their own serializer before. Do we really need this check?

EventListener/ViewResponseListener.php
```diff
@@ -75,16 +76,28 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
             if ($configuration->getStatusCode() && (null === $view->getStatusCode() || Codes::HTTP_OK === $view->getStatusCode())) {
                 $view->setStatusCode($configuration->getStatusCode());
             }
+
+            // BC < 1.8
+            $viewClass = 'FOS\RestBundle\View\View';
+            if (get_class($view) == $viewClass) {
```
xabbuh: This is always true due to the code above, isn't it?

Ener-Getick: Not if ``View`` is extended.

xabbuh: You're right, but then we don't need this special case, do we? This should all be handled the same way if we only had the code from the `else` block, shouldn't it?

Ener-Getick: Yeah this is just to improve the performance as the reflection classes are very slow.

Context/Adapter/JMSContextAdapter.php
lsmith77: I would add a ``@TODO`` item for such stuff ... even if we are going to do this change right afterwards

lsmith77:
needs a rebase

EventListener/ViewResponseListener.php
xabbuh: good argument

Context/Context.php
```diff
@@ -0,0 +1,215 @@
...
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\RestBundle\Context;
+
+/**
+ * {@inheritdoc}
```
xabbuh: This won't do anything now that the interface is dropped.

Ener-Getick: Good catch, thanks!

Request/AbstractRequestBodyParamConverter.php
```diff
@@ -151,21 +163,55 @@ protected function execute(Request $request, ParamConverter $configuration)
     }
 
     /**
+     * @return Context
+     */
+    protected function getContext()
```
Ener-Getick: I'm not sure that we should keep this method as ``Context`` is a final class...
``configureContext`` is there to customize the ``Context`` class.

Ener-Getick:
After another review, ready to be merged for me.

Ener-Getick:
Status: Reviewed

DependencyInjection/Compiler/SerializerConfigurationPass.php
```diff
@@ -35,13 +45,27 @@ public function process(ContainerBuilder $container)
...
+            $class = $container->getParameterBag()->resolveValue(
+                $container->findDefinition('serializer')->getClass()
+            );
+
+            if (is_subclass_of($class, 'Symfony\Component\Serializer\SerializerInterface')) {
+                $container->setAlias('fos_rest.serializer', 'fos_rest.serializer.symfony');
+            } elseif (is_subclass_of($class, 'JMS\Serializer\SerializerInterface')) {
+                $container->setAlias('fos_rest.serializer', 'fos_rest.serializer.jms');
+            } else {
+                @trigger_error('Support of custom serializers is deprecated since 1.8 and will be dropped in 2.0. You should create a new class implementing FOS\RestBundle\Serializer\Serializer and define it as fos_rest.serializer', E_USER_DEPRECATED);
```
xabbuh: This will also be triggered when one implements/extends our classes, won't it?

Ener-Getick: Indeed you're right, I'll change it tomorrow

Ener-Getick: Changed

PR #1253, opened 2015-12-07
fix likely typo in param fetcher doc example

francescolaffi:
guess the plus got carried over from the other examples, not really important for the explanation of the feature, still it looked wrong


Ener-Getick:
:+1:

xabbuh:
Thank you @francescolaffi.

xabbuh:
This will also be applied on `1.7` branch with #1256.

PR #1254, opened 2015-12-07
add forward compatibility and deprecation notices

Ener-Getick:
This PR adds BC layers for #1107

### DIFF
--- a/Negotiation/FormatNegotiator.php
+++ b/Negotiation/FormatNegotiator.php
@@ -1,0 +1,18 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\RestBundle\Negotiation;
+
+use FOS\RestBundle\Util\FormatNegotiator as LegacyFormatNegotiator;
+
+class FormatNegotiator extends LegacyFormatNegotiator
+{
+}
--- a/Util/FormatNegotiator.php
+++ b/Util/FormatNegotiator.php
@@ -11,11 +11,15 @@
 
 namespace FOS\RestBundle\Util;
 
+use FOS\RestBundle\Negotiation\FormatNegotiator as NewFormatNegotiator;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\RequestMatcherInterface;
 use Negotiation\FormatNegotiator as BaseFormatNegotiator;
 use Negotiation\AcceptHeader;
 
+/**
+ * @deprecated since 1.8, to be removed in 2.0. Use {@link \FOS\RestBundle\Negotiation\FormatNegotiator} instead.
+ */
 class FormatNegotiator implements MediaTypeNegotiatorInterface
 {
     private $formatNegotiator;
@@ -23,6 +27,10 @@ class FormatNegotiator implements MediaTypeNegotiatorInterface
 
     public function __construct()
     {
+        if (!$this instanceof NewFormatNegotiator) {
+            @trigger_error(sprintf('%s is deprecated since version 1.8 and will be removed in 2.0. Use FOS\RestBundle\Negotiation\FormatNegotiator instead.', __CLASS__), E_USER_DEPRECATED);
+        }
+
         $this->formatNegotiator = new BaseFormatNegotiator();
     }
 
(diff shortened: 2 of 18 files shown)


Util/Inflector/DoctrineInflector.php
```diff
@@ -17,9 +17,16 @@
...
  *
  * @author Mark Kazemier <Markkaz>
+ *
+ * @deprecated since 1.7, to be removed in 2.0. Use {@link \FOS\RestBundle\Inflector\DoctrineInflector} instead.
  */
 class DoctrineInflector implements InflectorInterface
 {
+    public function __construct()
+    {
+        @trigger_error('FOS\RestBundle\Util\Inflector\DoctrineInflector is deprecated since version 1.7 and will be removed in 2.0. Use FOS\RestBundle\Inflector\DoctrineInflector instead.', E_USER_DEPRECATED);
```
stof: instead of overwriting the constructor in the child class to avoid the warning, you can use ``if (!$this instanceof NewDoctrineInflector) { @trigger_error(...) }`` (if it is an instance of the new class, it means the code is using the new class and so things are OK)

Util/ViolationFormatter.php
```diff
@@ -12,24 +12,29 @@
 namespace FOS\RestBundle\Util;
 
 use FOS\RestBundle\Controller\Annotations\Param;
-use FOS\RestBundle\Controller\Annotations\QueryParam;
+use FOS\RestBundle\Validator\ViolationFormatter as BaseViolationFormatter;
```
stof: the alias is weird, as it is not a base class

Validator/ViolationFormatter.php
```diff
@@ -0,0 +1,54 @@
...
+use FOS\RestBundle\Controller\Annotations\Param;
+use FOS\RestBundle\Controller\Annotations\QueryParam;
+use Symfony\Component\Validator\ConstraintViolationInterface;
+use Symfony\Component\Validator\ConstraintViolationListInterface;
+
+class ViolationFormatter implements ViolationFormatterInterface
+{
+    public function __construct()
+    {
+    }
```
stof: this constructor is useless

Validator/ViolationFormatterInterface.php
```diff
@@ -0,0 +1,40 @@
...
+{
+    /**
+     * Format a Param's ConstraintViolation into a string message.
+     *
+     * @param Param                        $param
+     * @param ConstraintViolationInterface $violation
+     *
+     * @return string
+     */
+    public function format(Param $param, ConstraintViolationInterface $violation);
```
stof: you should not redefine the methods of the parent interface. This causes compatibility issues with some PHP versions. Just inherit them from the parent

Util/FormatNegotiator.php
```diff
@@ -11,134 +11,45 @@
...
 {
-    private $formatNegotiator;
-    private $map = array();
+    private $negotiator;
 
     public function __construct()
     {
-        $this->formatNegotiator = new BaseFormatNegotiator();
-    }
+        @trigger_error('FOS\RestBundle\Util\FormatNegotiator is deprecated since version 1.7 and will be removed in 2.0. Use FOS\RestBundle\Negotiation\FormatNegotiator instead.', E_USER_DEPRECATED);
```
stof: in such case, Symfony generally puts the warning outside the class declaration, to trigger it when the class is autoloaded

Util/Inflector/DoctrineInflector.php
stof: btw, you can use ``__CLASS__`` to build the message

Negotiation/FormatNegotiator.php
```diff
@@ -0,0 +1,140 @@
...
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\RestBundle\Negotiation;
+
+use FOS\RestBundle\Util\FormatNegotiator as LegacyFormatNegotiator;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\RequestMatcherInterface;
+use Negotiation\FormatNegotiator as BaseFormatNegotiator;
```
stof: the alias is weird, as it is not a base class

Util/FormatNegotiator.php
Ener-Getick: The problem is the same as for the other classes, the legacy class is extended by the new one.

Negotiation/FormatNegotiator.php
Ener-Getick: Yeah I copied the old name but I have no idea of something better...

Util/FormatNegotiator.php
stof: then, why is it instantiating a new format negotiator internally to defined the old API through decoration ? You should just keep the code here, and have the child class being empty (moving the code only when removing the deprecated class)

Ener-Getick: I did as in symfony core but it was probably a bad idea for this...

stof: does Symfony really create a new class extending the old one, and instantantiate the new class inside the old one ? Please give me a link to a place doing such weird setup

Ener-Getick: https://github.com/symfony/symfony/blob/2.8/src/Symfony/Component/Debug/Exception/FlattenException.php

stof: hmm, indeed.

Renaming classes definetely sucks

Ener-Getick:
I made the changes thank you @stof.
The review should be easier now that I use inheritance.

Issue #1244, opened 2015-12-04
turn internal methods and properties into private 


lsmith77:
follow up for https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1243

PR #1257, opened 2015-12-07
Mark many classes/methods as private/internal

Ener-Getick:
Relates to #1243 and #1244


xabbuh:
I think we should do that step in 1.8 instead so that people could already check before the next major version if there could would work well.

xabbuh:
Well, actually I would split the PR: adding the `@internal` annotation in 1.8 and changing the visibility in 2.0.

Ener-Getick:
You're right I'll create another PR for 1.8

Ener-Getick:
See #1258

Ener-Getick:
Closes #1244

PR #1260, opened 2015-12-10
Merge branch 1.7 into 1.8

Ener-Getick:


Ener-Getick:
There's a strange error on php 7.0... May be caused by ``travis``.

stof:
I saw the same error in FOSUserBundle builds. Looks like the PHP 7 install is broken

Ener-Getick:
@stof it is fixed now

PR #1263, opened 2015-12-11
  Fix a bug in the ParamFetcher

### LINKED ISSUE #1255: Bug in Version 1.7.3?
Since the scope attribute was removed from the service, the ParamFetcher does not init the params property again for the sub request, so if an exception is triggered inside the paramFetcher instead of rendering the exception response the paramfetcher throws the same exception again

Ener-Getick:
Fixed #1255

### DIFF
--- a/Tests/Functional/Bundle/TestBundle/Controller/ParamFetcherController.php
+++ b/Tests/Functional/Bundle/TestBundle/Controller/ParamFetcherController.php
@@ -1,0 +1,56 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\RestBundle\Tests\Functional\Bundle\TestBundle\Controller;
+
+use FOS\RestBundle\Controller\Annotations\RequestParam;
+use FOS\RestBundle\Controller\Annotations\QueryParam;
+use FOS\RestBundle\Request\ParamFetcherInterface;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\HttpKernelInterface;
+use Symfony\Component\Validator\Constraints\IdenticalTo;
+
+class ParamFetcherController extends Controller
+{
+    /**
+     * @RequestParam(name="raw", requirements=@IdenticalTo(value="fooraw", message="foo"), default="invalid")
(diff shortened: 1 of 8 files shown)


Tests/Functional/Bundle/TestBundle/Controller/ParamFetcherController.php
```diff
@@ -0,0 +1,57 @@
...
+        $paramsBefore = $fetcher->all();
+
+        $newRequest = new Request();
+        $newRequest->query = $request->query;
+        $newRequest->request = $request->request;
+        $newRequest->attributes->set('_controller', sprintf('%s::paramsAction', __CLASS__));
+        $response = $this->container->get('http_kernel')->handle($newRequest, HttpKernelInterface::SUB_REQUEST, false);
+
+        // FIXME: The controller is not reinitialized after a sub request.
+        // $paramsAfter = $fetcher->all(false);
```
Ener-Getick: I'm not sure if this is a regression or not and how we can fix it...

stof: I don't understand what you mean

stof: OK, found the issue: ``$this->controller`` is also scoped to the request

Ener-Getick: If we call ``ParamFetcher::getParams`` after calling ``Kernel::handle`` the params returned are those of the controller of the sub request.

Ener-Getick:
A solution might be to instantiate a ``ParamFetcher`` for each controller in the ``ParamFetcherListener`` and force the user to get it as an argument..

Tests/Functional/Bundle/TestBundle/Controller/ParamFetcherController.php
stof: exactly my point: ``$this->controller`` needs to be reset to the master one after leaving the subrequest

Ener-Getick: Yeah but the paramfetcher can also receive params dynamically... IMO the best solution is to instantiate a param fetcher for each controller in the listener and forces the user to use the request attributes.

Ener-Getick: For bc we can create a ParamFetcherStack which would receive the current ParamFetcher but it would be bugged in part.

stof:
@Ener-Getick this would mean dropping the support for the injection as a service, which is not good as a bugfix.

Ener-Getick:
We can also use a ``ParameterBag`` which would use ``spl_object_hash`` internally to match the current params but I don't know the performances of this function...

Ener-Getick:
What do you think about this new solution @stof ?

lsmith77:
looks good to me.

reviewed

lsmith77:
Status: Reviewed

lsmith77:
BTW .. just noticed that carson doesn't seen to see it when one edits a comment.

Ener-Getick:
Yeah it only works when you submit a comment (may have a weird behavior if the edited comments were supported)

Issue #1009, opened 2015-04-10
Problem with camel case serialization and login form


nschurmann:
I'm using the `fos_rest.normalizer.camel_keys` serialization, so when i receive objects in javascript with underscore they will be transformed into camel case. The problem is that the `login_check` uses `_username` and `_password` values, so symfony receives the data as `Username` and `Password`

So i'm looking for 1 of 2 options:

1. Don't serialize for `login_check`
2. Override how the data is received.

The best would be the first, because serializing the data 2 times is overhead.

lsmith77:
can you elaborate a bit more about the entire request/response data flow so that I can understand the issue a bit better?

nschurmann:
Of course,

when i send a request to the symfony2 server, the data sent in the POST request is 
    {
        '_username': 'john',
        '_password': 'mypassword'
    }

and because of the serialization, the server respond a 'bad credentials.' exception. My guess is that the option `fos_rest.normalizer.camel_keys` is "taking away" the underscore before `username` and `password`, therefore it doesn't recognise the fields which is used to work the security bundle. I'm using FOSRestBundle with FOSUserBundle.

Hope this makes things more clear.

florianv:
This is the same issue than https://github.com/FriendsOfSymfony/FOSRestBundle/issues/760.

lsmith77:
maybe the fix is just to just add a flag to ignore keys with leading underscores which is not camel case anyway

stof:
@nschurmann as a workaround until it is implemented, note that the security system allows to configure the field names

lsmith77:
https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1267

cve:
Is it in the 2.0? I still have the issue...

xabbuh:
@cve Can you provide the necessary steps to reproduce your issue?

PR #1267, opened 2015-12-15
ignore any number of leading underscores when converting from snake case to camel case

lsmith77:
ie. leave '_username' unchanged

fixes https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1009

### DIFF
--- a/Normalizer/CamelKeysNormalizer.php
+++ b/Normalizer/CamelKeysNormalizer.php
@@ -69,7 +69,7 @@ class CamelKeysNormalizer implements ArrayNormalizerInterface
      *
      * @return string
      */
-    private function normalizeString($string)
+    protected function normalizeString($string)
     {
         if (false === strpos($string, '_')) {
             return $string;
--- a/Normalizer/CamelKeysNormalizer.php
+++ b/Normalizer/CamelKeysNormalizer.php
@@ -75,8 +75,17 @@ class CamelKeysNormalizer implements ArrayNormalizerInterface
             return $string;
         }
 
-        return preg_replace_callback('/_([a-zA-Z0-9])/', function ($matches) {
+        if (preg_match('/^(_+)(.*)/', $string, $matches)) {
+            $underscorePrefix = $matches[1];
+            $string = $matches[2];
+        } else {
+            $underscorePrefix = '';
+        }
+
+        $string = preg_replace_callback('/_([a-zA-Z0-9])/', function ($matches) {
             return strtoupper($matches[1]);
         }, $string);
+
+        return $underscorePrefix.$string;
     }
 }
--- a/Normalizer/CamelKeysNormalizerWithLeadingUnderscore.php
+++ b/Normalizer/CamelKeysNormalizerWithLeadingUnderscore.php
@@ -1,0 +1,45 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\RestBundle\Normalizer;
+
+/**
+ * Normalizes the array by changing its keys from underscore to camel case, while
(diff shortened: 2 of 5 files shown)


Normalizer/CamelKeysNormalizer.php
```diff
@@ -75,8 +75,17 @@ private function normalizeString($string)
             return $string;
         }
 
-        return preg_replace_callback('/_([a-zA-Z0-9])/', function ($matches) {
+        if (preg_match('/^(_+)(.*)/', $string, $matches)) {
```
lsmith77: or should we just ignore anything with a leading underscore? or should we just ignore a single underscore?

Ener-Getick:
Why don't we use ``Serializer::deserialize`` ?

Ener-Getick:
I'm not sure that we should keep this feature in 2.0 as it's not intuitive and I don't really see the interest of converting the request parameters from snake case to camel case...

lsmith77:
reason for its existence is explained here https://florian.voutzinos.com/blog/handling-camelcase-with-fosrestbundle/

Ener-Getick:
hum... What about creating a black list instead of changing the behaviour for all params ?

lsmith77:
yeah .. I guess this can indeed break things .. I do not think that a black list is the way to go .. but we can either delay this until 1.8 or even 2.0 .. or we can provide a switch to get this new behavior.

Ener-Getick:
Well I'm still not sure this is a good solution to have an array normalizer on top of the request parameters... As we can see, it can break many things.
IMO it would be better to at least specify on each request if we want to use the normalizer (via a request matcher or an annotation in the controller) or to create a new ``FormSubmitter`` providing a method to convert the parameters from snake case to camel case.

lsmith77:
in 2.0 we have a way to enable listeners for specific paths

Ener-Getick:
Yeah, so wouldn't it be better to orientate people to 2.0 instead of changing the array normalizer behavior?

lsmith77:
well it fixes something I would consider a bug and it has been reported as a bug.
now we can say this bug was there so long it has become a "feature" that would be a BC break to fix. in this case we can offer it as a config option or we provide 2 services, ie. add a new service with the new behavior.

Ener-Getick:
I personally would prefer a new service/option. 
What do you think @xabbuh?

lsmith77:
moved it to a new class/service

Normalizer/CamelKeysNormalizerWithLeadingUnderscore.php
```diff
@@ -0,0 +1,44 @@
...
+     *
+     * @return string
+     */
+    protected function normalizeString($string)
+    {
+        if (false === strpos($string, '_')) {
+            return $string;
+        }
+
+        if (preg_match('/^(_+)(.*)/', $string, $matches)) {
```
Ener-Getick: IMO you should add a ``$`` to be sure that you match the entire string

Ener-Getick:
:+1:

Status: reviewed

Normalizer/CamelKeysNormalizerWithLeadingUnderscore.php
lsmith77: sure .. actually I am wondering if there is a more efficient way than to use a regexp here ..

Ener-Getick: Sure you can maybe use strspn http://php.net/manual/fr/function.strspn.php

Ener-Getick: You can also replace the preg_replace by an explode with an array_reduce

lsmith77: good call .. updated the code accordingly

lsmith77: yeah .. but I doubt that this will be more efficient or more readable ..

Ener-Getick: We have to benchmark it ;-)

PR #1272, opened 2015-12-21
Change docblock return value to 'static'

rokkie:
When creating a new instance using the 'static' keyword but the
docblock uses the class name, autocomplete can't figure out the late
static binding. Changing this to 'static' fixes this.

### DIFF
--- a/Controller/Annotations/Param.php
+++ b/Controller/Annotations/Param.php
@@ -42,6 +42,8 @@ abstract class Param
 
     /**
      * @return string
+     *
+     * @internal
      */
     public function getKey()
     {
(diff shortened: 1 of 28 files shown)


Ener-Getick:
IMO it should also be changed for ``createRedirect``.

Ener-Getick:
Would be better to target ``1.8`` instead of ``master``.

rokkie:
Should I create a new PR for this?

Ener-Getick:
Just rebase this one on ``1.8``, it will do the trick.

rokkie:
I rebased the branch, however, PR still seems to want to merge into master.

Ener-Getick:
I created #1273. I thought I could merge your changes with the command lines but it didn't work...

Anyway, thank you @rokkie.

PR #1273, opened 2015-12-21
Change docblock return value to 'static'

Ener-Getick:
When creating a new instance using the 'static' keyword but the
docblock uses the class name, autocomplete can't figure out the late
static binding. Changing this to 'static' fixes this.


Ener-Getick:
Closes #1272

PR #1268, opened 2015-12-18
Only touch fos_rest.converter.request_body when body_converter is enabled

ruudk:
In my setup I have `body_converter` disabled but `serializer` enabled. 

```yml
fos_rest:
    serializer:
        groups: ['Default']
```

This would throw:
```
 [Symfony\Component\DependencyInjection\Exception\InvalidArgumentException]  
 The service definition "fos_rest.converter.request_body" does not exist.  
```

This PR fixes that.

PR #1275, opened 2015-12-21
Fix a bug in FOSRestExtension

Ener-Getick:
``fos_rest.converter.request_body`` is always touched even if it doesn't exist in ``loadSerializer``.

Fix #1268


Ener-Getick:
Status: Reviewed

PR #1276, opened 2015-12-22
Removed useless interfaces

Ener-Getick:
This PR removes several interfaces:
* ParamFetcherInterface
* ViolationFormatterInterface
* ViewHandlerInterface
* ConfigurableViewHandlerInterface


lsmith77:
what is the logic here? that these classes should be extended only via inheritance?

Ener-Getick:
Yeah I don't see any use case for this interfaces. 
I opened this PR to discuss about it. (I forgot to add a message describing the issue)

xabbuh:
I think we should only remove interfaces when there is no need for custom implementations of a class. And then we should make our own implementation final.

Ener-Getick:
Well imo this classes can be extended but there is no need to provide a complete implementation.

stof:
@Ener-Getick Inheritance is generally a worse way to provide an extension point than composition (and much harder to provide BC btw).
So making inheritance the recommended extension is a big -1 from me. I would rather have an interface to allow composition and forbig inheritance.

The issue with the ViewHandler is of course that it has far too many responsibilities, making it impossible to use composition to provide useful extension points.

Ener-Getick:
Makes sense. Instead of removing them we should probably find a way to improve them.

PR #1286, opened 2015-12-28
stop injecting the container when possible

Ener-Getick:
Fork of #1171 on ``FOSRestBundle`` repository to make it easier to contribute to this PR.

> Injecting the container in your services is not a good practice. After this, there are only three classes left that rely on the container being injected: The ``RestRouteLoader`` class needs it to resolve controllers registered as services, the ``ParamFetcher`` class injects the container into container aware ``ParamInterface`` implementations, and the parent class of the ``ViewResponseListener`` uses the container.

### DIFF
--- a/View/ViewHandler.php
+++ b/View/ViewHandler.php
@@ -17,15 +17,15 @@ use FOS\RestBundle\Context\ContextInterface;
 use FOS\RestBundle\Context\GroupableContextInterface;
 use FOS\RestBundle\Context\SerializeNullContextInterface;
 use FOS\RestBundle\Context\VersionableContextInterface;
+use Symfony\Bundle\FrameworkBundle\Templating\EngineInterface;
 use Symfony\Bundle\FrameworkBundle\Templating\TemplateReference;
-use Symfony\Component\DependencyInjection\ContainerAwareInterface;
-use Symfony\Component\DependencyInjection\ContainerAwareTrait;
 use Symfony\Component\Form\FormInterface;
 use Symfony\Component\HttpFoundation\RedirectResponse;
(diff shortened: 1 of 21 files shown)


lsmith77:
looks good to me
once this is merged we can also continue https://github.com/FriendsOfSymfony/FOSRestBundle/commit/93758654c0ac2ed29576c0dcbe7d42122d71bf37

Status: Reviewed

View/ViewHandler.php
```diff
@@ -243,25 +266,13 @@ public function isFormatTemplating($format)
...
-
-    /**
      * Gets the serializer service.
      *
-     * @param View $view view instance from which the serializer should be configured
-     *
      * @return object that must provide a "serialize()" method
      */
-    protected function getSerializer(View $view = null)
+    protected function getSerializer()
```
Ener-Getick: I'm wondering if we should not just remove this function.

View/ViewHandler.php
```diff
@@ -297,11 +308,11 @@ protected function getSerializationContext(View $view)
     /**
      * Gets the templating service.
      *
-     * @return \Symfony\Bundle\FrameworkBundle\Templating\EngineInterface
+     * @return EngineInterface
      */
     protected function getTemplating()
```
Ener-Getick: same thing here

View/ViewHandler.php
lsmith77: lets not bother with this in here .. its big enough of a PR

xabbuh:
Thank you very much for rebasing @Ener-Getick!

Ener-Getick:
You're welcome ☺

PR #1287, opened 2015-12-28
Remove some getters in the ViewHandler

Ener-Getick:
See https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1286#discussion-diff-48508016

### DIFF
--- a/View/ViewHandler.php
+++ b/View/ViewHandler.php
@@ -266,16 +266,6 @@ class ViewHandler implements ConfigurableViewHandlerInterface
     }
 
     /**
-     * Gets the serializer service.
-     *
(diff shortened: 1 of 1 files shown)


lsmith77:
I see these as potential extension points then again since we now use constructor injection, there are less uses .. ie. likely anything that needs to be done could/should be done in the constructor anyway.

so I am +0 on this change

xabbuh:
I cannot think of a use case where you would want to use this method in custom code. So I am :+1: on this change. But I think we should trigger a deprecation in 1.8 and document the removal of this method in case someone really used this.

PR #1288, opened 2015-12-28
moved all twig related aspects of the ExceptionController into a new TwigExceptionController

lsmith77:


DependencyInjection/FOSRestExtension.php
```diff
@@ -345,15 +345,20 @@ private function loadException(array $config, XmlFileLoader $loader, ContainerBu
 
             if ($config['exception']['exception_controller']) {
                 $container->getDefinition('fos_rest.exception_listener')->replaceArgument(0, $config['exception']['exception_controller']);
+            } else {
+                $bundles = $container->getParameter('kernel.bundles');
+                if (isset($bundles['TwigBundle'])) {
```
Ener-Getick: You can use a shorter syntax here:
```php
if () { // ...
} elseif (isset($container->getParameter('kernel.bundles')['TwigBundle')) {
    // ...
}

**Comment by Ener-Getick** [comment]
Sounds good :+1:

Status: Reviewed

**Comment by lsmith77** [comment]
We should then do the same with the ViewHandler

File: Controller/TwigExceptionController.php [hunk 2]
```diff
@@ -0,0 +1,117 @@
...
+class TwigExceptionController extends ExceptionController
+{
+    /**
+     * @var EngineInterface
+     */
+    private $templating;
+
+    public function setTemplating(EngineInterface $templating)
+    {
+        $this->templating = $templating;
```
xabbuh: I remember one thing we talked about in #1171, but for which either did not come to a conclusion or I missed to implement that: Should we add special handling if the user doesn't have the templating component but only wants to use the `Twig_Environment` class?

Controller/TwigExceptionController.php
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    public function getTemplating()
```
xabbuh: I think we should make this method `private`.

Controller/TwigExceptionController.php
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        return $parameters;
+    }
+
+    /**
+     * Finds the template for the given format and status code.
+     *
+     * Note this method needs to be overridden in case another
+     * engine than Twig should be supported;
```
xabbuh: I don't think that this is true anymore. If we want people to be able to use other templating engines, they should not extend the `TwigExceptionController`. We should then rather create a base `TemplatingExceptionController` or something like that which defines all the base logic and has an abstracht `findTemplate()` method (so you don't depend on Twig in a way when using something different).

xabbuh:
@lsmith77 Sorry, my review overlapped with you merging the PR.

xabbuh:
see #1291

peshi:
this PR breaks working setup.

```
ServiceNotFoundException in Container.php line 264:
You have requested a non-existent service "os_rest.exception.twig_controller".
```
```
InvalidArgumentException in View.php line 218:
The template should be a string or extend TemplateReference
```

xabbuh:
@peshi @phoenixgao The issue with the non-existent `fos_rest.exception.twig_controller` service is fixed in the latest version of the `master` branch.

@peshi I'll also take a look into the other issue you reported.

xabbuh:
@peshi Are you able to provide a complete stack trace for your second issue?

peshi:
@xabbuh this is the current error I receive after pulling form master.
```
in vendor/friendsofsymfony/rest-bundle/View/View.php at line 218   -
    public function setTemplate($template)
    {
        if (!(is_string($template) || $template instanceof TemplateReference)) {
            throw new \InvalidArgumentException('The template should be a string or extend TemplateReference');
        }
        $this->template = $template;
```
```
at View ->setTemplate (object(Template)) 
in vendor/friendsofsymfony/rest-bundle/EventListener/ViewResponseListener.php at line 140   + 
at ViewResponseListener ->onKernelView (object(GetResponseForControllerResultEvent), 'kernel.view', object(TraceableEventDispatcher)) 
at call_user_func (array(object(ViewResponseListener), 'onKernelView'), object(GetResponseForControllerResultEvent), 'kernel.view', object(TraceableEventDispatcher)) 
in vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/Debug/WrappedListener.php at line 61   + 
at WrappedListener ->__invoke (object(GetResponseForControllerResultEvent), 'kernel.view', object(ContainerAwareEventDispatcher))
at call_user_func (object(WrappedListener), object(GetResponseForControllerResultEvent), 'kernel.view', object(ContainerAwareEventDispatcher)) 
in var/cache/dev/classes.php at line 1754   + 
at EventDispatcher ->doDispatch (array(object(WrappedListener)), 'kernel.view', object(GetResponseForControllerResultEvent)) 
in var/cache/dev/classes.php at line 1672   + 
at EventDispatcher ->dispatch ('kernel.view', object(GetResponseForControllerResultEvent)) 
in vendor/symfony/symfony/src/Symfony/Component/EventDispatcher/Debug/TraceableEventDispatcher.php at line 132   + 
at TraceableEventDispatcher ->dispatch ('kernel.view', object(GetResponseForControllerResultEvent)) 
in vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php at line 144   + 
at HttpKernel ->handleRaw (object(Request), '1') 
in vendor/symfony/symfony/src/Symfony/Component/HttpKernel/HttpKernel.php at line 62   + 
at HttpKernel ->handle (object(Request), '1', true) 
in vendor/symfony/symfony/src/Symfony/Component/HttpKernel/Kernel.php at line 169   + 
at Kernel ->handle (object(Request)) 
in web/app_dev.php at line 30   + 
```

xabbuh:
@peshi Thank you, this helps a lot.

PR #1290, opened 2015-12-29
update Composer cache directory on Travis CI

xabbuh:
We only need to cache downloaded packages but not metadata from Packagist.


Ener-Getick:
:+1: see https://github.com/puli/manager/pull/9#discussion_r40499480

Ener-Getick:
Status: Reviewed

PR #1291, opened 2015-12-29
introduce a base TemplatingExceptionController

xabbuh:


Controller/TwigExceptionController.php
```diff
@@ -11,40 +11,19 @@
...
-        if (!$this->templating instanceof EngineInterface) {
-            throw new \RuntimeException('No templating engine set');
-        }
-
-        return $this->templating;
-    }
-
-    /**
-     * {inheritDoc}.
+     * {@inheritdoc}
```
xabbuh: @Ener-Getick Looks like lowercasing `inheritdoc` indeed does the trick.

Controller/TemplatingExceptionController.php
```diff
@@ -0,0 +1,50 @@
...
+    ) {
+        parent::__construct($exceptionWrapperHandler, $formatNegotiator, $viewHandler, $exceptionCodes, $exceptionMessages, $showException);
+
+        $this->templating = $templating;
+    }
+
+    /**
+     * Finds the template for the given format and status code.
+     *
+     * This code is inspired by TwigBundle and should be synchronized on a regular basis
```
Ener-Getick: Should be moved in the ``TwigExceptionController``

DependencyInjection/Compiler/TwigExceptionPass.php
```diff
@@ -26,5 +26,9 @@ public function process(ContainerBuilder $container)
         if ($container->has('fos_rest.exception_listener') && $container->has('twig.exception_listener')) {
             $container->removeDefinition('twig.exception_listener');
         }
+
+        if (!$container->has('templating.engine.twig')) {
+            $container->removeDefinition('fos_rest.exception.twig_controller');
```
Ener-Getick: This service is private so I don't think that we need to remove it manually.

xabbuh: If I am not mistaken, the DependencyInjection will complain about the not existing argument before removing unused private services.

Controller/TemplatingExceptionController.php
xabbuh: good catch, updated

DependencyInjection/Compiler/TwigExceptionPass.php
Ener-Getick: Remember [this](https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1171#discussion-diff-46002142)

xabbuh: You are right. Technically this should work without removing it. I am just not sure if we should really rely on that.

Ener-Getick: This [optimization pass is always executed](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/DependencyInjection/Compiler/PassConfig.php#L59) so I think that we rely on it.

xabbuh: Yes, the thing is just if someone injects our service somewhere else (or we start do that at some point for whatever reason), it wouldn't be removed anymore and the error message would be harder to understand imo.

Ener-Getick: Ok so leave it as is, it isn't a big point and it doesn't cost anything to remove this service earlier.

Ener-Getick:
:+1: 
Status: Reviewed

DependencyInjection/Compiler/TwigExceptionPass.php
lsmith77: actually I think the "trick" with making the service private is likely causing a problem:
https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1288#issuecomment-167811904

Controller/TemplatingExceptionController.php
```diff
@@ -0,0 +1,47 @@
...
+    public function __construct(
+        ExceptionWrapperHandlerInterface $exceptionWrapperHandler,
+        FormatNegotiator $formatNegotiator,
+        ViewHandlerInterface $viewHandler,
+        array $exceptionCodes,
+        array $exceptionMessages,
+        $showException,
+        EngineInterface $templating
+    ) {
+        parent::__construct($exceptionWrapperHandler, $formatNegotiator, $viewHandler, $exceptionCodes, $exceptionMessages, $showException);
```
lsmith77: personally I prefer using setting injection for stylistic reasons when extending a class .. especially if there are so many constructor parameters already. I realize that this has the unelegant side effect of kind of binding things to a dependency injection container in order to get a properly working instance of the service.

But I will not fight over this ...

xabbuh: Well, I see one drawback with that solution: We needed to do runtime checks then to ensure that we can actually access the value which adds a bit overhead for (in my opinion) no added value. And if someone really wants to use the class outside the bundle, they will need to go the hard way to figure out which methods to call to properly set up the object.

Ener-Getick: This class can't work without an ``EngineInterface`` so IMO we should use the constructor injection.

DependencyInjection/Compiler/TwigExceptionPass.php
Ener-Getick: Yeah seems logic, the service is never injected and is private... so it is removed.
This service must be public

Controller/TemplatingExceptionController.php
stof: I'm with @xabbuh and @Ener-Getick here.

stof:
What is the benefit of this child class though ?

xabbuh:
@stof Which child class? The abstract one?

Controller/TemplatingExceptionController.php
lsmith77: well one can also make the constructor private and provide a factory method instead .. anyway .. its not a big issue for me and you all seem to think its fine .. so its good for me as well :)

stof: @lsmith77 this does not change the fact that your factory method will need to receive all deps, otherwise you have the same issue about creating an incomplete object.

stof:
@xabbuh yeah sorry, the abstract class, not the child class.

xabbuh:
@stof Imo it does not make sense for people using another templating engine than Twig to extend the `TwigExceptionController` class. I think they should use some other base class instead. So the only thing they have to do is provide a different implementation for the `findTemplate()` method.

DependencyInjection/Compiler/TwigExceptionPass.php
xabbuh: Indeed, the `ControllerResolver` class from the Symfony FrameworkBundle [lazy loads](https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/Controller/ControllerResolver.php#L66) controllers as services at runtime. For this, the service must not be `private` (similar to what applies to event listeners and form types). I opened #1296 as a hotfix to not keep the `master` branch in a broken state until we decide how to proceed with this pull request.

Ener-Getick:
Everything seems ready for me. Do you see anything else to change?

Status: Needs review

xabbuh:
I don't have anything to add. But I would like to wait for @stof to explain what he meant with his comment.

Controller/TemplatingExceptionController.php
lsmith77: @stof that is true ..

lsmith77:
looks good to me

PR #1292, opened 2015-12-29
Deprecate ViewHandler getters in favor of constructor injections

Ener-Getick:
See https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1287#issuecomment-167755837

### DIFF
--- a/View/ViewHandler.php
+++ b/View/ViewHandler.php
@@ -241,6 +241,8 @@ class ViewHandler implements ConfigurableViewHandlerInterface, ContainerAwareInt
      * Gets the router service.
      *
      * @return \Symfony\Component\Routing\RouterInterface
+     *
+     * @deprecated since 1.8, to be removed in 2.0.
      */
     protected function getRouter()
     {
@@ -253,6 +255,8 @@ class ViewHandler implements ConfigurableViewHandlerInterface, ContainerAwareInt
      * @param View $view view instance from which the serializer should be configured
      *
      * @return object that must provide a "serialize()" method
+     *
+     * @deprecated since 1.8, to be removed in 2.0.
      */
     protected function getSerializer(View $view = null)
     {
@@ -290,6 +294,8 @@ class ViewHandler implements ConfigurableViewHandlerInterface, ContainerAwareInt
      * Gets the templating service.
      *
      * @return \Symfony\Bundle\FrameworkBundle\Templating\EngineInterface
+     *
+     * @deprecated since 1.8, to be removed in 2.0.
      */
     protected function getTemplating()
     {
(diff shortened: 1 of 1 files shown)


Ener-Getick:
Should I use the same method used [here](https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1252/files#diff-afaa78b4ed090b310e8499c8ecd788f5R112) to trigger a deprecation?

xabbuh:
Isn't it sufficient to simply always trigger the deprecation when one of these methods is called?

Ener-Getick:
The bundle would trigger deprecations that can't be fixed so I don't really like that...

xabbuh:
In Symfony we had a few places where we passed a special argument to silence deprecations in a method. But what are the parts where we still call these methods? And is that really necessary?

Ener-Getick:
We always call this methods in the ``ViewHandler`` when we need to access the serializer, the router or the template engine.
I'll use the reflection because in the case these methods are extended nothing will be triggered with your method (it seems that they are only there to be extended not to be called).

Ener-Getick:
Done. 

@lsmith77 @xabbuh would be amazing to release ``1.8`` and ``2.0`` before the end of the year! We have almost finished.

xabbuh:
:+1: LGTM

View/ViewHandler.php
```diff
@@ -533,4 +546,19 @@ private function getDataFromView(View $view)
...
+    /**
+     * Triggers a deprecation if a getter is extended.
+     *
+     * @todo remove this in 2.0.
+     */
+    private function deprecateGetter($name)
+    {
+        if (is_subclass_of($this, __CLASS__)) {
+            $method = new \ReflectionMethod($this, $name);
+            if ($method->getDeclaringClass()->getName() != __CLASS__) {
```
stof: this should use strict comparison

View/ViewHandler.php
```diff
@@ -533,4 +546,19 @@ private function getDataFromView(View $view)
...
+                @trigger_error(sprintf('%s::%s() is deprecated since version 1.8 and will be removed in 2.0.', __CLASS__, $name), E_USER_DEPRECATED);
```
stof: You should mention that overwriting it is deprecated (and maybe mention ``get_class($this)`` in the message too to help finding where it happens)

Ener-Getick: Is it better?

View/ViewHandler.php
Ener-Getick: Oops good catch

lsmith77:
I guess we didn't manage that and I also think we should do an RC phase before we make the final releases. As for this PR .. fine for me .. though I do not have so much experience with managing deprecations

xabbuh:
@stof Do you see anything else we may want to change? Otherwise this look good to merge to me.

PR #1296, opened 2015-12-31
make Twig exception controller service public

xabbuh:
The controller resolver in the Symfony FrameworkBundle lazy loads
controllers as services. Thus, the Twig exception controller service
must be `public` and we must remove it in a compiler pass in case the
injected `templating.engine.twig` service definition does not exist.


Ener-Getick:
:+1:

Issue #1126, opened 2015-09-25
200 not possible for OPTIONS with empty response content


paq85:
Hi,

AFAIK OPTIONS request should return 200 with no content, just header with "Allowed" for example.
http://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html

It's not possible to do it using FOSRestBundle

Eg. following code would still result in 204 (No content) instead of 200
```php
        $headers = ['Allow' => 'GET,POST,DELETE'];
        return $this->handleView($this->view(null, Response::HTTP_OK, $headers));
```

It's because `vendor/friendsofsymfony/rest-bundle/FOS/RestBundle/View/ViewHandler.php` has following check
```php
        if (200 !== ($code = $view->getStatusCode())) {
            return $code;
        }

        return null !== $content ? Codes::HTTP_OK : $this->emptyContentCode;
```
which overrides my requested 200 to 204 (by default)

Please consider making it possible to send 200 with no content.

lsmith77:
hmm yeah .. we might be overdoing the "magic" here .. one way to solve this would be to check the http verb and limit this behavior to POST/PUT .. 

maybe the cleaner thing would be to stop defaulting to a specific status code and leaving it at ``null`` so that we know if an explicit status code is set, we should always return that, rather than doing any sort of magic. though this would be a change for 2.x only.

JHGitty:
Fixed and merged in #1299. @paq85 @lsmith77

PR #1299, opened 2016-01-03
make status code configurable even if content is empty

### LINKED ISSUE #1278: Response should not be automatically 204 if empty
The response is automatically on status code 204 if the response is empty.

```php
$view = $this->view(null, 200);
return $this->handleView($view);
```
.. results in a 204 response. I've expected status code 200.

Ma27:
resolves #1278 and #1126 

the status code in *FOS\RestBundle\View\View* is null by default (on getResponse() null will be 200 in the response object).
If the code is null and the content is null, the emptyContentErrorCode will be used.
But if the status code is not null, the given status code will be used.

### DIFF
--- a/Resources/doc/empty-content-status-code.rst
+++ b/Resources/doc/empty-content-status-code.rst
@@ -1,0 +1,27 @@
+Status code when responding with no content
+===========================================
+
+In some use cases the api should not send any content, especially when deleting (*DELETE*) or updating (*PUT* or *PATCH*) a resource.
+
+In order to provide a status code that should be sent when having no content, the configuration must be customized:
+
+.. code-block:: yaml
+
+    fost_rest:
+        view:
+            empty_content: 204
+
+If a controller returns no respone, this will be the status code.
+
+.. versionadded:: 2.0
+  Until FOSRestBundle 2.0 this code will be used even if another code is configured inside the view object!
+
+Changes in 2.0
+--------------
+
+In the ticket .. _#1278: https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1278 has been decided to make it possible for
+users to adjust a custom status code. So if the configured code is *204*, but you want to send a 200 with empty content, you simply
+have to adjust a status code to the *View* annotation or the view api of the *ControllerTrait*.
+
+Another reason for this change was that for the *OPTIONS* http verb a *200* in favour of a *204* should be sent with empty
+content as reported in .. _#1126: https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1126
--- a/Resources/doc/empty-content-status-code.rst
+++ b/Resources/doc/empty-content-status-code.rst
@@ -15,12 +15,5 @@ If you want to use another status code for empty responses, you can update your 
 .. versionadded:: 2.0
   Until FOSRestBundle 2.0 this code will be used even if another code is configured manually inside the view object!
 
-Changes in 2.0
---------------
-
-In the ticket .. _#1278: https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1278 has been decided to make it possible for
-users to adjust a custom status code. So if the configured code is *204*, but you want to send a 200 with empty content, you simply
-have to adjust a status code to the *View* annotation or the view api of the *ControllerTrait*.
-
-Another reason for this change was that for the *OPTIONS* http verb a *200* in favour of a *204* should be sent with empty
-content as reported in .. _#1126: https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1126
+If you don't want to use the default empty content status for a specific empty ``Response``, you just
+have to set a status code manually thanks to the ``@View()`` annotation or the ``View`` class.
--- a/Resources/doc/empty-content-status-code.rst
+++ b/Resources/doc/empty-content-status-code.rst
@@ -3,7 +3,7 @@ Status code when responding with no content
 
 In some use cases the api should not send any content, especially when deleting (*DELETE*) or updating (*PUT* or *PATCH*) a resource.
 
-In order to provide a status code that should be sent when having no content, the configuration must be customized:
+In order to provide a status code that should be sent when having no content, the configuration must can customized (the default value for the shown parameter would be "204"):
 
 .. code-block:: yaml
 
@@ -11,10 +11,10 @@ In order to provide a status code that should be sent when having no content, th
         view:
             empty_content: 204
 
-If a controller returns no respone, this will be the status code.
+If a controller returns an empty response, this status will be used.
 
 .. versionadded:: 2.0
-  Until FOSRestBundle 2.0 this code will be used even if another code is configured inside the view object!
+  Until FOSRestBundle 2.0 this code will be used even if another code is configured manually inside the view object!
 
 Changes in 2.0
 --------------
--- a/Resources/doc/empty-content-status-code.rst
+++ b/Resources/doc/empty-content-status-code.rst
@@ -3,15 +3,14 @@ Status code when responding with no content
 
 In some use cases the api should not send any content, especially when deleting (*DELETE*) or updating (*PUT* or *PATCH*) a resource.
 
-In order to provide a status code that should be sent when having no content, the configuration must can customized (the default value for the shown parameter would be "204"):
+By default, ``FOSRestBundle`` will send a *204* status if the response is empty.
+If you want to use another status code for empty responses, you can update your configuration file:
 
 .. code-block:: yaml
 
     fost_rest:
         view:
             empty_content: 204
-
-If a controller returns an empty response, this status will be used.
 
 .. versionadded:: 2.0
   Until FOSRestBundle 2.0 this code will be used even if another code is configured manually inside the view object!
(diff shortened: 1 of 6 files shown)


View/View.php
```diff
@@ -441,6 +483,10 @@ public function getResponse()
     {
         if (null === $this->response) {
             $this->response = new Response();
+
+            if ($code = $this->getStatusCode()) {
+                $this->response->setStatusCode($this->getStatusCode());
```
Ma27: should be *$code* actually

View/View.php
```diff
@@ -24,14 +24,54 @@
...
+     */
+    private $statusCode;
+
+    /**
+     * @var mixed[]
+     */
     private $templateData = [];
+
+    /**
+     * @var string
```
Ma27: string|null

Ma27:
ping @Ener-Getick @lsmith77

Tests/View/ViewHandlerTest.php
```diff
@@ -85,8 +85,16 @@ public function testRegisterHandleExpectsException()
...
-    {
+    public function testGetStatusCode(
+        $expected,
+        $data,
+        $isSubmitted,
+        $isValid,
+        $isSubmittedCalled,
+        $isValidCalled,
+        $noContentCode,
+        $actualStatusCode
```
Ener-Getick: can you make this default to ``null`` please to have less changes [here](https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1299/files#diff-7476d64421812cdc549a3911c73a9602L115) and to make this PR easier to review?

Ener-Getick:
please change [this](https://github.com/Ma27/FOSRestBundle/blob/bug/1278/View/View.php#L221) to ``int|null``.

View/ViewHandler.php
```diff
@@ -246,8 +246,9 @@ protected function getStatusCode(View $view, $content = null)
             return $this->failedValidationCode;
         }
 
-        if (200 !== ($code = $view->getStatusCode())) {
-            return $code;
+        $statusCode = $view->getStatusCode();
+        if (!empty($statusCode)) {
```
Ener-Getick: please replace this by:
```php
if(null === $view->getStatusCode()) {}
```

View/View.php
```diff
@@ -178,13 +218,15 @@ public function setHeaders(array $headers)
...
      *
-     * @param int $code
+     * @param int|null $code
      *
      * @return View
      */
     public function setStatusCode($code)
     {
-        $this->getResponse()->setStatusCode($code);
+        if (!empty($code)) {
```
Ener-Getick: should be ``null !== $code``

View/View.php
```diff
@@ -178,13 +218,15 @@ public function setHeaders(array $headers)
...
+            $this->statusCode = (int) $code;
```
Ener-Getick: IMO we don't need this type hint as this function must receive an integer.

View/View.php
```diff
@@ -441,6 +483,10 @@ public function getResponse()
     {
         if (null === $this->response) {
             $this->response = new Response();
+
+            if ($code = $this->getStatusCode()) {
```
Ener-Getick: ``null !== ($code = $this->getStatusCode())``

Ener-Getick:
Otherwise seems good to me :+1:

Ma27:
ok, everything should be fixed now.
Ah and thanks for the hint with the default params in phpunit's dataprovider :)

Ener-Getick:
Awesome, thank you!
Status: Reviewed

Maybe another review @lsmith77 @xabbuh ?

View/View.php
```diff
@@ -24,14 +24,54 @@
...
+     */
     private $data;
+
+    /**
+     * @var int|null
+     */
+    private $statusCode;
+
+    /**
+     * @var mixed|null
```
SofHad: array|callable|null ?

Ma27: https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1299/files#diff-5959da22e28b61bcb047ba680731fdcdR380
as you can see, that returns type hint matches the hint of this property.

xabbuh:
Wouldn't it be possible to do this in the 1.8 release?

Ma27:
I don't think so: the *FOS\RestBundle\View\View#getStatusCode()* returned in 1.7 200 if no status code was provided. In this change it returns null if no status code is provided just like @lsmith77 suggested it in #1126 . I didn't have such a usecase yet, but this class is not declared as @internal, so this change is an BC break that should not appear in minor versions.

Ma27:
ping @lsmith77 @xabbuh

lsmith77:
@Ma27 can you also look to add some information about this into the docs?

Ma27:
rebased master. Writing now some docs and the it should be ready.

Ma27:
ping @lsmith77 review the docs and then this can be hopefully merged :)

Ma27:
ok, rebased master again :)

Resources/doc/empty-content-status-code.rst
```diff
@@ -0,0 +1,27 @@
+Status code when responding with no content
+===========================================
```
lsmith77: @xabbuh can you have a look here?

Ma27:
rebased master

Resources/doc/empty-content-status-code.rst
```diff
@@ -0,0 +1,27 @@
...
+    fost_rest:
+        view:
+            empty_content: 204
+
+If a controller returns no respone, this will be the status code.
+
+.. versionadded:: 2.0
+  Until FOSRestBundle 2.0 this code will be used even if another code is configured inside the view object!
+
+Changes in 2.0
```
Ener-Getick: Not sure we need this section. May be better to move it to the ``UPGRADING.md`` file

Ma27: but I think that the ability to pass custom status codes even if the content is empty should be part of the documentation

Resources/doc/empty-content-status-code.rst
```diff
@@ -0,0 +1,27 @@
...
+
+In order to provide a status code that should be sent when having no content, the configuration must be customized:
+
+.. code-block:: yaml
+
+    fost_rest:
+        view:
+            empty_content: 204
+
+If a controller returns no respone, this will be the status code.
```
Ener-Getick: There is a typo ``response`` and I'd prefer ``an empty response, this status will be used``

Resources/doc/empty-content-status-code.rst
```diff
@@ -0,0 +1,27 @@
+Status code when responding with no content
+===========================================
+
+In some use cases the api should not send any content, especially when deleting (*DELETE*) or updating (*PUT* or *PATCH*) a resource.
+
+In order to provide a status code that should be sent when having no content, the configuration must be customized:
```
Ener-Getick: That's not true, ``204`` is the default status code, the configuration should only be used to use another status code.

Resources/doc/empty-content-status-code.rst
```diff
@@ -0,0 +1,27 @@
...
+.. code-block:: yaml
+
+    fost_rest:
+        view:
+            empty_content: 204
+
+If a controller returns no respone, this will be the status code.
+
+.. versionadded:: 2.0
+  Until FOSRestBundle 2.0 this code will be used even if another code is configured inside the view object!
```
Ener-Getick: IMO you should add ``manually``

Resources/doc/empty-content-status-code.rst
Ener-Getick: Then IMO the section should not refer to ``2.0`` and to the issue.

May be something like:
```rst
Don't use the empty content status
----------------------------------

If you don't want to use the empty content status for an specific empty``Response``, you just 
have to set a status code manually thanks to the ``@View()`` annotation or the ``View`` class.
```

Resources/doc/empty-content-status-code.rst
```diff
@@ -0,0 +1,27 @@
+Status code when responding with no content
+===========================================
+
+In some use cases the api should not send any content, especially when deleting (*DELETE*) or updating (*PUT* or *PATCH*) a resource.
+
+In order to provide a status code that should be sent when having no content, the configuration must can customized (the default value for the shown parameter would be "204"):
```
Ener-Getick: Another typo :-P
```rst
By default, ``FOSRestBundle`` will send a *204* status if the response is empty. 
If you want to use another status code for empty responses, you can update your configuration file:
```

Resources/doc/empty-content-status-code.rst
```diff
@@ -0,0 +1,27 @@
...
+
+.. code-block:: yaml
+
+    fost_rest:
+        view:
+            empty_content: 204
+
+If a controller returns an empty response, this status will be used.
```
Ener-Getick: Can be remove IMO

Ma27:
ping @lsmith77

Ener-Getick:
What about https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1299#discussion-diff-49263613?

And @lsmith77, after all it may be easier to protect the branches but disable required checks.

Ma27:
@lsmith77 is there a solution found for the issue with styleci?

Ener-Getick:
I think we have to wait for it to be fixed :-/

Did you run ``php-cs-fixer``? If yes, we can just remove the required check and merge your PR

lsmith77:
ok .. style ci issue is fixed

PR #1303, opened 2016-01-08
Fix TemplatingExceptionController

brunoroux:
I had some troubles with my API views and noticed some errors in TemplatingExceptionController. 

It's just some use statements that are missing (unless I'm missing something).


hmlb:
:+1:

Ener-Getick:
Thank you @brunoroux! 
We need a test case for this class...

Issue #1301, opened 2016-01-05
Trouble when upgrade from 1.7.7 to 044f4e0


marcobarcaro:
Hi,
I have two bundle: DemoBundle and ApiBUndle. 
I've upgraded SF from 2.7 to 2.8 and FOSRestBundle (dev-master) is not working anymore. I guess the problem is related to view_response_listener. When I use this configuration:

```yaml
fos_rest:
    param_fetcher_listener:
        enabled: true
    routing_loader:
        default_format: json
    view:
        view_response_listener:
            enabled: true
            force: false
```

I got this exception "The controller must return a response" from any controller on DemoBundle. 

 DEBUG - Notified event "kernel.view" to listener "FOS\RestBundle\EventListener\ViewResponseListener::onKernelView".
CRITICAL - Uncaught PHP Exception LogicException: "The controller must return a response (Array())." at ...\vendor\symfony\symfony\src\Symfony\Component\HttpKernel\HttpKernel.php line 157 

Maybe this is due to:
"Note that it is necessary to disable view annotations in SensioFrameworkExtraBundle so that FOSRestBundle can take over the handling. However, FOSRestBundle will do this automatically but it does not override any explicit configuration. So make sure to remove or disable the following setting:"
http://symfony.com/doc/master/bundles/FOSRestBundle/view_response_listener.html

Best regards

Ener-Getick:
Can you show us the code of one of your controller concerned please?

Ener-Getick:
And your full configuration

marcobarcaro:
Hi,
I've found what's going on.

With version 1.7.7 it works, with version 044f4e0 it doesn't work and I get this error when I go to /web/defaults:

"The template should be a string or extend TemplateReference"

 DEBUG - Notified event "kernel.controller" to listener "FOS\RestBundle\EventListener\ViewResponseListener::onKernelController".
CRITICAL - Uncaught PHP Exception InvalidArgumentException: "The template should be a string or extend TemplateReference" at ...\vendor\friendsofsymfony\rest-bundle\View\View.php line 218


This is the configuration that I used:

config.yml:

imports:
    - { resource: parameters.yml }
    - { resource: security.yml }
    - { resource: services.yml }

# Put parameters here that don't need to change on each machine where the app is deployed
# http://symfony.com/doc/current/best_practices/configuration.html#application-related-configuration
parameters:
    locale: en

framework:
    #esi:             ~
[...]

PR #1308, opened 2016-01-08
Fix an error with the ViewResponseListener

Ener-Getick:
See #1301, https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1301#issuecomment-170088363, https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1171#discussion_r49223924


Ener-Getick:
Since #1171, the ``ViewResponseListener`` do not longer extends the ``TemplateListener``. This was causing an issue because the ``TemplateListener`` converts ``_template`` to a ``TemplateReferenceInterface|string``.
This PR fixes this bug by alloying to use the ``TemplateListener`` with the ``ViewResponseListener`` and by doing the transformation internally when ``_view`` is used.

EventListener/ViewResponseListener.php
```diff
@@ -149,7 +131,6 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
     public static function getSubscribedEvents()
     {
         return array(
-            KernelEvents::CONTROLLER => array('onKernelController', -100),
             KernelEvents::VIEW => 'onKernelView',
```
Ener-Getick: We may need to have a bigger priority, needs to be changed in another PR...

kcivey:
I think the change in ViewResponseListener may now be required in 1.7 for people who upgrade SensioFrameworkExtraBundle from 3.0.12, because 3.0.13 is now guessing the template in onKernelController rather than onKernelView:
https://github.com/sensiolabs/SensioFrameworkExtraBundle/commit/82b8ad65409c99f693088ecbc09c857f01e3b3fa

xabbuh:
@kcivey see #1355

PR #1309, opened 2016-01-09
support for http verbs from RFC-2518

Ma27:
see https://tools.ietf.org/html/rfc2518 for further information about the http verbs

### DIFF
--- a/Resources/doc/5-automatic-route-generation_single-restful-controller.rst
+++ b/Resources/doc/5-automatic-route-generation_single-restful-controller.rst
@@ -67,7 +67,7 @@ Define resource actions
         {} // "unlock_users"    [UNLOCK] /users
 
         public function optionsUsersAction()
-        {} // "options_users "  [OPTIONS] /users
+        {} // "options_users"   [OPTIONS] /users
 
         public function getUsersAction()
         {} // "get_users"       [GET] /users
--- a/Resources/doc/5-automatic-route-generation_single-restful-controller.rst
+++ b/Resources/doc/5-automatic-route-generation_single-restful-controller.rst
@@ -45,44 +45,62 @@ Define resource actions
     <?php
     class UsersController
     {
+        public function copyUsersAction()
+        {} // "copy_users"      [COPY] /users
+
+        public function lockUsersAction()
+        {} // "lock_users"      [LOCK] /users
+
+        public function mkcolUsersAction()
+        {} // "mkcol_users"     [MKCOL] /users
+
+        public function moveUsersAction()
+        {} // "move_users"      [MOVE] /users
+
+        public function propfindUsersAction()
+        {} // "propfind_users"  [PROPFIND] /users
+
+        public function proppatchUsersAction()
+        {} // "proppatch_users" [PROPPATCH] /users
+
+        public function unlockUsersAction()
+        {} // "unlock_users"    [UNLOCK] /users
+
         public function optionsUsersAction()
-        {} // "options_users" [OPTIONS] /users
+        {} // "options_users "  [OPTIONS] /users
 
         public function getUsersAction()
-        {} // "get_users"     [GET] /users
+        {} // "get_users"       [GET] /users
 
         public function newUsersAction()
-        {} // "new_users"     [GET] /users/new
+        {} // "new_users"       [GET] /users/new
 
         public function postUsersAction()
-        {} // "post_users"    [POST] /users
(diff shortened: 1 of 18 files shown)


Resources/doc/5-automatic-route-generation_single-restful-controller.rst
```diff
@@ -45,44 +45,62 @@ Define resource actions
...
+
+        public function proppatchUsersAction()
+        {} // "proppatch_users" [PROPPATCH] /users
+
+        public function unlockUsersAction()
+        {} // "unlock_users"    [UNLOCK] /users
+
         public function optionsUsersAction()
-        {} // "options_users" [OPTIONS] /users
+        {} // "options_users "  [OPTIONS] /users
```
Ma27: mistyped

Ener-Getick:
I'm not sure that we need this classes:
IMO these annotations are shortcuts and should only be available for the most common http verbs.

Ma27:
but how common are verbs like "OPTIONS" or "LINK"?
Well love the way to configure my routes with those annotations and would love support for such http verbs, but if it you don't think that these methods should be present as annotations, feel free to close :)

lsmith77:
I am ok with adding these annotations .. if someone needs them I guess it proofs there is a use :)

Routing/Loader/Reader/RestActionReader.php
```diff
@@ -462,8 +480,14 @@ private function getCustomHttpMethod($httpMethod, array $resources, array $argum
...
         $annotations = array();
+        $annotationNames = array_map(
+            function ($httpVerb) {
+                return ucfirst($httpVerb);
+            },
+            $this->availableHTTPMethods
+        );
 
-        foreach (array('Route', 'Get', 'Post', 'Put', 'Patch', 'Delete', 'Link', 'Unlink', 'Head', 'Options') as $annotationName) {
+        foreach ($annotationNames as $annotationName) {
```
Ener-Getick: I think that we can improve this:
```php
foreach($this->reader->getMethodAnnotations($reflectionMethod) as $annotation) {
    if($annotation instanceof Route) {
        $annotations[] = $annotation;
    } 
} 
```

Ener-Getick:
Then I'm ok with it too 😉

Tests/Fixtures/Controller/UsersController.php
```diff
@@ -58,12 +100,6 @@ public function patchUserAction($slug)
 
  // [PATCH] /users/{slug}
 
-    public function lockUserAction($slug)
```
Ener-Getick: Why did you remove this ?

Ma27: as the lock stuff will be handled by the lock method. The patch thing is simply a fallback: https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Routing/Loader/Reader/RestActionReader.php#L525-L539

Ener-Getick:
Can you fix the tests?

Ma27:
I'm aware of some failing tests, but currently I don't have that much time, try to do this at the weekend, sry...

Ener-Getick:
Ok no problem 😉

JHGitty:
@Ma27 @lsmith77 
Maybe the classes should contain this comment:
```
RFC2518 - see https://tools.ietf.org/html/rfc2518
```

This would clarify the users to understand why this has been integrated. Otherwise some people could be confused.

Or we just add the information about RFC2518 just to the documentation.

Ma27:
I'll work on this PR on the weekend if I have time to...

Ma27:
as it contains BC breaks, I have to recreate a new PR that can be merged. Have to reopen it on 2.0 :/

Ener-Getick:
Where do you see  bc breaks?

Ma27:
introducing the lock method causes conflicts with the route loader.

When I wrote

``` php
public function lockUserAction() {}
```

Then the rest loader "casted" it to a PATCH method.
With the introduction of a http method called "LOCK", this behavior will be changed.

Ener-Getick:
ah yeah ok a solution would be to support both http methods for ``LOCK``

Ma27:
hmm, will be difficult as "PATCH" worked as a fallback.
I'm working on the tests and re-create a PR to be merged into master (I'll
use FOSRestBundle 2.0 when it comes out, so at least for me it won't be a
problem).

2016-01-15 17:31 GMT+01:00 Ener-Getick <notifications@github.com>:

PR #1311, opened 2016-01-13
Fix fallback format with priorities

egeloen:
Hey!

When trying to perform request on an API without providing the `Accept` header, the fallback format is not used (if the matched rule defines priotities) and instead a `Negotiation\Exception\InvalidArgument` is thrown because the Negotiation library seems not supporting empty header in its 2.x version.

### DIFF
--- a/Negotiation/FormatNegotiator.php
+++ b/Negotiation/FormatNegotiator.php
@@ -83,12 +83,16 @@ class FormatNegotiator extends BaseNegotiator
                 }
             }
 
-            $mimeTypes = $this->normalizePriorities($request,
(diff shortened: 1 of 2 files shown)


Ener-Getick:
Indeed thank you @egeloen :smiley: 
I'm just waiting for the travis build to pass... (there is [an incident](https://www.traviscistatus.com/incidents/pfcclfpffppn))

egeloen:
@Ener-Getick Thanks! If I want to add a new feature, should I target the `1.8` or the `master` branch?

Ener-Getick:
If it's backward compatible, you should target ``1.8`` ;-)

PR #1295, opened 2015-12-30
Add 'boolean' form type

fsevestre:
See #1224 

Sorry for the delay...

--
I hope I didn't forget anything with the deprecation / changes in the form component for the `Symfony` versions covering by this branch.

PR #1312, opened 2016-01-14
Add boolean extension support

egeloen:
Hey!

I was starting wrting this piece of text and realize there is #1295... Anyway, as it is done let's share it and additionally, I'm not really in favor of #1295. Instead, I propose this one which is more strict on the submitted value. IMO, in a API context, if a user sends something like `abcd`, it should not be considered as true or false for a checkbox field but instead it should be considered as invalid.

To enable it, you need to use the `api` type:

``` php
$builder->add('enabled', CheckboxType::class, [
    'type' => 'api',
    'true_values'  => ['1', 'true', 'yes', 'on'],
    'false_values' => ['0', 'false', 'no', 'off', '', null],
]);
```

Enabling it will automatically convert the defined `true_values` to `true` and the `false_values` to `false`. Anything else will trigger a transformation failed exception and mark the form as invalid. The true/false values are optional, I add them in the example in order to more easily explain how does it work. The values used here are the default ones except for the type one.

If you look it closely you will notice the values looks strange. We don't allow the `true`/`false` values or the `1`/`0` values but instead we allow there string representations. This is because the form component automatically converts scalar value to string and false to null (See [here](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Form/Form.php#L525-L529)). So, the values provided must reflect this form behavior.

For the list of supported submitted boolean values, you can take a look to the PHPUnit data provider in the extension tests.

WDYT?

### DIFF
--- a/Form/Extension/BooleanExtension.php
+++ b/Form/Extension/BooleanExtension.php
@@ -1,0 +1,79 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\RestBundle\Form\Extension;
+
+use FOS\RestBundle\Form\Transformer\BooleanTypeToBooleanTransformer;
+use Symfony\Component\Form\AbstractTypeExtension;
+use Symfony\Component\Form\FormBuilderInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+use Symfony\Component\OptionsResolver\OptionsResolverInterface;
+
+/**
+ * @author GeLo <geloen.eric@gmail.com>
+ */
+class BooleanExtension extends AbstractTypeExtension
+{
+    const TYPE_API = 'api';
+    const TYPE_CHECKBOX = 'checkbox';
+
+    /**
+     * {@inheritdoc}
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        if ($options['type'] === self::TYPE_API) {
+            $builder
+                ->resetViewTransformers()
+                ->addViewTransformer(new BooleanTypeToBooleanTransformer());
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function setDefaultOptions(OptionsResolverInterface $resolver)
+    {
+        $this->configureOptions($resolver);
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function configureOptions(OptionsResolver $resolver)
+    {
+        // Symfony >= 2.6
+        if (method_exists($resolver, 'setDefault')) {
+            $resolver
+                ->setDefault('type', self::TYPE_CHECKBOX)
+                ->setAllowedValues('type', array(self::TYPE_API, self::TYPE_CHECKBOX))
+            ;
+        } else {
+            $resolver
+                ->setDefaults(array('type' => self::TYPE_CHECKBOX))
+                ->setAllowedValues(array('type' => array(self::TYPE_API, self::TYPE_CHECKBOX)))
+            ;
+        }
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getExtendedType()
+    {
+        // Symfony >= 2.8
+        if (method_exists('Symfony\Component\Form\AbstractType', 'getBlockPrefix')) {
+            return 'Symfony\Component\Form\Extension\Core\Type\CheckboxType';
+        }
+
+        return 'checkbox';
+    }
+}
--- a/Form/Transformer/BooleanTypeToBooleanTransformer.php
+++ b/Form/Transformer/BooleanTypeToBooleanTransformer.php
@@ -1,0 +1,56 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\RestBundle\Form\Transformer;
+
+use Symfony\Component\Form\DataTransformerInterface;
+use Symfony\Component\Form\Exception\TransformationFailedException;
+
+/**
+ * @author GeLo <geloen.eric@gmail.com>
+ */
+class BooleanTypeToBooleanTransformer implements DataTransformerInterface
+{
+    /**
+     * {@inheritdoc}
+     */
+    public function transform($value)
+    {
+        if ($value === null) {
+            return;
+        }
+
+        if (!is_bool($value)) {
+            throw new TransformationFailedException(sprintf(
+                'The boolean type expects a boolean or null value, got "%s"',
+                is_object($value) ? get_class($value) : gettype($value)
+            ));
+        }
+
+        return $value;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function reverseTransform($value)
+    {
+        if (in_array($value, array('1', 'true', 'on', 'yes'), true)) {
+            return true;
+        }
+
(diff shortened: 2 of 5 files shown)


egeloen:
Oops, not notice the `1.8` branch is compatible with PHP 5.3. Will need to update the array syntax.

Resources/config/form.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <parameters>
+        <parameter key="fos_rest.form.extension.boolean.class">FOS\RestBundle\Form\Extension\BooleanExtension</parameter>
```
Ener-Getick: Please set this class inline

Ener-Getick:
Seems good to me :+1: I'll review it again tomorrow when I'll be more awake.

Resources/config/form.xml
egeloen: Done

egeloen:
Btw, if this PR is accepted the code will need to be adjusted for the master branch (remove BC paths).

blazarecki:
@egeloen Why don't you use `filter_var($value, FILTER_VALIDATE_BOOLEAN)` ? 

See the doc here : http://php.net/manual/en/filter.filters.validate.php

egeloen:
@blazarecki Good point! Will update it after my day work :)

Form/Transformer/BooleanTypeToBooleanTransformer.php
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function reverseTransform($value)
+    {
+        $filterValue = !is_object($value)
+            ? filter_var($value, FILTER_VALIDATE_BOOLEAN, array('flags' => FILTER_NULL_ON_FAILURE))
+            : $value;
```
egeloen: I'm not really fan of this block but it seems `filter_var` is buggy with object.

Quoting the doc:

> Returns TRUE for "1", "true", "on" and "yes". Returns FALSE otherwise.
>
> If FILTER_NULL_ON_FAILURE is set, FALSE is returned only for "0", "false", "off", "no", and "", and NULL is returned for all non-boolean values.

Unfortunatelly, passing an `\stdClass` (as it is done in the tests) returns `false` instead of `null`. So I workaround it this way for now...

egeloen:
I have updated the PR to rely on `filter_var` as suggested by @blazarecki

egeloen:
Hum, according to travis, I'm not sure we can safely rely on `filter_var`. At least on PHP 5.3, the behavior seems not the same.

WDYT?

egeloen:
Any feedback on what you would prefer?

Form/Transformer/BooleanTypeToBooleanTransformer.php
Ener-Getick: The best and clearer solution might be to emulate ``filter_var`` behavior:
```php
in_array(array('1', 'true', 'on', 'yes'), $value) ?: (in_array(array('0', 'false', 'off', 'no', ''), $value) ? false : null);
```

xabbuh: I agree that we can be explicit here. But please let us not use the ternary operator here then as it becomes quite unreadable now.

Ener-Getick: Yeah that was a quick example of inlining I'd prefer a ``if/else`` structure.

egeloen: Thanks for the feedback! So, we all agree to get rid of `filter_var`. I will update it tonight.

egeloen:
After refactoring it using `in_array`, I'm still frustrating about the `null` and `false` values equality. Basically, the form component does not make difference between both as you can see [here](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Form/Form.php#L525-L529). False value are converted to null and null is kept as it is making no possibility to distinguish them latter whereas in an API context bot are totally different values. Not sure how to workaround it right now...

blazarecki:
I think this feature is already in the bundle.

This decoder allow to use boolean https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Decoder/JsonToFormDecoder.php

In the config you can add:

```yaml
fos_rest:
    body_listener:
        decoders:
            json: fos_rest.decoder.jsontoform
```

lsmith77:
@Ener-Getick or @xabbuh do you have time to look at this .. ?

Ener-Getick:
@blazarecki that's not the same thing, we're talking about converting ``'1'``, ``'true'``, ... to ``true`` in symfony forms.

Ener-Getick:
Otherwise seems good to me but I don't very well the Form component.

egeloen:
IMO, since we have not found a way to distinguish `null` & `false` values, this PR still need some work... I will regive it a try tomorrow.

lsmith77:
please also rebase on master

egeloen:
@lsmith77 on master? The target branch is 1.8.

egeloen:
Rebased on 1.8

lsmith77:
@xabbuh can you have a look here as I am not a form expert either ..

egeloen:
For now, in my project, I "pre-transform" the data to something I can deal with before calling the `submit` method: https://github.com/php-lug/lug/blob/master/src/Bundle/ResourceBundle/Controller/Controller.php#L246-L250

An other solution is to override the form request handler but then it will only work with it not directly with the submit method...

@webmozart Any suggestions?

Form/Transformer/BooleanTypeToBooleanTransformer.php
```diff
@@ -0,0 +1,56 @@
...
+
+        return $value;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function reverseTransform($value)
+    {
+        if (in_array($value, array('1', 'true', 'on', 'yes'), true)) {
```
webmozart: This should rather be done in a listener for the SUBMIT event.

Form/Extension/BooleanExtension.php
```diff
@@ -0,0 +1,79 @@
...
+{
+    const TYPE_API = 'api';
+    const TYPE_CHECKBOX = 'checkbox';
+
+    /**
+     * {@inheritdoc}
+     */
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        if ($options['type'] === self::TYPE_API) {
```
webmozart: IMO the option name "type" is too generic. Can you find a better option name that describes better what the option does? E.g. `strict_string_handling`?

webmozart:
The problem about `null`/`false` is that the Form component was never designed to be used in API requests. That's what usually a serializer is for. The Form component is designed against the CGI protocol, where each value is a string or not set.

I understand however that the Form component offers functionality that serializers are missing. The problem is that we can't change the Form component to pass `false` as `false` (instead of `null`) without breaking BC.

Form/Extension/BooleanExtension.php
xabbuh: I agree. The option could then also be changed to only except boolean values.

What do you also think about adding a global option to change the default value for this option? For example, in an API only application you might want to always set the option to `true` without changing all your form types while you rather do not want to do the same in an application that offers an API and a common HTML interface.

webmozart: Sure, why not. :)

lsmith77:
lets move this back to 2.1 then ..

Form/Transformer/BooleanTypeToBooleanTransformer.php
afurculita: IMO we can use the validate filter `FILTER_VALIDATE_BOOLEAN` with the option `FILTER_NULL_ON_FAILURE`

egeloen: @afurculita I have already give a try and the result is not consistent accross different PHP versions...

Tobion:
Closing as outdated.

PR #1314, opened 2016-01-15
Compatibility for RFC-2518

Ma27:
this pr aims to provide support for the http methods introduced in the RFC-2518 (e.g. LOCK/UNLOCK or MOVE)

**ToDo**

- [x] annotation classes
- [x] support for the rest routing loader
- [x] docs
- [x] tests with annotations

### DIFF
--- a/Routing/Loader/Reader/RestActionReader.php
+++ b/Routing/Loader/Reader/RestActionReader.php
@@ -14,10 +14,9 @@ namespace FOS\RestBundle\Routing\Loader\Reader;
 use Doctrine\Common\Annotations\Reader;
 use FOS\RestBundle\Controller\Annotations\Route as RouteAnnotation;
 use FOS\RestBundle\Inflector\InflectorInterface;
-use FOS\RestBundle\Request\ParamReader;
+use FOS\RestBundle\Request\ParamReaderInterface;
 use FOS\RestBundle\Routing\RestRouteCollection;
 use Symfony\Component\Routing\Route;
-use FOS\RestBundle\Request\ParamReaderInterface;
 
 /**
  * REST controller actions reader.
@@ -28,17 +27,81 @@ class RestActionReader
 {
     const COLLECTION_ROUTE_PREFIX = 'c';
 
+    /**
+     * @var Reader
+     */
     private $annotationReader;
+
+    /**
+     * @var ParamReaderInterface
+     */
     private $paramReader;
+
+    /**
+     * @var InflectorInterface
+     */
     private $inflector;
+
+    /**
+     * @var array
+     */
     private $formats;
+
+    /**
+     * @var bool
+     */
     private $includeFormat;
+
+    /**
+     * @var string|null
+     */
     private $routePrefix;
+
+    /**
+     * @var string|null
+     */
     private $namePrefix;
+
+    /**
+     * @var array|string|null
+     */
     private $versions;
+
+    /**
+     * @var bool|null
+     */
     private $pluralize;
+
+    /**
+     * @var array
+     */
     private $parents = [];
-    private $availableHTTPMethods = ['get', 'post', 'put', 'patch', 'delete', 'link', 'unlink', 'head', 'options'];
+
+    /**
+     * @var array
+     */
+    private $availableHTTPMethods = [
+        'get',
+        'post',
+        'put',
+        'patch',
+        'delete',
+        'link',
+        'unlink',
+        'head',
+        'options',
+        'mkcol',
+        'propfind',
+        'proppatch',
+        'move',
+        'copy',
+        'lock',
+        'unlock',
+    ];
+
+    /**
+     * @var array
+     */
     private $availableConventionalActions = ['new', 'edit', 'remove'];
 
     /**
@@ -536,7 +599,7 @@ class RestActionReader
             return 'get';
         }
 
-        //custom object
+        // custom object
         return 'patch';
     }
 
@@ -551,7 +614,7 @@ class RestActionReader
     {
         $annotations = [];
 
-        foreach (['Route', 'Get', 'Post', 'Put', 'Patch', 'Delete', 'Link', 'Unlink', 'Head', 'Options'] as $annotationName) {
+        foreach ($this->convertAvailableHttpMethodsToAnnotationNames() as $annotationName) {
             if ($annotations_new = $this->readMethodAnnotations($reflectionMethod, $annotationName)) {
                 $annotations = array_merge($annotations, $annotations_new);
             }
@@ -648,5 +711,20 @@ class RestActionReader
         } else {
             $collection->add($fullRouteName, $route);
         }
+    }
+
+    /**
+     * Converts all given http methods to their annotation names.
+     *
+     * @return string[]
+     */
+    private function convertAvailableHttpMethodsToAnnotationNames()
+    {
+        return array_map(
+            function ($httpMethod) {
+                return ucfirst($httpMethod);
+            },
+            $this->availableHTTPMethods
+        );
     }
 }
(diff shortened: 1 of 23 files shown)


composer.json
```diff
@@ -27,7 +27,8 @@
         "symfony/routing":                "^2.7|^3.0",
         "doctrine/inflector":             "^1.0",
         "willdurand/negotiation":         "^2.0",
-        "willdurand/jsonp-callback-validator": "^1.0"
+        "willdurand/jsonp-callback-validator": "^1.0",
+        "phpunit/phpunit": "^5.1"
```
Ma27: should be removed. (I prefer using the composer dependency and just forgot to remove)

Ma27:
ping @Ener-Getick @lsmith77 I've recreated the PR and now it is ready to review

Ma27:
ping @Ener-Getick @lsmith77

Ener-Getick:
Sorry i'll review your PR during the week I don't have so much time for now.

Ma27:
ok thank you @Ener-Getick

Routing/Loader/Reader/RestActionReader.php
```diff
@@ -550,9 +613,9 @@ private function readRouteAnnotation(\ReflectionMethod $reflectionMethod)
     {
         $annotations = [];
 
-        foreach (['Route', 'Get', 'Post', 'Put', 'Patch', 'Delete', 'Link', 'Unlink', 'Head', 'Options'] as $annotationName) {
-            if ($annotations_new = $this->readMethodAnnotations($reflectionMethod, $annotationName)) {
-                $annotations = array_merge($annotations, $annotations_new);
+        foreach ($this->convertAvailableHttpMethodsToAnnotationNames() as $annotationName) {
```
Ener-Getick: I do not understand why not just using:
```php
$annotations = $this->readMethodAnnotations($method, 'Route');
```
Thanks to inheritance, this should match all annotations.

Routing/RestRouteCollection.php
```diff
@@ -76,10 +76,17 @@ public function setDefaultFormat($format)
      */
     public function all()
     {
+        $regex = <<<REGEX
```
Ener-Getick: Not sure this is better that way.

UPGRADING.md
```diff
@@ -18,6 +18,7 @@ This document will be updated to list important BC breaks and behavioral changes
  * removed all ``.class`` parameters, instead overwriting services via explicit Bundle configuration is preferred
  * renamed ``AbstractScalarParam::$array`` to ``AbstractScalarParam::$map``
  * added `ControllerTrait` for developers that prefer to use DI for their controllers instead of extending ``FOSRestController``
+ * when having an action called ``lockUserAction``, then it will have use the http method ``LOCK`` (RFC-2518) instead of ``PATCH``
```
Ener-Getick: I'm still not sure if we should not allow both ways...

Routing/Loader/Reader/RestActionReader.php
Ma27: ok, let's check it out

Routing/RestRouteCollection.php
Ma27: I usually work on a 15" Notebook and I really hate it to scroll horizontal in PhpStorm, so I decided to write it that way

Ener-Getick:
Otherwise good for me :+1:

Routing/RestRouteCollection.php
Ener-Getick: There is no ``soft wrap`` ? :-P

Ma27: call me crazy, but this looks IMHO odd, too :D 
But hey, because of that the following rule exists: http://www.php-fig.org/psr/psr-2/ (120 char limit)

Ener-Getick: I'm not a huge fan of this annotation :x
Maybe concatenating multiple strings would be better?

Anyway that's not a big point, @xabbuh what do you use in symfony for the long regexps?

xabbuh: We have few places with such long regular expressions, but you may want to take a look at how this is done in the `UrlValidator`: https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Validator/Constraints/UrlValidator.php#L23-L37

xabbuh:
Can you please shortly explain what the use case is for these new methods (and maybe link to a short description or some other related resources)?

Ener-Getick:
@xabbuh you can see his previous pull request https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1309

UPGRADING.md
xabbuh: I don't think it's a good idea to automatically support both HTTP methods then.

xabbuh: But, doesn't this also apply to the other new methods and shouldn't they be listed here too?

Ener-Getick: You're right I was more thinking about a BC approach but this would look strange...

xabbuh: Might be an idea to think about this so that we could include it in 1.8. However, you will then still allow additional HTTP methods which one could consider a BC break too.

Routing/RestRouteCollection.php
Ma27: ok, then I'll change it. I observed that syntax in some parts of symfony, so I assume, it would be ok to use it (https://github.com/symfony/debug/blob/master/ExceptionHandler.php#L272)

UPGRADING.md
Ma27: oh yes, forgot to update the UPGRADING.md...

Routing/RestRouteCollection.php
Ener-Getick: ping

UPGRADING.md
Ener-Getick: And there is a typo here ``have to use``

Ma27:
@Ener-Getick is it okay when changing all of it tomorrow?

Ener-Getick:
For me the rest is ok :+1:

lsmith77:
sounds great!

xabbuh:
The rest looks great to me too. The only thing we could think about is to backport parts of this to 1.8 (the non BC breaking parts). But that's something we can decide after merging this PR.

Ma27:
well I think that adding more http methods is actually a BC break:

imagine the following case:

``` php
public function copyUserAction() {
    // do foo...
}
```

this method accepts a PATCH request when using 1.7, but when upgrading to 1.8, everything would break as a new method (COPY) is necessary now.

Ener-Getick:
@Ma27 We can introduce this new methods in ``1.8`` except ``COPY`` and ``LOCK`` which break BC.

xabbuh:
We could at least backport the annotations. Setting the HTTP method based on the method name for the new support methods would indeed be a BC break though.

Ma27:
@Ener-Getick all methods would break BC when introducing the changes of the rest route loader.
@xabbuh :+1: I'll open a PR after that has been merged

Ma27:
the first commit of the pr can be cherry-picked into 1.8. I'll open another PR for that tonight

Ma27:
@xabbuh rebased master

Ma27:
@lsmith77 @Ener-Getick it seems as scrutinizer fails because of anything

lsmith77:
heh .. now scrutinizer is happy but styleci is unhappy ..

Ma27:
ok, tbf tomorrow

Ener-Getick:
ping

Ma27:
sry for the delay, github and its services should be happy now.

Ma27:
ping @lsmith77 @Ener-Getick

Ener-Getick:
Seems good to me :+1:
Status: reviewed

Ma27:
ok :)

lsmith77:
awesome! .. @Ma27 thanks for working through all the review feedback! :)

Ma27:
you're welcome

Von meinem iPhone gesendet

PR #1315, opened 2016-01-16
Fixed typo in ExceptionController.php

palra:
In`ExceptionController::isSubclassOf`, when an invalid class was found in the array of exceptions that are allowed to display their message, it returned an exception message where `FOSUserBundle` was written, instead of `FOSRestBundle`. In the same message, the configuration parameter `fos_rest.exception.messages` was mispelled.

### DIFF
--- a/Controller/ExceptionController.php
+++ b/Controller/ExceptionController.php
@@ -190,7 +190,7 @@ class ExceptionController
                 }
             }
         } catch (\ReflectionException $re) {
-            return 'FOSUserBundle: Invalid class in fos_res.exception.messages: '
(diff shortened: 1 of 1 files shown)


Ener-Getick:
I think that we should refactor this function to use ``is_sublass_of`` which is normally faster than reflections.

Ener-Getick:
Or http://php.net/manual/en/function.is-a.php with allow string as true.

xabbuh:
closing here because of the changes @Ener-Getick made in #1316

@palra Thanks for starting this. Please let me know if you disagree and you think that the issue still isn't fixed fully.

PR #1323, opened 2016-01-22
merge branch 1.8 into master

xabbuh:


EventListener/ViewResponseListener.php
```diff
@@ -114,7 +114,7 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
             $template = null !== $configuration && $configuration->getTemplate()
                 ? $configuration->getTemplate()
                 : $request->attributes->get('_template');
-            if ($template) {
+            if ($template && !$view->getTemplate()) {
```
xabbuh: This was added in #1250. I first overlooked this but the tests passed nonetheless. So it looks like we should add a test on the `1.7` branch to avoid future regressions.

PR #1334, opened 2016-01-27
Serializer adapters no longer final

rokkie:
The serializers are marked lazy, but defined as final. This causes an
exception being thrown when de container is initialized since it tries
to create a proxy, but fails because the classes are final.
Fixes#1325

### DIFF
--- a/Serializer/JMSSerializerAdapter.php
+++ b/Serializer/JMSSerializerAdapter.php
@@ -22,7 +22,7 @@ use JMS\Serializer\SerializerInterface;
  *
  * @author Christian Flothmann <christian.flothmann@xabbuh.de>
  */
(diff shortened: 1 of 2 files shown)


Ener-Getick:
Thank you @rokkie

hmlb:
Is 1.8 going to be merged in master soon ? This fix would be useful for 2.0@dev

Ener-Getick:
I can't do it this week, so except if @xabbuh or @lsmith77 do it, not before next week.

hmlb:
Okay thanks @Ener-Getick  !

xabbuh:
I'll do this in #1335. Thanks for this fix @hmlb.

PR #1338, opened 2016-02-02
Backport/pr 1314

Ma27:
@xabbuh wanted me to backport the feature implemented in #1314 into the 1.8 branch to provide at least parts of the feature in older versions, so the first two commits of that feature containing no BC breaks are cherry-picked into this branch.

for more information please see the discussion in the actual PR for 2.0

### DIFF
--- a/Controller/Annotations/Copy.php
+++ b/Controller/Annotations/Copy.php
@@ -1,0 +1,28 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\RestBundle\Controller\Annotations;
+
+/**
+ * COPY Route annotation class.
+ *
+ * @Annotation
+ * @Target("METHOD")
+ *
+ * @author Maximilian Bosch <maximilian.bosch.27@gmail.com>
+ */
+class Copy extends Route
+{
+    public function getMethod()
+    {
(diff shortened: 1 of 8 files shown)


Ener-Getick:
I think that's confusing if they aren't supported by the router.

Ma27:
I assumed that I should import all the changes that don't contain bc (annotations only). Thewhole router changes are a BCbreak.

Von meinem iPhone gesendet

Ener-Getick:
No only for lock

Ma27:
actually not: the fallback chosen for controller actions is "PATCH".

now imagine you have any controller methods like this:

``` php
public function copyUserAction() {}
public function mkcolUserAction() {}
public function *UserAction() {}
```

Then everyone expects in the 1.x release to have "PATCH" as fallback. With our change we introduce new http methods and then the default is not patch anymore, but MKCOL/COPY and that's definetly a BC break.

lsmith77:
but you opened the PR again on master?

Ma27:
embarrasing... could you please do the whole stuff with the backport if you still wish to have it (I assume that the missing router feature is a :-1: for this feature)...

xabbuh:
@Ener-Getick @lsmith77 I think we can backport only the annotation related changes to 1.8. What do you think?

Ener-Getick:
What do you mean ? I didn't understand your sentence :x

xabbuh:
@Ener-Getick Sorry if that wasn't clear. I meant to backport the support for the new annotations to 1.8 and trigger deprecations if the method name prefixes whose behaviour will change with 2.0 are used. I prepare a pull request.

Ener-Getick:
Ah ok fine for me :+1:

Issue #1237, opened 2015-11-30
Not compatible with _error route in symfony 2.7+


lyrixx:
I have tested it with symfony 2.7 and symfony 2.8 and both generate the same error.
How to reproduce:

* `symfony new sf-error-fosrest 2.8` (2.7 is affected too)
* install jms serializer
* install fos rest
* enable body converter:
```
# config.yml
fos_rest:
    body_converter:
        enabled: true
```
* `app/console server:run`

Then go to http://127.0.0.1:8000/_error/404

![screenshot6](https://cloud.githubusercontent.com/assets/408368/11471885/bc328d92-9768-11e5-9943-d545e34a7507.png)

Ener-Getick:
This error is caused by the twig exception controller which requires the logger.
Only the RequestBodyParamConverter can support this class so it is executed but it wasn't expected to be...

I don't really see a solution to this problem except creating a list of the classes to ignore...

lyrixx:
I don't know fos rest at all, but I don't understand why fos rest try to do thing with constructor argument. It should deal only controller argument?

Ener-Getick:
FOSRestBundle tries to deal with [this class method](https://github.com/symfony/twig-bundle/blob/master/Controller/ExceptionController.php#L56) which requires the logger. This is not a constructor argument ;-)

lyrixx:
Oups, indeed. Sorry.

Ener-Getick:
We can also force the user to explicitly requests the ``RequestBodyParamConverter`` by checking that ``ParamConverter::getConverter() === 'fos_rest.request_body'`` but this is a BC break so only in 2.0...

lyrixx:
How can I work-arround it?

Ener-Getick:
What do you think about this @xabbuh @lsmith77 ?

ogizanagi:
@Ener-Getick : What about adding a check in the `supports` method for a list of approved paths from a new configuration parameter:

```yaml
fos_rest:
    body_converter:
        enabled: true
        paths:  ['^/'] # defaults 
```

By default, when enabled, the converter will be supported for every requests. Otherwise, it can be configured to include required paths (or can work with exclusions instead, or both).

It'll do the trick for FOSRestBundle  < 2.0, and could be removed in 2.0 in favor of explicitly requesting the converter on the controller.

Ener-Getick:
[...]

PR #1317, opened 2016-01-16
[FormatNegotiator] Re-add the ability to transform automatically most common mime types when using symfony 3.1

Ener-Getick:
Re-add the ability to transform automatically most common mime types.

See https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1149#discussion-diff-41261255

PR #1346, opened 2016-02-13
Define a priority of -50 on the fos_rest.request_body converter

Ener-Getick:
Fix https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1237

### DIFF
--- a/Request/RequestBodyParamConverter.php
+++ b/Request/RequestBodyParamConverter.php
@@ -96,11 +96,11 @@ class RequestBodyParamConverter implements ParamConverterInterface
                 $context
             );
         } catch (UnsupportedFormatException $e) {
-            throw new UnsupportedMediaTypeHttpException($e->getMessage(), $e);
+            return $this->throwException(new UnsupportedMediaTypeHttpException($e->getMessage(), $e), $configuration);
         } catch (JMSSerializerException $e) {
-            throw new BadRequestHttpException($e->getMessage(), $e);
+            return $this->throwException(new BadRequestHttpException($e->getMessage(), $e), $configuration);
         } catch (SymfonySerializerException $e) {
-            throw new BadRequestHttpException($e->getMessage(), $e);
+            return $this->throwException(new BadRequestHttpException($e->getMessage(), $e), $configuration);
         }
 
         $request->attributes->set($configuration->getName(), $object);
@@ -146,6 +146,18 @@ class RequestBodyParamConverter implements ParamConverterInterface
                 $context->setAttribute($key, $value);
             }
         }
+    }
+
+    /**
+     * Throws an exception or return false if a ParamConverter is optional.
+     */
+    private function throwException(\Exception $exception, ParamConverter $configuration)
+    {
+        if ($configuration->isOptional()) {
+            return false;
+        }
+
+        throw $exception;
     }
 
     /**
--- a/Tests/Functional/RequestBodyParamConverterTest.php
+++ b/Tests/Functional/RequestBodyParamConverterTest.php
@@ -28,4 +28,17 @@ class RequestBodyParamConverterTest extends WebTestCase
         $this->assertSame(200, $client->getResponse()->getStatusCode());
         $this->assertSame('Post 1', $client->getResponse()->getContent());
     }
+
+    /**
+     * @see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1237
+     */
+    public function testTwigErrorPage()
+    {
+        $client = $this->createClient(['test_case' => 'RequestBodyParamConverter']);
+        $client->request('GET', '/_error/404.txt');
+
+        // Status code 200 as this page describes an error but is not the result of an error.
+        $this->assertEquals(200, $client->getResponse()->getStatusCode());
+        $this->assertContains('The server returned a "404 Not Found".', $client->getResponse()->getContent());
+    }
 }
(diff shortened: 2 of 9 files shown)


Ener-Getick:
After all I'm not sure this fixes https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1237 but anyway I think it's better to execute our param converter after the others as it is less restrictive (its ``supports`` function doesn't check anything).

I think we should look at https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1237#issuecomment-160698182 instead for fixing #1237 but this won't be a BC break so we can do it later.

lsmith77:
I would say .. lets merge and see what reports we get during the RC phase

lsmith77:
but it would be good to have a test case for this ..

Ener-Getick:
@lsmith77 I fixed #1237 by not throwing exceptions when a parameter is optional.

Request/RequestBodyParamConverter.php
```diff
@@ -95,18 +95,26 @@ public function apply(Request $request, ParamConverter $configuration)
...
+                );
+            } catch (UnsupportedFormatException $e) {
+                throw new UnsupportedMediaTypeHttpException($e->getMessage(), $e);
+            } catch (JMSSerializerException $e) {
+                throw new BadRequestHttpException($e->getMessage(), $e);
+            } catch (SymfonySerializerException $e) {
+                throw new BadRequestHttpException($e->getMessage(), $e);
+            }
+        } catch (\Exception $e) {
+            if ($configuration->isOptional()) {
```
xabbuh: What about having this check in each `catch` block or moving it into a private method? I find the nested `try` blocks harder to read.

Tests/Functional/RequestBodyParamConverterTest.php
```diff
@@ -28,4 +28,13 @@ public function testRequestBodyIsDeserialized()
...
         $this->assertSame('Post 1', $client->getResponse()->getContent());
     }
+
+    /**
+     * @see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1237
+     */
+    public function testTwigErrorPage()
+    {
+        $client = $this->createClient(['test_case' => 'RequestBodyParamConverter']);
+        $client->request('GET', '/_error/404');
```
xabbuh: Having a test without an assertion looks wrong to me.

Ener-Getick: It is testing that no exceptions are thrown.

Request/RequestBodyParamConverter.php
Ener-Getick: I don't like duplications really much but I agree that it may be the best solution here :-\

Tests/Functional/RequestBodyParamConverterTest.php
xabbuh: I think so. Checking for the proper status code (404 instead of 500) should be enough though.

Ener-Getick:
@xabbuh I added a commit with your suggestions.

xabbuh:
@Ener-Getick Just to fully understand the changes: What is the underlying in #1237 and how is solved by this PR?

Ener-Getick:
@xabbuh The issue in #1237 is that the ``RequestBodyParamConverter`` tries to create a ``Logger`` instance for [this controller](https://github.com/symfony/twig-bundle/blob/master/Controller/ExceptionController.php#L56). 
This behavior is normal because the ``logger`` attribute is not present in the Request attributes as it is optional.

But the issue is that the ``RequestBodyParamConverter`` can't create a ``Logger`` instance from the ``Request`` content so it throws an error.
This PR fixes this by not throwing errors when a parameter is optional.

Ener-Getick:
@lsmith77 @xabbuh is this new behavior ok for you ?

lsmith77:
yup .. good for me ... @xabbuh ?

Tests/Functional/RequestBodyParamConverterTest.php
```diff
@@ -28,4 +28,16 @@ public function testRequestBodyIsDeserialized()
...
+
+    /**
+     * @see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1237
+     */
+    public function testTwigErrorPage()
+    {
+        $client = $this->createClient(['test_case' => 'RequestBodyParamConverter']);
+        $client->request('GET', '/_error/404.txt');
+
+        $this->assertEquals(200, $client->getResponse()->getStatusCode());
```
xabbuh: Why do we have a 200 status code here?

Ener-Getick: Because this page describes an error it is not a page always shown when there is a 404 status code.

xabbuh: Ah, you are right. Thanks. Maybe we should add a comment to avoid confusion for anybody else stumbling over this.

Ener-Getick: Well it seems logic to me. What kind of message do you think about?

xabbuh:
I will still need to play with this locally to completely understand the fix, but as this seems to let the original issue go away, we can merge this imo.

xabbuh:
By the way, does this affect 1.8 too?

Ener-Getick:
No, this is a full bc break not applicable to 1.8. BTW i'll have to add it to the upgrade file.

Ener-Getick:
@xabbuh @lsmith77 I updated ``UPGRADING.md`` and I added a comment in ``RequestBodyParamConverterTest`` to explain why the page returns a 200 status.
This PR is ready for me and https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1317 too. Do you see anything else to change before merging them and releasing 1.8/2.0 ?

lyrixx:
Hello. Do you have an ETA for the release of this Patch?

Thanks.

Ener-Getick:
This is included in 2.0.0-BETA1 but we don't know when we will release it as stable as it depends if we have feedbacks or not.

PR #1350, opened 2016-02-15
fix lower version bounds

xabbuh:


Ener-Getick:
@xabbuh it doesn't seem to work see https://travis-ci.org/FriendsOfSymfony/FOSRestBundle/jobs/109469333

xabbuh:
I wasn't sure if we need to update the requirement or the conflict rule. But test seem to confirm that the conflict rule indeed needs to be changed.

PR #1352, opened 2016-02-15
Fixed arguments passed to custom handlers

jonjonw:
When trying to use custom view handlers, an error would be thrown saying a view is expected for the first parameter to the createResponse method. It looks like $this was being added as a parameter in the call which was causing the error. This changed fixed it for me.


Ener-Getick:
It is not an error, see [this comment](https://github.com/jonjonw/FOSRestBundle/blob/patch-1/View/ViewHandler.php#L188).

PR #1358, opened 2016-02-26
Proposal to improve the ExceptionController

Ener-Getick:
This PR removes the ExceptionWrapper in favor of normalizers.
This would allow people to easily customize the exception controller output by creating a normalizer.
An example:
```php
use AppBundle\Exception\MyCustomException;
use Symfony\Component\Serializer\Normalizer\NormalizerInterface;

class MyCustomExceptionNormalizer implements NormalizerInterface
{
    /**
     * {@inheritdoc}
     */
    public function normalize($object, $format = null, array $context = [])
    {
        return [
            'code' => 'foo',
            'message' => $object->getMessage(),
            'user_output' => 'Please select another option',
        ];
    }

    /**
     * {@inheritdoc}
     */
    public function supportsNormalization($data, $format = null)
    {
        return $data instanceof MyCustomException;
    }
}
```

What do you think ?


Controller/ExceptionController.php
```diff
@@ -71,68 +53,17 @@ protected function getViewHandler()
...
         $code = $this->getStatusCode($exception);
-        $parameters = $this->getParameters($currentContent, $code, $exception, $logger, $format);
-        $showException = $request->attributes->get('showException', $this->showException);
-
-        try {
-            $view = $this->createView($format, $exception, $code, $parameters, $request, $showException);
+        $data = $this->getData($currentContent, $code, $exception, $logger);
 
-            $response = $this->viewHandler->handle($view);
-        } catch (\Exception $e) {
```
Ener-Getick: IMO we should encourage people to look their logs instead of this ugly try/catch which doesn't log anything BTW.

Controller/TwigExceptionController.php
```diff
@@ -28,31 +27,13 @@ class TwigExceptionController extends TemplatingExceptionController
     protected function createView($format, FlattenException $exception, $code, $parameters, Request $request, $showException)
     {
         $view = parent::createView($format, $exception, $code, $parameters, $request, $showException);
-
-        if ($this->getViewHandler()->isFormatTemplating($format)) {
-            $view->setTemplate($this->findTemplate($request, $format, $code, $showException));
-            $view->setData($parameters);
-        }
+        $view->setTemplate($this->findTemplate($request, $format, $code, $showException));
```
Ener-Getick: What do you think about searching the available templates in a compiler pass instead to avoid calls to the template engine which probably cost a lot of resources ?

xabbuh: We would then need to access the templating service during compilation. I don't think that this is a good idea.

A better approach would be do this when warming up the cache for a reasonable matrix of formats and status codes and then rely on the cached data if present.

Controller/ExceptionController.php
```diff
@@ -71,68 +53,17 @@ protected function getViewHandler()
...
-            return $this->createPlainResponse($message, Response::HTTP_NOT_ACCEPTABLE, $exception->getHeaders());
-        }
-
-        $currentContent = $this->getAndCleanOutputBuffering(
-            $request->headers->get('X-Php-Ob-Level', -1)
-        );
+        $currentContent = $this->getAndCleanOutputBuffering($request->headers->get('X-Php-Ob-Level', -1));
         $code = $this->getStatusCode($exception);
-        $parameters = $this->getParameters($currentContent, $code, $exception, $logger, $format);
-        $showException = $request->attributes->get('showException', $this->showException);
```
xabbuh: Is it intended to not check the request attribute anymore?

Controller/ExceptionController.php
```diff
@@ -71,68 +53,17 @@ protected function getViewHandler()
...
+        $data = $this->getParameters($currentContent, $code, $exception, $logger);
 
-            $response = $this->viewHandler->handle($view);
-        } catch (\Exception $e) {
-            $message = 'An Exception was thrown while handling: ';
-            $message .= $this->getExceptionMessage($exception);
-            $response = $this->createPlainResponse($message, Response::HTTP_INTERNAL_SERVER_ERROR, $exception->getHeaders());
-        }
+        $view = $this->createView($exception, $code, $data, $request, $this->showException);
+        $response = $this->viewHandler->handle($view);
```
xabbuh: What happens now if an exception is thrown by the view handler?

Serializer/Normalizer/ExceptionNormalizer.php
```diff
@@ -23,13 +22,13 @@
...
-class ExceptionWrapperNormalizer implements NormalizerInterface
+class ExceptionNormalizer implements NormalizerInterface
 {
-    private $translator;
+    private $exceptionMessages;
 
-    public function __construct(TranslatorInterface $translator)
+    public function __construct($exceptionMessages)
     {
-        $this->translator = $translator;
```
xabbuh: The translator is still used in `getErrorMessage()`.

xabbuh:
I left some comments, but in general I like the approach. Maybe this could be something for 2.1? In we could mark some more classes as internal (if needed) so that we could easily drop them in 2.1 without breaking BC.

Ener-Getick:
@xabbuh the removing of the ExceptionWrapper will impact many people but of course we can include that in 2.1 with deprecations.

Serializer/Normalizer/ExceptionNormalizer.php
Ener-Getick: Yeah it's just a first approach ^_^ I plan to move the form normalization to another normalizer.

Controller/ExceptionController.php
Ener-Getick: Yeah the request attributes are reinitialized in the ExceptionListener so this is useless IMO.

Controller/ExceptionController.php
Ener-Getick: It is catched by the exception listener and logged.

Serializer/Normalizer/ExceptionHandler.php
```diff
@@ -93,11 +93,11 @@ public function serializeToXml(
...
-     * @param ExceptionWrapper $exceptionWrapper
+     * @param \Exception $exception
      *
      * @return array
      */
-    protected function convertToArray(ExceptionWrapper $exceptionWrapper)
+    protected function convertToArray(\Exception $exception, Context $context)
     {
         return [
             'code' => $exceptionWrapper->getCode(),
```
Abam: The `$exceptionWrapper` is not known here.

Ener-Getick: Yeah you should not look the normalizer inner. 
 I just wanted to have feedbacks on the globality of the proposal so I didn't update everything

Abam:
You should also consider users of jms serializer. You normalier is only for symfony serializer.
When Symfony call the exception controller, it changes any custom exception to a `Symfony\Component\Debug\Exception\FlattenException`. 
You custom exception is never passed to you serializer.

Ener-Getick:
@Abam This PR is marked as "Work in progress" so it's just an example of implementation and I didn't fixed the implementation details yet.

Abam:
@Ener-Getick I've got it. Just highlighting that using a nomalizer for handling exceptions doesn't seems to be the right way for tackling this problem.
FOSRestBundle uses the base symfony exception listener to handle exceptions and that listener convert any exception to a `Symfony\Component\Debug\Exception\FlattenException`. So when are in FOSRestBundle ExceptionController, you have only access to the flatten exception.

Ener-Getick:
@Abam I overwritted the exception listener to pass the raw exception to the controller ;-)

Ener-Getick:
This PR is almost finished. 
I just have to finish to fix the tests; if anyone has an idea why they fail, it would be helpful ☺

Controller/TwigExceptionController.php
Ener-Getick: :+1: to be done for 2.1 !

DependencyInjection/Compiler/FormatListenerRulesPass.php
```diff
@@ -41,7 +41,7 @@ public function process(ContainerBuilder $container)
                 'path' => "^/$path/",
                 'priorities' => ['html', 'json'],
                 'fallback_format' => 'html',
-                'exception_fallback_format' => 'html',
+                'attributes' => array(),
```
Ener-Getick: I think that's better to avoid having multiple format negotiator and instead allow to detect when the ExceptionController is used.
Our code is clearer and easier to maintain that way.

Ener-Getick:
Only one error remaining !

.travis.yml
```diff
@@ -12,11 +12,6 @@ cache:
   directories:
     - $HOME/.composer/cache/files
 
-branches:
-  only:
-    - master
-    - /^\d+\.\d+$/
```
Ener-Getick: It seems that travis doesn't run twice the build now.

Serializer/JMSHandlerRegistry.php
```diff
@@ -0,0 +1,59 @@
...
+use JMS\Serializer\SerializationContext as JMSSerializationContext;
+use JMS\Serializer\SerializerInterface;
+use JMS\Serializer\Handler\SubscribingHandlerInterface;
+use JMS\Serializer\Handler\HandlerRegistryInterface;
+
+/**
+ * @author Ener-Getick <egetick@gmail.com>
+ * @internal do not depend on this class directly.
+ */
+class JMSHandlerRegistry implements HandlerRegistryInterface
```
Ener-Getick: Should be proposed directly to the JMSSerializer but done in FOSRestBundle for the moment to be able to use it without changing our version constraints.

Tests/View/ViewHandlerTest.php
```diff
@@ -513,13 +507,11 @@ public function testCreateResponseWithFormErrorsAndSerializationGroups($format)
...
-                'errors' => $form,
-            ]
-        );
+        $exceptionWrapper = [
+            'status_code' => 400,
+            'message' => 'Validation Failed',
+            'errors' => $form,
+        ];
 
         $view = new View($exceptionWrapper);
```
Ener-Getick: What does this actually test?

lsmith77:
this PR now conflicts too .. but yeah seems like a good idea!

Ener-Getick:
@lsmith77 should be fixed now :-)

BTW if we could merge this in 2.0 it would be much easier for me (there are many bc breaks in it).

lsmith77:
@Ener-Getick well we kind of are stuck in the never releasing phase right now .. that being said .. you are doing the bulk of the work in the past months .. so I would say its your call :)

Ener-Getick:
Well let's say if I finish it today we merge it in 2.0 ? Otherwise let's just release 1.8/2.0, I think we have waited enough ^_^

Ener-Getick:
And I found why the tests fail. Symfony have a weird behaviour when serializing in XML an object implementing \Traversable (see [here](https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Serializer/Encoder/XmlEncoder.php#L369)): it doesn't even consider the possibility of normalizing it.
I'll create an issue to see if this can be considered as a bug.

Ener-Getick:
The last commit fix the tests :-)

Ener-Getick:
I opened https://github.com/symfony/symfony/pull/17984

Ener-Getick:
@lsmith77 @xabbuh Sorry to disturb you again, but this PR is ready for me so I would like to know if you agree with this points to know if I can merge it:
- removing the option ``exception_fallback_format`` in favor of matching request attributes
- overriding the jms ``HandlerRegistry`` until https://github.com/schmittjoh/serializer/pull/556 is merged
- Transform the ``FormInterface`` in a ``FormWrapper`` until https://github.com/symfony/symfony/pull/17984 is merged
- Overriding the default ``ExceptionListener`` to pass the raw exception to the controller

peshi:
@Ener-Getick It is a bit unclear how to use this.

Ener-Getick:
@peshi I'm not really good at writing docs :x
Moreover this feature is specific to the serializer used so this is complicated to write a generic documentation. But if you are motivated, feel free to open a PR :-)

peshi:
@Ener-Getick I think we share the same boat when it comes to docs. :)
Well, my first thought is.. 'where do I configure fos restbundle' to use my custom normalizer, that might be enough. Anyway - I need this feature as the wrapper has been scrapped, i'll see if I can make any sense of it during the week.

Ener-Getick:
Well FOSRestBundle doesn't have to be configured to use your custom normalizer, it depends on the serializer you uses (if you use the symfony serializer, that's really easy).
The best solution might be to add links to the documentation of the serializers we support.

Maxooo:
First of all, thank you for this improvement : this approach make sense in the Symfony ecosystem. You're right, it is very easy to create and add a normalizer via a service. But I agree with @peshi : the documentation is very unclear about how to configure the normalizer with the bundle. Could you please explain how to automatically use the normalizer for every catched exceptions in a REST controller action ?

Another question, the effect of the `fost_rest.exception.enabled` option is unclear too. I think the full configuration reference (http://symfony.com/doc/master/bundles/FOSRestBundle/configuration-reference.html) would gain to be annotated with some comments.

Thank you.

@peshi Any idea or suggestion since your last experience ?

peshi:
@Maxooo I have no clue where to start since there is no documentation. I'm also using jms serializer.
This is still very much on my todo list.

Ener-Getick:
@Maxooo if you use the symfony serializer, you can take a look at the [`UPGRADING-2.0.md` file](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/UPGRADING-2.0.md). It is explaining how to create a symfony normalizer dedicated to an exception.

If you're using the jms serializer (i advise you to migrate to the symfony serializer as it is not updated frequently), you have to create [an handler][http://jmsyst.com/libs/serializer/master/handlers) targetting the exception you want.

It would indeed be better to explain that in the docs but I didn't have the time to do that yet...

About `fost_rest.exception.enabled` it is enabling the FOSRestBundle exception controller. I agree, we should add more comments in the configuration reference section.

mvrhov:
Well if you use a jms serializer with BazingaHateoasBundle and a bunch of groups migrating to symfony serializer is not an option

Maxooo:
Thank you everybody for your helpful answers. The upgrading documentation is exactly what I was looking for.
Unfortunately I'm using JMS Serializer and the handler strategy seems to be the one I have to use. 
I also see an another solution, overriding the ExceptionController to implement the custom serialization and normalization : is this approach sounds good to you ?

Thank you again.

Ener-Getick:
@Maxooo if you override the ExceptionController, you'll end up changing everything as the serialization part is managed by the `ViewHandler`.
The easiest solution imo is to create a jms handler.

PR #1365, opened 2016-03-04
Fixes 1361 - Define the serializer groups parameter as a collection

### LINKED ISSUE #1361: Mandatory "Default" Serializer group in config
I just spent a few hours trying to understand why the serialization was always returning an empty json, so although I am not sure if this is a lack of documentation or a bug, I'd figure I'll open this issue for others to find.

I'm using FOSRestBundle 1.8 and JMSSerializationBundle on master.
If there is no groups defined in the configuration for the serializer, all properties will be excluded from the serialization.
So this minimum configuration is mandatory:

```
fos_rest:
    serializer:
        groups: ["Default"]
```
If this is not set, the serialization context will get an array with 1 element whose key is an empty string, probably due to this cast in the ViewHandler

```
public function setExclusionStrategyGroups($groups)
{
    $this->exclusionStrategyGroups = (array) $groups;
}
```
This in turns introduces 1 exclusion group and the shouldSkipProperty method of JMS\Serializer\Exclusion\GroupsExclusionStrategy will always return true.

I have not looked much deeper yet as to why this is passed an empty string instead of an empty array.
I guess this is due to some changes in 1.8 related to switching the serialization context from JMS to FOSRest.

jbenezech:
Fixes #1361

### DIFF
--- a/Tests/DependencyInjection/FOSRestExtensionTest.php
+++ b/Tests/DependencyInjection/FOSRestExtensionTest.php
@@ -334,6 +334,13 @@ class FOSRestExtensionTest extends \PHPUnit_Framework_TestCase
         $this->assertFalse($this->container->has('fos_rest.validator'));
     }
 
+    public function testDefaultSerializerGroupsIsArray()
+    {
+        $config = array();
+        $this->extension->load($config, $this->container);
+        $this->assertTrue(is_array($this->container->getParameter('fos_rest.serializer.exclusion_strategy.groups')));
+    }
+
     /**
      * Test that extension loads properly.
      */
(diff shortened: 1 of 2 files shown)


Tests/DependencyInjection/FOSRestExtensionTest.php
```diff
@@ -334,6 +334,13 @@ public function testValidatorAliasWhenDisabled()
         $this->assertFalse($this->container->has('fos_rest.validator'));
     }
 
+    public function testDefaultSerializerGroupsIsArray()
+    {
+        $config = [];
```
lsmith77: needs to be ``array()`` as we still support PHP 5.3

Ener-Getick:
Thank you @jbenezech !

PR #1370, opened 2016-03-05
Add fos_rest.exception.show_exceptions config param

munkie:
In order to be able manually enable/disable exception in error response.
Now kernel.debug value is used and it is not useful in some cases.
For example in functional test, thou it is run in debug mode, be able to check error response as it would be in prod environment (without exception).

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -29,6 +29,21 @@ use Symfony\Component\HttpFoundation\Response;
 final class Configuration implements ConfigurationInterface
 {
     /**
+     * Default debug mode value.
+     *
+     * @var bool
+     */
+    private $debug;
+
+    /**
+     * @param bool $debug
+     */
+    public function __construct($debug)
+    {
+        $this->debug = (bool) $debug;
+    }
+
+    /**
      * Generates the configuration tree.
      *
      * @return TreeBuilder
@@ -387,6 +402,9 @@ return $v; })
                         ->arrayNode('messages')
                             ->useAttributeAsKey('name')
                             ->prototype('boolean')->end()
+                        ->end()
+                        ->booleanNode('debug')
+                            ->defaultValue($this->debug)
                         ->end()
                     ->end()
                 ->end()
--- a/DependencyInjection/FOSRestExtension.php
+++ b/DependencyInjection/FOSRestExtension.php
@@ -33,7 +33,8 @@ class FOSRestExtension extends Extension
      */
     public function load(array $configs, ContainerBuilder $container)
     {
-        $config = $this->processConfiguration(new Configuration(), $configs);
+        $configuration = new Configuration($container->getParameter('kernel.debug'));
+        $config = $this->processConfiguration($configuration, $configs);
 
         $loader = new XmlFileLoader($container, new FileLocator(__DIR__.'/../Resources/config'));
         $loader->load('view.xml');
@@ -326,6 +327,13 @@ class FOSRestExtension extends Extension
                 ->replaceArgument(0, $config['exception']['codes']);
             $container->getDefinition('fos_rest.exception.messages_map')
                 ->replaceArgument(0, $config['exception']['messages']);
+
+            $container->getDefinition('fos_rest.exception.controller')
+                ->replaceArgument(2, $config['exception']['debug']);
+            $container->getDefinition('fos_rest.serializer.exception_normalizer.jms')
+                ->replaceArgument(1, $config['exception']['debug']);
+            $container->getDefinition('fos_rest.serializer.exception_normalizer.symfony')
+                ->replaceArgument(1, $config['exception']['debug']);
         }
 
         foreach ($config['exception']['codes'] as $exception => $code) {
--- a/Resources/config/exception_listener.xml
+++ b/Resources/config/exception_listener.xml
@@ -15,7 +15,7 @@
         <service id="fos_rest.exception.controller" class="FOS\RestBundle\Controller\ExceptionController">
             <argument type="service" id="fos_rest.view_handler" />
             <argument type="service" id="fos_rest.exception.codes_map" />  <!-- exception codes -->
-            <argument>%kernel.debug%</argument>
+            <argument /><!-- show exception -->
         </service>
 
         <service id="fos_rest.exception.twig_controller" class="FOS\RestBundle\Controller\TwigExceptionController" parent="fos_rest.exception.controller">
@@ -32,13 +32,13 @@
 
         <service id="fos_rest.serializer.exception_normalizer.jms" class="FOS\RestBundle\Serializer\Normalizer\ExceptionHandler">
             <argument type="service" id="fos_rest.exception.messages_map" /> <!-- exception messages -->
-            <argument>%kernel.debug%</argument>
+            <argument /><!-- show exception message -->
             <tag name="jms_serializer.subscribing_handler" />
         </service>
 
         <service id="fos_rest.serializer.exception_normalizer.symfony" class="FOS\RestBundle\Serializer\Normalizer\ExceptionNormalizer" public="false">
             <argument type="service" id="fos_rest.exception.messages_map" /> <!-- exception messages -->
-            <argument>%kernel.debug%</argument>
+            <argument /><!-- show exception message -->
             <tag name="serializer.normalizer" />
         </service>
     </services>
--- a/Tests/DependencyInjection/FOSRestExtensionTest.php
+++ b/Tests/DependencyInjection/FOSRestExtensionTest.php
@@ -54,6 +54,7 @@ class FOSRestExtensionTest extends \PHPUnit_Framework_TestCase
     {
         $this->container = new ContainerBuilder();
         $this->container->setParameter('kernel.bundles', ['JMSSerializerBundle' => true]);
+        $this->container->setParameter('kernel.debug', false);
         $this->extension = new FOSRestExtension();
         $this->includeFormat = true;
         $this->formats = [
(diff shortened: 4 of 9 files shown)


Ener-Getick:
You can launch your functional tests as if they were in a prod environment so I'm :-1: on this

munkie:
@Ener-Getick By functional tests i meant something like this http://symfony.com/doc/current/book/testing.html#your-first-functional-test - testing using Symfony WebTestCase and Crawler.
Most of the time tests are run with debug off (it speeds up execution).
But while developing, debug is switched on, so cache could be rebuild on every code change (not to run cache clear every time).
In this case error response JSON assertions will fail because they will also contain exception.

lsmith77:
also PR needs a rebase

lsmith77:
but I like the idea

munkie:
@lsmith77 @Ener-Getick Rebased master.
Also added _fos_rest.exception.debug_message_ option to enable/disable usage of exception message in error response manually.

munkie:
Rebased one more time

DependencyInjection/Configuration.php
```diff
@@ -388,6 +403,14 @@ private function addExceptionSection(ArrayNodeDefinition $rootNode)
                             ->useAttributeAsKey('name')
                             ->prototype('boolean')->end()
                         ->end()
+                        ->booleanNode('show_exception')
+                            ->treatNullLike($this->debug)
+                            ->defaultValue($this->debug)
+                        ->end()
+                        ->booleanNode('debug_message')
```
Ener-Getick: I don't see any benefits to have two fields.

DependencyInjection/Configuration.php
```diff
@@ -388,6 +403,14 @@ private function addExceptionSection(ArrayNodeDefinition $rootNode)
                             ->useAttributeAsKey('name')
                             ->prototype('boolean')->end()
                         ->end()
+                        ->booleanNode('show_exception')
+                            ->treatNullLike($this->debug)
```
Ener-Getick: This conversion should be done in the Extension instead IMO.

DependencyInjection/Configuration.php
munkie: One _show_exception_ forces to add exception to error response
Second _debug_message_ forces to use exception message instead of standard response status text
Maybe second config param should renamed to _show_exception_message_ to be more clear

DependencyInjection/Configuration.php
munkie: Dont see much benefits in moving to extension:
```
$showException = (null === $config['exception']['show_exception']) ? $container->getParameter('kernel.debug') : $config['exception']['show_exception'];
$container->getDefinition('fos_rest.exception.controller')->replaceArgument(2, $showException);
```
Looks way more complicated.
Actually i used same approach as in symfony docs - http://symfony.com/doc/current/cookbook/configuration/using_parameters_in_dic.html

Ener-Getick: ``$container->getParameter('kernel.debug')`` can be inlined in the xml file.
And the rest can be replaced by
```php
if (null !== $config['exception']['show_exception']) {
   $container->getDefinition('fos_rest.exception.controller')->replaceArgument(2, $config['exception']['show_exception']);
}
```

IMO that's clearer and that way we do not need to add a construtor argument to our ``Configuration`` class.

Ener-Getick:
As there is not a lot of use cases, I'm wondering if it wouldn't be easier to have a container parameter aliasing ``kernel.debug``.

munkie:
I noticed that now exception message would not be added to json/xml response with debug flag on, cause normalizer add only two fields to result data: code and message

munkie:
@Ener-Getick @lsmith77 So whats up with this PR ?

DependencyInjection/Configuration.php
```diff
@@ -388,6 +388,7 @@ private function addExceptionSection(ArrayNodeDefinition $rootNode)
                             ->useAttributeAsKey('name')
                             ->prototype('boolean')->end()
                         ->end()
+                        ->booleanNode('debug')->treatNullLike(false)->end()
```
Ener-Getick: IMO ``null`` should be treated as ``%kernel.debug``.

Tests/DependencyInjection/FOSRestExtensionTest.php
```diff
@@ -603,6 +646,62 @@ public function testSerializerExceptionNormalizer()
...
+        $this->assertSame($expectedValue, $definition->getArgument(1));
+    }
+
+    /**
+     * @dataProvider getSerializerExceptionNormalizerDebugMessageData
+     *
+     * @param array|bool  $config        fost_rest.exception config
+     * @param bool|string $expectedValue Expected service debug argument value
+     */
+    public function testJmsSerializerExceptionNormalizerDebugMessage($config, $expectedValue)
```
Ener-Getick: Why not merging this 3 tests?

Ener-Getick:
Except my messages fine for me :+1:

DependencyInjection/Configuration.php
munkie: Boolean node does not allow to use other value that boolean. Thats why i was injecting 'kernel.debug' value previously.
Without `treatNullLike(false)` _null_ will be treated as _true_

DependencyInjection/FOSRestExtension.php
```diff
@@ -326,11 +326,20 @@ private function loadException(array $config, XmlFileLoader $loader, ContainerBu
...
+                    ->replaceArgument(1, $config['exception']['debug']);
+                $container->getDefinition('fos_rest.serializer.exception_normalizer.symfony')
+                    ->replaceArgument(1, $config['exception']['debug']);
+            }
         }
 
         foreach ($config['exception']['codes'] as $exception => $code) {
             if (!is_numeric($code)) {
-                $config['exception']['codes'][$exception] = constant("\Symfony\Component\HttpFoundation\Response::$code");
+                $config['exception']['codes'][$exception] = constant("\\Symfony\\Component\\HttpFoundation\\Response::$code");
```
xabbuh: Can you revert this change please?

DependencyInjection/Configuration.php
xabbuh: I would also inject the `kernel.debug` parameter to the `Configuration` constructor (for example, in Symfony we do the same in the FrameworkBundle: https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/FrameworkBundle/DependencyInjection/FrameworkExtension.php#L197).

DependencyInjection/FOSRestExtension.php
```diff
@@ -326,11 +326,20 @@ private function loadException(array $config, XmlFileLoader $loader, ContainerBu
...
                 ->replaceArgument(0, $config['exception']['messages']);
+
+            if (isset($config['exception']['debug'])) {
+                $container->getDefinition('fos_rest.exception.controller')
+                    ->replaceArgument(2, $config['exception']['debug']);
+                $container->getDefinition('fos_rest.serializer.exception_normalizer.jms')
+                    ->replaceArgument(1, $config['exception']['debug']);
+                $container->getDefinition('fos_rest.serializer.exception_normalizer.symfony')
+                    ->replaceArgument(1, $config['exception']['debug']);
+            }
```
xabbuh: I would not wrap this in an `if` statement. Let's just always do the replacement here.

Resources/config/exception_listener.xml
```diff
@@ -15,7 +15,7 @@
         <service id="fos_rest.exception.controller" class="FOS\RestBundle\Controller\ExceptionController">
             <argument type="service" id="fos_rest.view_handler" />
             <argument type="service" id="fos_rest.exception.codes_map" />  <!-- exception codes -->
-            <argument>%kernel.debug%</argument>
+            <argument>%kernel.debug%</argument><!-- show exception -->
```
xabbuh: Let's use something like this for clarity:

```xml
<argument /> <!-- show exception -->
```

xabbuh:
I like this and I think we should do it in `1.8` (if I didn't miss anything this feature is fully backwards compatible).

DependencyInjection/Configuration.php
Ener-Getick: I told the opposite earlier but if you think this is better then why not.

xabbuh: I think so as it enables us to avoid the inconsistencies we would need to discuss otherwise here.

munkie:
@Ener-Getick @lsmith77 @xabbuh Done :sweat_smile:

DependencyInjection/Configuration.php
```diff
@@ -29,6 +29,23 @@
...
 {
     /**
+     * kernel.debug param value.
+     *
+     * @var bool
+     */
+    private $debug;
+
+    /**
+     * Configuration constructor.
```
xabbuh: I would omit this description as it doesn't add any value any just keep the parameter description.

DependencyInjection/Configuration.php
```diff
@@ -29,6 +29,23 @@
 final class Configuration implements ConfigurationInterface
 {
     /**
+     * kernel.debug param value.
```
xabbuh: This is an implementation detail which is only true when used in the extension class. We can just use something like "Whether or not to enable the debug mode".

xabbuh:
:+1:

DependencyInjection/Configuration.php
munkie: Fixed :+1:

Resources/config/exception_listener.xml
```diff
@@ -15,7 +15,7 @@
         <service id="fos_rest.exception.controller" class="FOS\RestBundle\Controller\ExceptionController">
             <argument type="service" id="fos_rest.view_handler" />
             <argument type="service" id="fos_rest.exception.codes_map" />  <!-- exception codes -->
-            <argument>%kernel.debug%</argument>
+            <argument/><!-- show exception -->
```
Ener-Getick: There's missing space here (should be ``<argument />``)

munkie: Fixed

Ener-Getick:
Thank you @munkie :-)
@xabbuh can you merge this in the 1.8 branch please?

xabbuh:
@Ener-Getick Do we want to treat this PR as a kind of bug fix?

Ener-Getick:
@xabbuh I don't think so but 1.8 is still in beta so shouldn't we merge this PR in it ?

Ener-Getick:
@xabbuh I created a 2.0 branch so let's just merge this in master.

PR #1373, opened 2016-03-07
Use ExceptionValueMap to store exception values (codes or message) instead of trait

munkie:
Use ExceptionValueMap to store map of exception values (codes or messages)
And use it instead of ClassMapHandlerTrait to get code or message associated with exception
In this case resolve logic is incapsulated in ExceptionValueMap and is not exposed into controller or serialize normalizer.

### DIFF
--- a/Resources/config/exception_listener.xml
+++ b/Resources/config/exception_listener.xml
@@ -14,7 +14,7 @@
 
         <service id="fos_rest.exception.controller" class="FOS\RestBundle\Controller\ExceptionController">
             <argument type="service" id="fos_rest.view_handler" />
-            <argument type="collection" /> <!-- exception codes -->
+            <argument type="service" id="fos_rest.exception.codes_map" />  <!-- exception codes -->
             <argument>%kernel.debug%</argument>
         </service>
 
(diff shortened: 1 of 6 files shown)


Resources/config/exception_listener.xml
```diff
@@ -22,14 +22,22 @@
             <argument type="service" id="templating.engine.twig" />
         </service>
 
+        <service id="fos_rest.exception.codes_map" class="FOS\RestBundle\Util\ExceptionValueMap">
+            <argument type="collection"/> <!-- exception codes -->
+        </service>
+
+        <service id="fos_rest.exception.messages_map" class="FOS\RestBundle\Util\ExceptionValueMap">
```
Ener-Getick: should be private (``public="false"``)

Util/ExceptionValueMap.php
```diff
@@ -11,24 +11,35 @@
...
+    /**
      * Resolves the value corresponding to a class from an array.
      *
      * @param string $class
-     * @param array  $map
      *
      * @return mixed|false if not found
      */
-    public function resolveValue($class, array $map)
+    public function resolveClass($class)
```
Ener-Getick: I don't think that's a good idea to have a function dedicated to the tests.

Ener-Getick:
Seems really good :+1:

Ener-Getick:
should be rebased on master

lsmith77:
do you want to put this into 2.0?

Ener-Getick:
Yeah it certainly don't affect people and it would be easier to respect bc

Util/ExceptionValueMap.php
munkie: So what's your suggestion: make it private ?

munkie:
Rebased.
Made map services private.
Made ExceptionValueMap::resolveClass method private

Ener-Getick:
Status: Reviewed

@lsmith77 what do you think about removing the required statuses ? It's really annoying to have to rebase a PR each another one is merged :-/

lsmith77:
the conflicts is I think nothing related to the required checks .. this is more that git cannot cleanly merge the source branch into the target branch

Ener-Getick:
@lsmith77 right here yeah but the other rebase were needed by github, nothing related to git. (and BTW it doesn't mean removing the protection against force pushing)

Ener-Getick:
Merged in #1380 thank you @munkie

Issue #1375, opened 2016-03-07
Error on TwigExceptionController


shakly:
Since  #1372 / #1374 , i have this issue when i get a NotFoundHttpException or MethodNotAllowedHttpException
```
ContextErrorException in TwigExceptionController.php line 29:
Notice: Undefined variable: format
```

Ener-Getick:
We should really test this class... and I should look a bit more to my PR before merging it...
Thank you, I'll search how to fix it :-)

Ener-Getick:
Can you take a look at https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1376 please?

PR #1376, opened 2016-03-07
Fix the TwigExceptionController

Ener-Getick:
see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1375


shakly:
Thanks you, it's work.
Test with the "format_listener", the response is send in good format

boast:
Looks good to me (used `"friendsofsymfony/rest-bundle": "dev-FIX"`) and the errors were gone. Thank you!

Issue #1381, opened 2016-03-09
Warning: ReflectionObject::__construct() expects parameter 1 to be object, null given


howardem:
Hello, 

I'm building a REST API layer using Symfony 2.8 and FOS Rest version 2.0.0-BETA2. Every time I send a request to the API I get a 500 error:

```
       "message": "Warning: ReflectionObject::__construct() expects parameter 1 to be object, null given",
        "class": "Symfony\Component\Debug\Exception\ContextErrorException",
        "trace": [
          {
            "namespace": "",
            "short_class": "",
            "class": "",
            "type": "",
            "function": "",
            "file": "C:\xampp\htdocs\cms-api\vendor\friendsofsymfony\rest-bundle\EventListener\ViewResponseListener.php",
            "line": 153,
            "args": []
          },
          {
            "namespace": "Symfony\Component\Debug",
            "short_class": "ErrorHandler",
            "class": "Symfony\Component\Debug\ErrorHandler",
            "type": "->",
            "function": "handleError",
            "file": null,
            "line": null,
            "args": [
              [
                "string",
                "2"
              ],
              [
                "string",
                "ReflectionObject::__construct() expects parameter 1 to be object, null given"
              ],
              [
                "string",
                "C:\xampp\htdocs\cms-api\vendor\friendsofsymfony\rest-bundle\EventListener\ViewResponseListener.php"
              ],
              [
                "string",
                "153"
              ],
              [
                "array",
                {
                  "template": [
                    "object",
                    "FOS\RestBundle\Controller\Annotations\View"
                  ],
                  "controller": [
                    "null",
                    null
                  ],
                  "action": [
                    "null",
                    null
                  ],
                  "arguments": [
                    "array",
                    []
                  ]
                }
              ]
            ]
          },
		  
```
This the resource I'm requesting. Traced the error and when the response is being dispatched it seems like the bundle implementation loses the information about the controller. This was working perfectly until Friday. At that time I was using FOS Rest 2.0.x-dev and everything was working fine. 

```
    /**
     * @REST\View()
     * @REST\Get("")
     */
    public function getDotInfosAction() {
        $dotInfos = $this->dotInfoService->findAll();
        return array('dotInfos' => $dotInfos);
    }	
```

This is my FOS configuration:
```
# FoS Rest
fos_rest:
    routing_loader:
        include_format: false    
    format_listener:
        enabled: true
        rules:
[...]

PR #1382, opened 2016-03-09
Throw an exception an exception when the SensioFrameworkExtraBundle is wrongly configured

Ener-Getick:
People doesn't seem to read the upgrading file so it's probably better to throw an exception when their configuration won't work.
see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1381


lsmith77:
needs a rebase

Ener-Getick:
@lsmith77 done.

PR #1383, opened 2016-03-10
fix configuration example

xabbuh:


Resources/doc/format_listener.rst
```diff
@@ -36,7 +36,7 @@ Setting ``priorities`` to a non-empty array enables Accept header negotiations.
                 # setting fallback_format to null means that in case of a priority mismatch the next rule will be considered
                 - { path: '^/admin', methods: [ 'GET', 'POST'], priorities: [ 'xml', 'html'], fallback_format: ~, prefer_extension: false }
                 # you can specifically target the exception controller
-                - { path: '^/api', priorities: [ 'xml', 'json'], fallback_format: xml, attributes: [ _controller: FOS\RestBundle\Controller\ExceptionController ], prefer_extension: false }
+                - { path: '^/api', priorities: [ 'xml', 'json'], fallback_format: xml, attributes: { _controller: FOS\RestBundle\Controller\ExceptionController }, prefer_extension: false }
```
javiereguiluz: Any idea why do the original examples add a white space only after the opening bracket? `[ 'xml', 'json']` Why not `[ 'xml', 'json' ]` or `['xml', 'json']`?

Ener-Getick: That's probably a mistake from me ...

lsmith77: https://github.com/FriendsOfSymfony/FOSRestBundle/commit/6fb5d3bfe995d29f08fd36e7d7f351e52e23f7df

xabbuh: By the way, was there a reason that we merged support for `attributes` in rules into the `master` branch and not in `1.8`? Looks to me like this would be backwards compatible.

Ener-Getick: This was part of #1358 which wasn't much reviewed so it's still time to backport things :-)

xabbuh: ah, thanks for the hint :)

PR #1385, opened 2016-03-10
fix loading route directory resources

### LINKED ISSUE #1377: Configuration for loading route annotations from a directory
Since this merged [PR](https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1339/files) we should be able to load all route annotations in a folder.

I tried this syntax : 

```yml
api:
    type: rest
    resource: "@AppBundle/Controller/"
```

But I still have this error message : 

> Can't find class for controller "@AppBundle/Controller/" in /home/me/src/AppBundle/Resources/config/api-routing.yml (which is being imported from "/home/me/app/config/routing.yml").

I fetched the version with the tag "2.0.0-BETA1"

```yml
"friendsofsymfony/rest-bundle": "2.0.0-BETA1",
```
And I checked if I had the class ``FOS\RestBundle\Routing\Loader\DirectoryRouteLoader`` and it's the case so I should not use the right syntax I think

Thanks for your help

xabbuh:
This fixes support for the bundle shortcut notation when loading routes from a directory (fixes #1377).

### DIFF
--- a/Resources/config/routing.xml
+++ b/Resources/config/routing.xml
@@ -17,17 +17,18 @@
 
     <services>
 
+        <service id="fos_rest.routing.loader.directory" class="FOS\RestBundle\Routing\Loader\DirectoryRouteLoader" public="false">
+            <argument type="service" id="file_locator" />
+            <argument type="service" id="fos_rest.routing.loader.processor" />
+            <tag name="routing.loader" />
+        </service>
+
         <service id="fos_rest.routing.loader.controller" class="%fos_rest.routing.loader.controller.class%">
             <argument type="service" id="service_container" />
             <argument type="service" id="file_locator" />
             <argument type="service" id="controller_name_converter" />
             <argument type="service" id="fos_rest.routing.loader.reader.controller" />
             <argument>%fos_rest.routing.loader.default_format%</argument>
-            <tag name="routing.loader" />
-        </service>
-
-        <service id="fos_rest.routing.loader.directory" class="FOS\RestBundle\Routing\Loader\DirectoryRouteLoader">
-            <argument type="service" id="fos_rest.routing.loader.processor" />
             <tag name="routing.loader" />
         </service>
 
--- a/Routing/Loader/DirectoryRouteLoader.php
+++ b/Routing/Loader/DirectoryRouteLoader.php
@@ -11,6 +11,7 @@
 
 namespace FOS\RestBundle\Routing\Loader;
 
+use Symfony\Component\Config\FileLocatorInterface;
 use Symfony\Component\Config\Loader\Loader;
 use Symfony\Component\Finder\Finder;
 use Symfony\Component\Routing\RouteCollection;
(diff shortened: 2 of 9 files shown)


Resources/config/routing.xml
```diff
@@ -17,6 +17,12 @@
 
     <services>
 
+        <service id="fos_rest.routing.loader.directory" class="FOS\RestBundle\Routing\Loader\DirectoryRouteLoader">
```
Ener-Getick: Any reason to not make this private?

xabbuh: At least no good reason :) (we already made the other loader services private in 2.0). I'll change this.

Routing/Loader/DirectoryRouteLoader.php
```diff
@@ -34,6 +37,10 @@ public function __construct(RestRouteProcessor $processor)
      */
     public function load($resource, $type = null)
     {
+        if (isset($resource[0]) && '@' === $resource[0]) {
+            $resource = $this->fileLocator->locate($resource);
```
Ener-Getick: why not always use the FileLocator ?

xabbuh: You mean relying on the concrete class?

Ener-Getick: I mean removing the ``if``. That way we would have clear messages in case the folder does not exist and we would be able to support relative paths.

xabbuh: No, this would file when a FQCN is passed to the loader because the `FileLocator` class then throws an exception.

Ener-Getick: Why not just catching it?

xabbuh: Not sure if that makes sense. I mean what advantage do we have by passing everything to the file locator?

Ener-Getick: After all I don't know ... The advantage would be to support relative paths but without using a specific type for directories I don't know how we could do that (the current path is not passed to the sub ``FileLoader``s, see [here](https://github.com/symfony/config/blob/master/Loader/FileLoader.php#L85-L87)).
So let's forget it :-/

Resources/config/routing.xml
```diff
@@ -26,11 +32,6 @@
             <tag name="routing.loader" />
         </service>
 
-        <service id="fos_rest.routing.loader.directory" class="FOS\RestBundle\Routing\Loader\DirectoryRouteLoader">
-            <argument type="service" id="fos_rest.routing.loader.processor" />
-            <tag name="routing.loader" />
-        </service>
-
         <service id="fos_rest.routing.loader.processor" class="%fos_rest.routing.loader.processor.class%" />
```
Ener-Getick: BTW did we forget to make the other services private?

Ener-Getick:
Status: Reviewed
:+1:

Resources/config/routing.xml
xabbuh: If I am not mistaken we did that in 2.0 (it's a BC break).

Ener-Getick: Oops sorry I didn't see that the target branch was 1.8..

PR #1388, opened 2016-03-12
Avoid use of __DIR__ in RestXmlCollectionLoader by using the FileLocator

chalasr:
This fixes the corresponding [major violation](https://insight.sensiolabs.com/projects/0be23389-2e85-49cf-b333-caaa36d11c62/analyses/271) on SL Insight.
(i.e. "Absolute path constants `__DIR__` and `__FILE__` should not be used")

Maybe someone with a better fix ?

### DIFF
--- a/Routing/Loader/RestXmlCollectionLoader.php
+++ b/Routing/Loader/RestXmlCollectionLoader.php
@@ -207,9 +207,10 @@ class RestXmlCollectionLoader extends XmlFileLoader
      */
     protected function validate(\DOMDocument $dom)
     {
-        $restRoutinglocation = realpath(__DIR__.'/../../Resources/config/schema/routing/rest_routing-1.0.xsd');
+        $configBasePath = realpath($this->locator->locate('../../../Resources/config'));
+        $restRoutinglocation = $configBasePath.'/schema/routing/rest_routing-1.0.xsd';
         $restRoutinglocation = rawurlencode(str_replace('\\', '/', $restRoutinglocation));
-        $routinglocation = realpath(__DIR__.'/../../Resources/config/schema/routing-1.0.xsd');
+        $routinglocation = $configBasePath.'/schema/routing-1.0.xsd';
         $routinglocation = rawurlencode(str_replace('\\', '/', $routinglocation));
         $source = <<<EOF
 <?xml version="1.0" encoding="utf-8" ?>


sstok:
It's more a warning then an actual problem, in this case I would (should) not be possible to replace the XSD right?

chalasr:
It's effectively not a problem or a bug, but it's the only major warning of the analysis.
I'm not sure to understand what you say, but because the path of the XSD is currently hard-coded, I think this commit doesn't involve any change in the way of how the XSD is loaded. 
Do I missed something ?

Ener-Getick:
This syntax looks harder to read to me ... Is there any gain to use the locator?

chalasr:
There is no real gain and no loss, but the locator is already accessible and until now never used in the class itself.
The primary goal is to avoid the only major warning on SLI (BTW change the medal from bronze to silver), and so use a better practice. 
If you have some input to make it more readable, I can adapt it.

Ener-Getick:
What about using ``@FOSRestBundle/Resources/config`` ?

Routing/Loader/RestXmlCollectionLoader.php
```diff
@@ -207,9 +207,10 @@ public function supports($resource, $type = null)
      */
     protected function validate(\DOMDocument $dom)
     {
-        $restRoutinglocation = realpath(__DIR__.'/../../Resources/config/schema/routing/rest_routing-1.0.xsd');
+        $configBasePath = realpath($this->locator->locate('../../../Resources/config'));
```
Ener-Getick: What about using ``@FOSRestBundle/Resources/config`` ?

chalasr:
That would be perfect but will not work out of a Symfony project, and so break the test suit, right ?

Routing/Loader/RestXmlCollectionLoader.php
chalasr: That would be perfect but will not work out of a Symfony project, and so break the test suit, right ?

Ener-Getick:
@chalasr yeah you're probably right...
What do you think @lsmith77 @xabbuh ?

lsmith77:
fone for me ..

xabbuh:
I see no real value in making this change (it only adds one additional method call that cannot be optimised). Imo this is an example where SensioLabs Insight gives a warning that can safely be ignored.

chalasr:
Agreed. For me, this warning could be ignored in most cases. Maybe I should contact the support  rather than trying to solve it.

Ener-Getick:
I guess we can close this ?

stof:
@Ener-Getick yes. this warning does not make sense in this case

PR #1394, opened 2016-03-16
document view annotation requirement for upgrades

xabbuh:
The change was introduced in #1308. Since #1382 the `ConfigurationCheckPass` will also throw an exception.

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -5,6 +5,15 @@ This document will be updated to list important BC breaks and behavioral changes
 
 ### upgrading to 2.0.0
 
+ * The SensioFrameworkExtraBundle view annotations must be enabled to use the `ViewResponseListener`:
+
+   ```yaml
+   # app/config/config.yml
+   sensio_framework_extra:
+       view:
+           annotations: true
+   ```
+
  * dropped support for the legacy ``Symfony\Component\Validator\ValidatorInterface``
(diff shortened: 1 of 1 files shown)


Ener-Getick:
Status: Reviewed
I was pretty sure that I already added it but apparently not... :+1:

UPGRADING.md
```diff
@@ -5,6 +5,15 @@ This document will be updated to list important BC breaks and behavioral changes
 
 ### upgrading to 2.0.0
 
+ * The SensioFrameworkExtraBundle `@View` annotation must be enabled to use the `ViewResponseListener`:
+
+   ````yaml
```
Ener-Getick: contains too much `

xabbuh: thanks, fixed

UPGRADING.md
```diff
@@ -5,6 +5,15 @@ This document will be updated to list important BC breaks and behavioral changes
 
 ### upgrading to 2.0.0
 
+ * The SensioFrameworkExtraBundle `@View` annotation must be enabled to use the `ViewResponseListener`:
```
Ener-Getick: In fact that's the SensioFrameworkExtraBundle ``@Template`` annotation. 
What about replacing it by ``the SensioFrameworkExtraBundle view annotations`?

xabbuh: Agreed, we use the same wording in the `ConfigurationCheckPass` exception message. So it makes sense to use that here too.

Ener-Getick:
I see nothing else to change :+1:

PR #1400, opened 2016-03-21
Fix #1399

### LINKED ISSUE #1399: Custom MIME type and versioning cause failures
FOSRestBundle 2.0 Beta 2
Symfony 2.7

I'm trying to implement a custom MIME type on my API, and require all clients to use it.  This part appears to work fine until I start to try to add versioning.

**config.yml**

```
fos_rest:
    view:
        view_response_listener: force
        mime_types:
            json: ['application/vnd.foo.api+json;version=3']
    versioning:
        enabled: true
        resolvers:
            media_type:
                enabled: true
                regex: '/version=(?P<version>[0-9]+)/'
    format_listener:
        enabled: true
        rules:
            -
                path: '^/bracket/'
                priorities: ['json']
                fallback_format: ~
                prefer_extension: false
```

When I send a request, I get a HTTP406 (Not Acceptable) response:

```
curl -H 'Accept: application/vnd.foo.api+json;version=3' http://localhost:8080/bar/20
```

Individually, both of these features appear to work.  For example, if I change my mime type as follows:

```
        mime_types:
            json: ['application/json;version=3']
```

And send a request:

```
curl -H 'Accept: application/json;version=3' http://localhost:8080/bar/20
```

Then I get the response I expect.

Also if I set it to my custom vendor descriptor:

```
        mime_types:
            json: ['application/vnd.foo.api+json;version=3']

curl -H 'Accept: application/vnd.foo.api+json' http://localhost:8080/bar/20
```

This also works (I get a HTTP200 response).

Ener-Getick:
Fix #1399

Should be merged in 2.0 (wrong target...)

### DIFF
--- a/DependencyInjection/FOSRestExtension.php
+++ b/DependencyInjection/FOSRestExtension.php
@@ -154,10 +154,6 @@ class FOSRestExtension extends Extension
                 'fos_rest.format_listener.rules',
                 $config['format_listener']['rules']
             );
-
-            if ($config['view']['mime_types']['enabled'] && !method_exists(Request::class, 'getMimeTypes')) {
-                $container->getDefinition('fos_rest.format_negotiator')->addArgument($config['view']['mime_types']['formats']);
-            }
         }
     }
 
--- a/EventListener/FormatListener.php
+++ b/EventListener/FormatListener.php
@@ -59,7 +59,7 @@ class FormatListener
             if (null === $format) {
                 $accept = $this->formatNegotiator->getBest('');
                 if (null !== $accept && 0.0 < $accept->getQuality()) {
-                    $format = $request->getFormat($accept->getType());
+                    $format = $request->getFormat($accept->getValue());
                     if (null !== $format) {
                         $request->attributes->set('media_type', $accept->getValue());
                     }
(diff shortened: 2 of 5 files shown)


DependencyInjection/FOSRestExtension.php
```diff
@@ -154,10 +154,6 @@ private function loadFormatListener(array $config, XmlFileLoader $loader, Contai
                 'fos_rest.format_listener.rules',
                 $config['format_listener']['rules']
             );
-
-            if ($config['view']['mime_types']['enabled'] && !method_exists(Request::class, 'getMimeTypes')) {
-                $container->getDefinition('fos_rest.format_negotiator')->addArgument($config['view']['mime_types']['formats']);
-            }
```
xabbuh: Are you sure that this is not needed?

Ener-Getick: Yeah the mime types are already registered by the mime type listener in the Request and the format negotiator fetch them from the request.

xabbuh: I see, so the fix basically is to call get `getValue()` instead of `getType()`?

Ener-Getick: That's a part of the fix, we also need https://github.com/symfony/symfony/pull/18255 to be merged.

Ener-Getick: In fact this PR is useless until https://github.com/symfony/symfony/pull/18255 is merged.

xabbuh:
symfony/symfony#18255 was merged

Ener-Getick:
@xabbuh thank you, I forgot this PR...

PR #1404, opened 2016-03-25
Add namespace in documentation examples that show a Controller class

### LINKED ISSUE #1266: Add namespace to documentation
The current documentation does not contain any namespaces.
http://symfony.com/doc/current/bundles/FOSRestBundle/2-the-view-layer.html

I think this namespace should be used in the example:
```
namespace AppBundle\Controller;
```

Or what's about ``ApiBundle\Controller``?

chalasr:
Added namespace `AppBundle\Controller` to all documentation examples that show a Controller class, according to symfony standards for documentation.

Closes #1266.

### DIFF
--- a/Resources/doc/2-the-view-layer.rst
+++ b/Resources/doc/2-the-view-layer.rst
@@ -27,6 +27,8 @@ which adds several convenience methods:
 
     <?php
 
+    namespace AppBundle\Controller;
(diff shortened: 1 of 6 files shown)


lsmith77:
could in theory also be back ported to 1.7/1.8

PR #1405, opened 2016-03-25
Fix QueryParam map with no constraints breaking

mikelinington:
When upgrading to [2.0.0-BETA2](https://github.com/FriendsOfSymfony/FOSRestBundle/releases/tag/), we ran into cases where we had this kind of annotation:

```php
/*
 * @QueryParam(name="filter", map=true, nullable=true)
 */
```

This broke as of #1191 (I believe) because using `Constraints\All` requires at least one constraint. Simple fix is to check to see if any actual constraints were specified.

### How to test

```bash
phpunit --filter ParamFetcherTest
```

### Notes/thoughts
This may be considered user error, but we also rely on the `map` property (previously `array`) to indicate that we expect to treat this parameter as an array. This is a case where `be liberal in what you accept, conservative in what you emit` informs my opinion that we should let consumers of this code use the `map=true` option liberally.

Tangential note is that I agree with the removal of type-checking in #1191; we actually are very flexible with these parameters, and expect either a comma-separated string or an actual array param.

### DIFF
--- a/Tests/Controller/Annotations/AbstractScalarParamTest.php
+++ b/Tests/Controller/Annotations/AbstractScalarParamTest.php
@@ -103,6 +103,8 @@ class AbstractScalarParamTest extends \PHPUnit_Framework_TestCase
     {
         $this->param->nullable = true;
         $this->param->map = true;
-        $this->assertEmpty($this->param->getConstraints());
+        $this->assertEquals(array(new Constraints\All(array(
+            'constraints' => [],
+        ))), $this->param->getConstraints());
     }
 }
--- a/Tests/Controller/Annotations/AbstractScalarParamTest.php
+++ b/Tests/Controller/Annotations/AbstractScalarParamTest.php
@@ -98,4 +98,11 @@ class AbstractScalarParamTest extends \PHPUnit_Framework_TestCase
             new Constraints\NotNull(),
         ))), $this->param->getConstraints());
     }
+
+    public function testArrayWithNoConstraintsDoesNotCreateInvalidConstraint()
+    {
+        $this->param->nullable = true;
+        $this->param->map = true;
+        $this->assertEmpty($this->param->getConstraints());
+    }
 }
--- a/Controller/Annotations/AbstractScalarParam.php
+++ b/Controller/Annotations/AbstractScalarParam.php
@@ -57,8 +57,9 @@ abstract class AbstractScalarParam extends AbstractParam
             $constraints[] = new NotBlank();
         }
 
-        // If the user wants to map the value
-        if ($this->map) {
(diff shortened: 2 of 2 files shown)


Ener-Getick:
What is the exact error you have ? I didn't find anything that could throw an error.

mikelinington:
@Ener-Getick Here's the exception being thrown:

```
Symfony\Component\Validator\Exception\MissingOptionsException: The options "constraints" must be set for constraint Symfony\Component\Validator\Constraints\All
```

You can see this by creating a controller with this query param and making a request to the generated endpoint:
```php
/*
 * @QueryParam(name="filter", map=true, nullable=true)
 */
public function cgetAction(ParamFetcher $fetcher) {
}
```

Or reverting the change I made in `AbstractScalarParam` and running the provided unit test `(AbstractParamTest::testNotNullConstraint`). I can add a test case to `ParamFetcherTest` if you'd like to see a functional test around this.

Let me know if anything else is unclear!

Ener-Getick:
Thanks for your detailed answer. Wouldn't it be better to use the following then? 
```php
new All(array('constraints' => $constraints));
```

mikelinington:
@Ener-Getick good call, that's a bit more succinct! I just pushed that change, let me know if you want anything else on this.

Tests/Controller/Annotations/AbstractScalarParamTest.php
```diff
@@ -98,4 +98,11 @@ public function testConstraintsTransformWhenParamIsAnArray()
             new Constraints\NotNull(),
         ))), $this->param->getConstraints());
     }
+
+    public function testArrayWithNoConstraintsDoesNotCreateInvalidConstraint()
```
Ener-Getick: This test is useless now

mikelinington: @Ener-Getick I just pushed up a change to test that it's creating the constraint with an empty `constraints` key, would you prefer I just remove the test entirely?

Ener-Getick: That's fine as us :-)

Ener-Getick:
Sorry, I forgot your PR for a while... thank you @mikelinington :-)

Ener-Getick:
@xabbuh I forgot again to merge this into 2.0 instead of master... any idea how to backport it ? :/

xabbuh:
@Ener-Getick Cherry picking the needed commits should solve the issue. Can this also be applied to 1.8?

Ener-Getick:
@xabbuh no the ``ParamFetcher`` was refactored in 2.0.

Issue #1408, opened 2016-03-28
[View] Add View Format Constant


ihsanudin:
Add view format constant to make it easier

PR #1409, opened 2016-03-28
Add view format constants

ihsanudin:
Add view format constants

purpose this issue #1408

### DIFF
--- a/View/View.php
+++ b/View/View.php
@@ -26,7 +26,7 @@ class View
     const FORMAT_HTML = 'html';
     const FORMAT_JSON = 'json';
     const FORMAT_XML = 'xml';
-    
+
     /**
      * @var mixed|null
      */
--- a/View/View.php
+++ b/View/View.php
@@ -23,6 +23,10 @@ use Symfony\Component\Templating\TemplateReferenceInterface;
  */
 class View
 {
+    const FORMAT_HTML = 'html';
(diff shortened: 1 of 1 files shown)


Ener-Getick:
I'm not convinced that this is useful as the value of this constants will never change and is very intuitive.

sstok:
An FORMAT_HTML constant for 'html', that's a bit overdoing it :) for status codes I would say yes but in this case :-1:

ihsanudin:
@sstok : it's just my case, i don't know how many format can be supported by FOSRestBundle and we need to know that

Ener-Getick:
@ihsanudin in what would this help you? do you have any real examples?

ihsanudin:
This is my example

```lang=php
$view = View::create()
            ->setStatusCode(Response::HTTP_OK)
            ->setFormat(View::FORMAT_HTML)
            ->setTemplate('template.html.twig')
        ;
```
With that constants, My IDE can suggest (auto complete) the view format. It's more efficient and easier for me

xabbuh:
Usually I would agree that adding constants to improve the developer experience through auto-completion was a nice improvement. However, in this concrete example I feel that this might lead to the misperception that we only support the given three formats which might confuse people.

Ener-Getick:
Then I guess we can close this.
Thank you for your contribution anyway @ihsanudin :-)

PR #1410, opened 2016-03-31
fall back to core Symfony exception listener outside configured REST zones

xabbuh:
@jovobe discovered that core Symfony exception listener was removed when the one from the FOSRestBundle was registered. This leads to errors when an exception is thrown for a request that does not match a configured REST zone meaning that the exception is entirely ignored by the exception handling process.

### DIFF
--- a/DependencyInjection/Compiler/TwigExceptionPass.php
+++ b/DependencyInjection/Compiler/TwigExceptionPass.php
@@ -23,10 +23,6 @@ final class TwigExceptionPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container)
     {
-        if ($container->has('fos_rest.exception_listener') && $container->has('twig.exception_listener')) {
-            $container->removeDefinition('twig.exception_listener');
-        }
-
         if (!$container->has('templating.engine.twig')) {
             $container->removeDefinition('fos_rest.exception.twig_controller');
         }
--- a/EventListener/ExceptionListener.php
+++ b/EventListener/ExceptionListener.php
@@ -12,9 +12,10 @@
 namespace FOS\RestBundle\EventListener;
 
 use FOS\RestBundle\FOSRestBundle;
+use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\Event\GetResponseForExceptionEvent;
 use Symfony\Component\HttpKernel\EventListener\ExceptionListener as HttpKernelExceptionListener;
-use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\KernelEvents;
 use Symfony\Component\HttpKernel\Log\DebugLoggerInterface;
 
 /**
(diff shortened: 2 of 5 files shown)


Ener-Getick:
Do we really need this, isn't it better to rely on listeners priority?

DependencyInjection/Compiler/TwigExceptionPass.php
```diff
@@ -24,6 +25,16 @@
...
+            // wrap our exception listener around twig exception listener
+            $wrappedExceptionListener = $container->findDefinition('twig.exception_listener');
+            $wrappedExceptionListener->setPublic(false);
+            $wrappedExceptionListener->setTags(array());
+
+            $container->setDefinition('fos_rest.wrapped_exception_listener', $wrappedExceptionListener);
+            $container->findDefinition('fos_rest.exception_listener')->addArgument(
+                new Reference('fos_rest.wrapped_exception_listener')
+            );
             $container->removeDefinition('twig.exception_listener');
```
Ener-Getick: In fact I don't see why we needed this on first place.

xabbuh: Do you remember why you added that in #1358?

Ener-Getick: What do you mean? This wasn't added in this pr.

Ener-Getick: This was introduced in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/908

Ener-Getick: It seems to me that we can safely remove this but I'll investigate a bit more tomorrow to be completely sure.

Ener-Getick:
@xabbuh do you want to me to open a PR as I apparently forgot several changes ?
most of the doc introduced in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/908/files is irrelevant now (it isn't possible anymore to use the twig config to enable the FOSRestBundle exception listener).

DependencyInjection/Compiler/TwigExceptionPass.php
xabbuh: Ah indeed, you only introduced the new class. I'll check with @jovobe tomorrow the failing example we had.

xabbuh:
You mean updating the docs? I think that would be good (and we should check if we can trigger a deprecation in 1.8 in case someone use this way).

DependencyInjection/Compiler/TwigExceptionPass.php
xabbuh: @Ener-Getick Indeed, we can simply register our listener with a higher priority and do not remove the core one.

Ener-Getick:
LGTM :+1:
Status: Reviewed

xabbuh:
Thank you @jovobe.

Issue #1436, opened 2016-04-19
Fix multiple version for one controller route condition


boyez-joggin:
As I found out, it seems that having multiple version in one Version annotation for a Controller breaks the routing condition.

I found the condition is generated like this before evaluating to write cache:
```
request.attributes.get("version") == (<version> or <version> or <version>)
```

this evaluates to true during runtime for any non empty string and breaks the version limitation.

I corrected the code to generate with _in_ operator like this:
```
request.attributes.get("version") in [<version>, <version>, <version>]
```

dmaicher:
How exactly did the annotations your controller look like?

boyez-joggin:
My annotations were like this:
![symfony_rest_versions](https://cloud.githubusercontent.com/assets/18553271/14674754/cf6e9542-0705-11e6-8cd8-458c000a44c4.png)

And my final condition in the auto-generated url matcher is that:
![symfony_url_matcher](https://cloud.githubusercontent.com/assets/18553271/14674815/1ea6fb4a-0706-11e6-87ef-5fe27bc5b612.png)

Ener-Getick:
Fixed by https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1438

PR #1437, opened 2016-04-19
Fix multiple version in annotation condition handling

boyez-joggin:
By a _in []_ syntax instead of the current
For #1436

### DIFF
--- a/Routing/Loader/Reader/RestActionReader.php
+++ b/Routing/Loader/Reader/RestActionReader.php
@@ -352,16 +352,16 @@ class RestActionReader
         $condition = $annotation->getCondition();
 
         if (!empty($this->versions)) {
-            $versionCondition = "request.attributes.get('version') == (";
+            $versionCondition = "request.attributes.get('version') in [";
             $first = true;
             foreach ($this->versions as $version) {
                 if (!$first) {
-                    $versionCondition .= ' or ';
+                    $versionCondition .= ', ';
                 }
                 $versionCondition .= '\''.$version.'\'';
                 $first = false;
             }
-            $versionCondition .= ')';
+            $versionCondition .= ']';
             $condition = $condition ? '('.$condition.') and '.$versionCondition : $versionCondition;
         }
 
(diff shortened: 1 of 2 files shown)


dmaicher:
This is also fixed in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1438

Routing/Loader/Reader/RestActionReader.php
```diff
@@ -352,16 +352,16 @@ private function getCondition(\ReflectionMethod $method, RouteAnnotation $annota
         $condition = $annotation->getCondition();
 
         if (!empty($this->versions)) {
-            $versionCondition = "request.attributes.get('version') == (";
+            $versionCondition = "request.attributes.get('version') in [";
             $first = true;
             foreach ($this->versions as $version) {
```
dmaicher: I don't see how this loop could generate something like you mentioned in https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1436

`((<version> or <version>) or <version>)`

boyez-joggin: Sorry there were one set of parenthesis which came from symfony language auto generation.
The real result is this:

```
request.attributes.get('version') == (<version> or <version> or <version>)
```

which then auto-generates to:

```
request.attributes.get('version') == ((<version> || <version>) || <version>)
```
like I added to the related issue #1436 

So with my fix we get:
```
request.attributes.get('version') in [<version>, <version>, <version>]
```
which auto-generates to:
```
in_array(request.attributes.get('version'), array(0 => <version>, 1 => <version>, 2 => <version>))
```

Sorry for the mix, my bad :x

dmaicher: Ok now I undertand :+1:  Thx for the explanation. The old solution with the `or` is totally wrong indeed. I also fixed that problem in my other PR though:

https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1438

Ener-Getick:
Closed in favor of https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1438

PR #1442, opened 2016-04-24
Fix the tests

Ener-Getick:
see https://github.com/symfony/symfony/pull/18630, related to the debug mode in our functional tests.


xabbuh:
@Ener-Getick Wouldn't the fix in Symfony be enough?

Ener-Getick:
@xabbuh yes it would but I think it's a bad practice to run tests in normal mode.

Ener-Getick:
https://github.com/symfony/symfony/pull/18630 was merged so we do not need this anymore.

PR #1448, opened 2016-05-01
ignore leading underscores in base normalizer

xabbuh:
In #1267, a new normalizer class was introduced that ignores underscores
when converting snake case to camel case. However, the same change was
made to the base class. This means that both classes act in exactly the
same way. Furthermore, this change was a BC break as the behavior of the
base normalizer class was changed without fixing an actual bug.


Ener-Getick:
:+1: but the tests fail.

xabbuh:
@Ener-Getick We had an issue that led to array keys being normalised more than once (thus transforming `__username` to `Username` instead of `_Username`). This should work now.

lsmith77:
good catch .. +1

PR #1451, opened 2016-05-01
Use SplObjectStorage instead of spl_object_hash

Ener-Getick:
See [this benchmark](http://technosophos.com/2009/05/29/set-objects-php-arrays-vs-splobjectstorage.html)

### DIFF
--- a/Request/ParameterBag.php
+++ b/Request/ParameterBag.php
@@ -23,38 +23,39 @@ use Symfony\Component\HttpFoundation\Request;
 final class ParameterBag
 {
     private $paramReader;
-    private $params = array();
+    private $params;
 
     public function __construct(ParamReaderInterface $paramReader)
     {
         $this->paramReader = $paramReader;
+        $this->params = new \SplObjectStorage();
     }
 
     public function getParams(Request $request)
     {
-        $requestId = spl_object_hash($request);
-        if (!isset($this->params[$requestId]) || empty($this->params[$requestId]['controller'])) {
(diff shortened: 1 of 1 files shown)


xabbuh:
I am 👍 for the change, but I do not think that this counts as a bug fix and so should be merged into 1.8 instead.

mvrhov:
in php 7 the speed difference is it's not that drastic anymore, spl is 50% faster in TTF
and uses 50% less memory.

Whats more interesting is that the test below was run on i7 6700. And the author run it on Core2Duo. 7 generations and only ~33% difference.

```
SplObjectStorage:
Time to fill: 0.063576936722.
Time to check: 0.079049000000.
Memory: 15121864

Arrays:
Time to fill: 0.135654926300.
Time to check: 0.055525000000.
Memory: 30318672
```

stof:
There is a big difference here though: you keep a reference to the Request object, which is not released at the end of the request. Sso this will leak (even more) memory when handling several requests with the same kernel (which can happen in tests, or when using Symfony in php-pm for instance).
Note that the current code already leaks the params, but the new code also forces to leak the Request object and everything it contains.

Ener-Getick:
You're right @stof, maybe we should add an event listener on kernel.request_finished to release memory what do you think?

Ener-Getick:
@stof in fact the memory leak won't be resolved entirely just by using a hash to represent the request, the parameters will still be preserved.
So IMO the event listener is the best solution we have.

Ener-Getick:
I close this for now (it can reported to 2.1 or later).

PR #1452, opened 2016-05-01
[1.7] Improve the ViolationFormatter output message

Ener-Getick:
This exception message looks weird to me and hard to understand, WDYT?

### DIFF
--- a/Tests/Util/ViolationFormatterTest.php
+++ b/Tests/Util/ViolationFormatterTest.php
@@ -43,7 +43,7 @@ class ViolationFormatterTest extends \PHPUnit_Framework_TestCase
 
         $formatter = new ViolationFormatter();
         $this->assertEquals(
-            "Query parameter foo value 'bar' violated a constraint (expected message)",
+            'Query parameter "foo" of value "bar" violated a constraint "expected message"',
             $formatter->format($param, $violation)
         );
     }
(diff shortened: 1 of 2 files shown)


Ener-Getick:
Well maybe we can consider this as a BC break (users might relly on it) so I don't know...

lsmith77:
maybe just put it into 1.8?

Ener-Getick:
:+1:

xabbuh:
I am okay for 1.7 too. You shouldn't rely on a third-party's exceptions messages imo.

Ener-Getick:
That was the only way to match this exception until now so i wouldn't be surprise if someone rely on its message.
As @lsmith77 i think this is safer to put this in 1.8 and to document the change.

BTW it would be great to create more custom exceptions in 2.0.

xabbuh:
Hm, that's true. But then we can only make this change in 2.0. I agree with you about using more custom exceptions (though that doesn't help here as the exception directly leads to a response).

Ener-Getick:
You're right, i close this in favor of https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1453 then.
BTW as i'm thinking about it, we could provide a default configuration for the exception controller to convert common exceptions that are safe (including this one).

PR #1453, opened 2016-05-01
Replace the ViolationFormatter by a new exception

Ener-Getick:
This is particularly useful as now we can normalize exceptions fields (we can imagine have a cleaner output for this exception in the future).


lsmith77:
I guess we already test the transformation of exceptions else where?

Ener-Getick:
yeah @lsmith77 the exception transformation is managed by the ExceptionController.
i'll just have to test the exception message.

PR #1463, opened 2016-05-09
deprecate the Param class

xabbuh:


xabbuh:
Most of the methods coming from the `ParamInterface` are pretty useless in 1.8, but backporting the interface (and its derived abstract classes) while deprecating the `Param` class should help users migrating to 2.0.

lsmith77:
there appears to be a conflict here .. (maybe due to the merge I just did)

xabbuh:
Indeed, I rebased here.

PR #1467, opened 2016-05-09
[#1463] add empty implementation for required method

xabbuh:
The new `ParamInterface` requires implementing classes to provide the
body of the `getValue()` method. This adds an empty implementation to
the `Param` class to prevent BC breaks for existing classes that extend
the `Param` class.

### DIFF
--- a/Controller/Annotations/Param.php
+++ b/Controller/Annotations/Param.php
@@ -11,6 +11,8 @@
 
 namespace FOS\RestBundle\Controller\Annotations;
 
+use Symfony\Component\HttpFoundation\Request;
+
 /**
  * Represents a parameter that can be present in GET or POST data.
  *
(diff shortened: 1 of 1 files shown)


Controller/Annotations/Param.php
```diff
@@ -51,4 +53,9 @@ public function getKey()
     {
         return $this->key ?: $this->name;
     }
+
+    public function getValue(Request $request, $default)
+    {
+        return null;
```
Ener-Getick: this ``null`` doesn't pass the StyleCi validation, should be only ``return;``

xabbuh: fixed

Ener-Getick:
Thank you @xabbuh

PR #1468, opened 2016-05-10
Invalid parameter name change.

nurolopher:
Just made the doc parameter name consistent with actual parameter name.


Ener-Getick:
👍  but should be applied on 1.8

Ener-Getick:
In fact it was already fixed in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1454, but thank you anyway !

PR #1460, opened 2016-05-07
Backport #1453

xabbuh:

PR #1470, opened 2016-05-11
Merge branch 1.8 into 2.0

Ener-Getick:


Tests/Functional/app/Debug/bundles.php
```diff
@@ -9,9 +9,9 @@
  * file that was distributed with this source code.
  */
 
-return [
+return array(
```
Ener-Getick: I reverted some of this changes as they only make merges harder...

UPGRADING.md
```diff
@@ -41,6 +41,14 @@ This document will be updated to list important BC breaks and behavioral changes
  * removed the ``format_negotiator`` option ``exception_fallback_format``; you can match the ``ExceptionController`` thanks to the ``attributes`` option instead
  * `View::setSerializationContext` and `View::getSerializationContext` have been removed. Use `View::setContext` and `View::getContext` together with the new Context class instead.
 
+### upgrading to 1.8 (unreleased)
```
Ener-Getick: maybe we should only keep the ``2.0`` section in the ``2.0`` branch?

xabbuh: I agree. We can remove this file after #1460 has been merged up.

PR #1471, opened 2016-05-13
handle throwable

fprochazka:


xabbuh:
I am not sure if we should update the docs this way. If this is your application, you will know if you run it on PHP 7 and in this case you could simply stop catching `\Exception` but catch `\Throwable` instead (there is no need to have both). Maybe we should simply add a comment that you can catch `\Throwable` instead when using PHP 7.

fprochazka:
Oh, you're right, I didn't notice it's only the doc. So this can be closed then?

Ener-Getick:
I'm :-1: for this PR as it targets docs but the others are relevant ;-)

fprochazka:
Ok, np :)

PR #1472, opened 2016-05-13
handle throwable

fprochazka:


Ener-Getick:
BTW, maybe we should also remove the type hint on [this method](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/2.0/Controller/ExceptionController.php#L64)

Resources/doc/examples/RssHandler.php
```diff
@@ -65,6 +65,10 @@ public function createResponse(ViewHandler $handler, View $view, Request $reques
             $content = $this->createFeed($view->getData());
             $code = Codes::HTTP_OK;
         } catch (\Exception $e) {
+        } catch (\Throwable $e) {
```
Ener-Getick: It depends on what we decide in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1471 but this may have to be removed

xabbuh:
Removing the type hint would be a BC break.

Ener-Getick:
@xabbuh yeah but the ``2.0`` is still in beta and I think it's something we should do now as php7 will become more and more used and we could have issues related to this in the future.

fprochazka:
I've dropped the change in the doc FYI

Ener-Getick:
Status: reviewed
LGTM :+1:

xabbuh:
@Ener-Getick Oh well, given that 1.8 even doesn't have the type hint, you are actually even more right and we can simply drop it again in 2.0.

xabbuh:
see #1477

Ener-Getick:
#1473 is merged so no need for this anymore :-)
Thank you @fprochazka

Issue #217, opened 2012-03-29
look into how best way to support streaming responses


lsmith77:
this will need to be handled likely in View, ViewHandler and ViewResponseListener

for the later this is what SensioFrameworkExtraBundle

```
        if (!$request->attributes->get('_template_streamable')) {
            $event->setResponse($templating->renderResponse($template, $parameters));
        } else {
            $callback = function () use ($templating, $template, $parameters) {
                return $templating->stream($template, $parameters);
            };


            $event->setResponse(new StreamedResponse($callback));
        }
```

Wirone:
+1 this is crucial :exclamation:

We have API method for generating report. It can be returned as `json` or `xml` with simple array passed from controller. However, we would like to serve `csv` and `xlsx` formats, then controller would return `StreamedResponse`. Now we have to register custom handler (in development now, I hope it will work..). Would be so easier with this feature.

Wasn't it implemented since 2012 or I'm missing something?

jdwilliams19:
I'd like to see this as well, though I think it's most useful in conjunction with a serializer that supports streams as well, and JMS/Serializer does not.

PR #1476, opened 2016-05-14
Feature/add pagination and streaming support

jdwil:
#217 This PR adds streaming support as well as automatic pagination. Both require a Doctrine QueryBuilder to be returned as the View data, so Doctrine DBAL is required to use the feature, but no dependency is added to rest-bundle.

If a QueryBuilder is returned in the View, and the limit and offset parameters are in the ParamFetcher (both keys are configurable in View) then it will automatically paginate the results and let rest-bundle deliver the content normally. It also automatically sets the "Content-Range" header in the response for user by front-end pagers.

If no pagination parameters are present, then the results of from the QueryBuilder will be streamed. When being streamed, the rows are fetched one at a time, encoded using the encoder or a callback set in View, and pushed to stdout. There is also a header and footer callback for more customizable output.


EventListener/ViewResponseListener.php
```diff
@@ -96,6 +98,17 @@ public function onKernelView(GetResponseForControllerResultEvent $event)
         if (null === $view->getFormat()) {
             $view->setFormat($request->getRequestFormat());
         }
+        
+        $range = null;
+
+        if (class_exists("Doctrine\DBAL\Query\QueryBuilder")
```
Ener-Getick: The ``class_exists`` is useless here

xabbuh:
I am not convinced that this feature belongs to the bundle. If it is currently not possible to easily implement this in userland code, I think we should rather think about exposing clean extension points to make it possible.

jdwilliams19:
@xabbuh That's a fair point. I can refactor this to take a custom interface rather than a Doctrine QueryBuilder. Would it be okay to add an implementation of the interace for Doctrine to rest-bundle? Others could add implementations for Propel and other dbal's...

xabbuh:
I more thought of an event that we could dispatch here which would allow to modify the view before it was transformed into a response.

jdwilliams19:
Roger that. I've got a new version using a StreamableInterface and PageableInterface. I'll try to get the tests cleaned up this week and push it up. Then we can talk about the event idea... not sure I completely follow.

jdwilliams19:
So, this is what the Pageable/Streamable interfaces look like. I did make a vendor-specific Doctrine dao in the DBAL code, which I've found useful. Before I write more tests, can I get some feedback on this approach? I've been using this code in a development project for a few weeks and it makes hooking up to a data table front end, like Bootstrap Table or jQuery DataTables, quite easy.

Ener-Getick:
Sorry for the late answer.. I think we should at first implement @xabbuh proposition and then eventually include new classes to easily manage pagination and streaming even if i'm not sure this belongs to this bundle.

jdwil:
@Ener-Getick @xabbuh Roger that. Thanks for the feedback. I'll look into refactoring this as I have time and creating a separate bundle.

hipio:
Any update on this? A REST API should have pagination IMO.

jdwil:
@hipio Sorry, I haven't had any time to work on this. I've been using my FOS Rest Bundle fork in my current side-project, which has the functionality described here. I don't want to roll it into production that way so I will circle back to this at some point.

hipio:
@jdwil Ah, I see, no worries, just wondering as I'm having to add pagination as well. Using a fork isn't something I want to do though so I guess I'll have to wait or implement something myself without touching core Symfony files.

c410-f3r:
Any news about this feature?

vasilvestre:
Any news ? Fix checks ? Or already up ?

Neirda24:
Any news ? Seems to me that any good API should have a pagination support. Keep it up ;)

Tobion:
Closing as outdated.

PR #1477, opened 2016-05-14
remove \Exception type hint

xabbuh:
Without the type hint we can also support objects implementing the
\Throwable interface when using PHP 7.

### DIFF
--- a/Controller/ExceptionController.php
+++ b/Controller/ExceptionController.php
@@ -54,14 +54,14 @@ class ExceptionController
      * Converts an Exception to a Response.
      *
      * @param Request                   $request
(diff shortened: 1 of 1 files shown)


Ener-Getick:
:+1: should also be added to the symfony normalizer (won't work for jms anyway as it doesn't support interfaces for handlers)

Ener-Getick:
In fact the symfony kernel doesn't support \Throwable so I don't know what we should do... (see https://github.com/symfony/symfony/blob/master/src/Symfony/Component/HttpKernel/HttpKernel.php#L67)

Ener-Getick:
Thank you @xabbuh

PR #1480, opened 2016-05-14
remove no longer needed $debug property

xabbuh:


Ener-Getick:
Oops you're right, i thought it was a fix we made in 1.8 :+1:

PR #1489, opened 2016-05-17
decrease exception normalizer priority

### LINKED ISSUE #1488: Decrease the priority of the exception normalizer
In #1358 we added a new `ExceptionNormalizer` which will be used to normalize exceptions with the Symfony Serializer component. However, one now has to register their own normalizers with a higher priority to make sure that they are executed before our normalizer which is a bit inconvenient. Thus, I suggest that we decrease the priority (let's say -10) of our listeners to make this a bit easier to customise (no need to specify a priority for custom normalizers).

xabbuh:
This makes it easier for custom normalizers to be executed before the
one from FOSRestBundle as one would not need to specify a priority
anymore.

This fixes #1488.


Ener-Getick:
What about also updating [the one converting forms](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/2.0/Resources/config/serializer.xml#L21) ?

Ener-Getick:
Thank you @xabbuh

PR #1490, opened 2016-05-18
Fix link in documention to UPGRADING-2.0.md

oliveradria:


Ener-Getick:
Thank you @oliveradria !

PR #1506, opened 2016-05-31
Deprecate the maxDepth context attribute in favor of enableMaxDepth

Ener-Getick:
Both serializers don't allow to define maxDepth at runtime so it shouldn't have been called this way...
I'm targeting master as this may break apps using 2.0.

### DIFF
--- a/Context/Context.php
+++ b/Context/Context.php
@@ -37,6 +37,10 @@ final class Context
     /**
      * @var bool
      */
+    private $isMaxDepthEnabled;
+    /**
+     * @var bool
+     */
     private $serializeNull;
 
     /**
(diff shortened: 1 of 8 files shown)


stof:
> Both serializers don't allow to define maxDepth at runtime

What do you mean here ?

Ener-Getick:
@stof they both only allow to [enable max depth checks](https://github.com/symfony/symfony/pull/17113/files#diff-cd90b97b50abf20fa94468689e7486afR24).
To customize the maximal depth, you must edit your entity metadata.

lsmith77:
fine for me .. guess there are some style ci issues to fix

Ener-Getick:
@lsmith77 they were fixed in another pr, this one should be rebased

Ener-Getick:
done.

PR #1518, opened 2016-06-17
Version annotation exists only in branch 2.0

hpatoio:
Let's remove this part from 1.8 doc


stof:
note for maintainers: this commit should be reverted in the master branch after merging 1.8 into master

xabbuh:
@lsmith77 @Ener-Getick I wonder if we should not rather backport #1182 to the `1.8` branch instead. The PR looks fully backwards compatible to me. What do you think?

lsmith77:
I am fine with not backporting the version annotation

xabbuh:
Thank you @hpatoio.

PR #1520, opened 2016-06-21
Changed a type hint in FormErrorHandler class

jhcasado:
Changed \ArrayObject to array in a private method parameter.

### DIFF
--- a/Serializer/Normalizer/FormErrorHandler.php
+++ b/Serializer/Normalizer/FormErrorHandler.php
@@ -86,7 +86,7 @@ class FormErrorHandler extends JMSFormErrorHandler
         return $result;
     }
 
-    private function adaptFormArray(\ArrayObject $serializedForm, Context $context = null)
+    private function adaptFormArray(array $serializedForm, Context $context = null)
     {
         $statusCode = $this->getStatusCode($context);
         if (null !== $statusCode) {


Ener-Getick:
What's the reason behind this ? The jms serializer uses an ArrayObject [here](https://github.com/schmittjoh/serializer/blob/master/src/JMS/Serializer/Handler/FormErrorHandler.php#L132)

jhcasado:
I'm sorry, you are right,

I checked dependencies and I don't know why the project had the jms version fixed in composer:

`"jms/serializer-bundle": "^0.13.0",`

In that version they used an array in that line.

Updated to 1.1.0 and no problems.

Just for the record this was my error:

```
{
"code": 500,
"message": "Type error: Argument 1 passed to FOS\\RestBundle\\Serializer\\Normalizer\\FormErrorHandler::adaptFormArray() must be an instance of ArrayObject, array given, called in /home/r00t/Workspace/corporapp/vendor/friendsofsymfony/rest-bundle/Serializer/Normalizer/FormErrorHandler.php on line 68"
}

```

Ener-Getick:
No problem, thanks for the report ☺
To prevent this in the future maybe we should have more stuff in the conflicts section of our composer file.

PR #1523, opened 2016-06-27
sllh/php-cs-fixer-styleci-bridge v2.1 upgrade

Soullivaneuh:
I also made your `.php_cs` config file compatible for both PHP-CS-Fixer v1 and v2.

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -14,8 +14,18 @@ For the full copyright and license information, please view the LICENSE
 file that was distributed with this source code.
 EOF;
 
-HeaderCommentFixer::setHeader($header);
+// PHP-CS-Fixer 1.x
+if (method_exists('Symfony\CS\Fixer\Contrib\HeaderCommentFixer', 'getHeader')) {
+    HeaderCommentFixer::setHeader($header);
+}
 
-return ConfigBridge::create()
-    ->setUsingCache(true)
-;
+$config = ConfigBridge::create();
+
+// PHP-CS-Fixer 2.x
+if (method_exists($config, 'setRules')) {
+    $config->setRules(array_merge($config->getRules(), array(
+        'header_comment' => array('header' => $header)
+    )));
+}
+
+return $config;
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,7 @@
         "symfony/css-selector":           "^2.7|^3.0",
         "phpoption/phpoption":            "^1.1",
         "jms/serializer-bundle":          "^1.0",
-        "sllh/php-cs-fixer-styleci-bridge": "^1.3"
+        "sllh/php-cs-fixer-styleci-bridge": "^2.1"
     },
 
     "suggest": {


xabbuh:
I don't think that we need to support both versions as the CS fixer is only required as a dev dependency.

Soullivaneuh:
PHP-CS-Fixer is not required at project level dependencies and user can use both version ATM.

.php_cs
```diff
@@ -1,6 +1,6 @@
 <?php
 
-require_once './vendor/autoload.php';
+require __DIR__.'/vendor/sllh/php-cs-fixer-styleci-bridge/autoload.php';
```
Ener-Getick: Isn't it enough to load composer ?

Soullivaneuh: You might have some common library conflict with the composer autoload.

See: https://github.com/Soullivaneuh/php-cs-fixer-styleci-bridge#troubleshooting

Ener-Getick: @Soullivaneuh your autoloader doesn't fix anything (seldaek proposed a solution in https://github.com/composer/composer/issues/1493 to automatically call the ``__static`` method of every objects loaded which is not what you tried to fix in https://github.com/puli/cli/pull/21).
Whatever is the solution it is not this one, so should be reverted ;)

Soullivaneuh: You mean this solution? https://github.com/composer/composer/issues/1493#issuecomment-12492276

AFAIK, this is **exactly** what I did: https://github.com/Soullivaneuh/php-cs-fixer-styleci-bridge/blob/v2.1.1/autoload.php#L32-L34

And this solve the issue I got on Puli.

stof: @Soullivaneuh except that your bridge does not have anything with a ``__static`` method needing to be called when the class is loaded. This was the specific need asked in this github issue, not a requirement when writing custom autoloaders.
And your "custom" autoloader will not prevent library conflicts, as it will still delegate calls to the composer autoloader for any class, meaning it will still load everything.

Soullivaneuh: If you have better to propose to fix issue like puli/cli#21, I'm hearing because this is the only thing I found so far.

As this library is not concerned yet, I'll revert this part for now.

Soullivaneuh:
@Ener-Getick updated.

Ener-Getick:
LGTM now :+1:

.php_cs
Ener-Getick: @Soullivaneuh maybe the conflict wasn't longer here but the aim of this autoloader is not to fix conflicts.

Soullivaneuh: On puli, the conflict is still here:

```bash
sullivan@sweetnexy:~/projects/fork/puli/cli(styleci-bridge)$ php-cs-fixer fix -v
Fixer "ordered_use" does not exist, skipping.
Fixer "empty_return" does not exist, skipping.
Loaded config from "/home/sullivan/projects/fork/puli/cli/.php_cs".
Using cache file ".php_cs.cache".
PHP Fatal error:  Call to undefined method PhpCsFixer\FixerFileProcessedEvent::setDispatcher() in /home/sullivan/.composer/vendor/symfony/event-dispatcher/EventDispatcher.php on line 42
```

But this happen only with a composer globally installed php-cs-fixer. The `.phar` way works great.

Not sure if this can be easily resolvable.

Ener-Getick:
@Soullivaneuh the solution to your problem would be to create your own autoloader:
```php
// /vendor/.../autoload.php
$loader = new \Composer\Autoload\ClassLoader();
$loader->add('SLLH\StyleCIBridge', __DIR__);
$loader->register();
```

But I'm wondering if we really need your bridge for this bundle as we use the default configuration of ``php-cs-fixer`` anyway...

Soullivaneuh:
> as we use the default configuration of php-cs-fixer anyway...

You are not using the default, you disabled braces.

Also, StyleCI integrate fixer from php-cs-fixer dev-master, so this could be different.

This bridge avoid config duplication and ensure working for both PHP-CS-Fixer v1 and v2 at the moment.

About the autoloader, this might be right, I have to test.

But I have also dependencies to add. I think I'll add it manually.

Ener-Getick:
@Soullivaneuh in fact we disabled ``braces`` to have the same behaviour than ``php-cs-fixer``. WDYT @xabbuh @lsmith77 ?

lsmith77:
fine for me.

Ener-Getick:
I think we should close this. In the current state, it is very likely to create conflicts between libraries and the php-cs-fixer api is good enough to be managed by hand imo.

Ener-Getick:
I'm closing this as we're back to the standard classes.

PR #1542, opened 2016-07-22
Fixed test on windows

### LINKED ISSUE #1492: Allows to define a default format for the body converter
To be consistent with the body listener

see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1219#issuecomment-219803248

AyrtonRicardo:
Fixed some tests that failed on Windows(yeah, I'm using Windows for a while :neutral_face: :cry: ),


Ener-Getick:
Why not using the format listener with a default format ?

EventListener/BodyListener.php
```diff
@@ -86,13 +86,14 @@ public function onKernelRequest(GetResponseEvent $event)
         $contentType = $request->headers->get('Content-Type');
         $normalizeRequest = $this->normalizeForms && $this->isFormRequest($request);
 
-        if ($this->isDecodeable($request)) {
-            $format = null === $contentType
-                ? $request->getRequestFormat()
-                : $request->getFormat($contentType);
+        $format = null === $contentType
```
Ener-Getick: This should stay inside the ``if``

EventListener/BodyListener.php
```diff
@@ -86,13 +86,14 @@ public function onKernelRequest(GetResponseEvent $event)
...
+            ? $request->getRequestFormat($this->defaultFormat)
+            : $request->getFormat($contentType);
+        $format = $format ?: $this->defaultFormat;
```
Ener-Getick: That's not needed as already handle L90.

EventListener/BodyListener.php
```diff
@@ -86,13 +86,14 @@ public function onKernelRequest(GetResponseEvent $event)
...
 
-            $format = $format ?: $this->defaultFormat;
+        $contentType ?: $request->headers->set('Content-Type', $request->getMimeType($format));
```
Ener-Getick: not needed

EventListener/BodyListener.php
```diff
@@ -86,13 +86,14 @@ public function onKernelRequest(GetResponseEvent $event)
         $contentType = $request->headers->get('Content-Type');
         $normalizeRequest = $this->normalizeForms && $this->isFormRequest($request);
 
-        if ($this->isDecodeable($request)) {
-            $format = null === $contentType
-                ? $request->getRequestFormat()
-                : $request->getFormat($contentType);
+        $format = null === $contentType
+            ? $request->getRequestFormat($this->defaultFormat)
+            : $request->getFormat($contentType);
```
Ener-Getick: should be replaced by ``$request->getContentType()``

EventListener/BodyListener.php
```diff
@@ -86,13 +86,13 @@ public function onKernelRequest(GetResponseEvent $event)
...
 
-        if ($this->isDecodeable($request)) {
-            $format = null === $contentType
-                ? $request->getRequestFormat()
-                : $request->getFormat($contentType);
-
-            $format = $format ?: $this->defaultFormat;
+        if (null === $contentType) {
+            $format = $request->getRequestFormat($this->defaultFormat);
+            $request->headers->set('Content-Type', $request->getMimeType($format));
```
Ener-Getick: still not needed

EventListener/BodyListener.php
```diff
@@ -86,13 +86,13 @@ public function onKernelRequest(GetResponseEvent $event)
...
         $normalizeRequest = $this->normalizeForms && $this->isFormRequest($request);
 
-        if ($this->isDecodeable($request)) {
-            $format = null === $contentType
-                ? $request->getRequestFormat()
-                : $request->getFormat($contentType);
-
-            $format = $format ?: $this->defaultFormat;
+        if (null === $contentType) {
+            $format = $request->getRequestFormat($this->defaultFormat);
```
Ener-Getick: Sorry if was unclear, I meant putting it in the ``if ($this->isDecodeable($request)) {``

AyrtonRicardo:
@Ener-Getick in this project I'm using configuration to "type: rest", but when I tryed to use "@ParamConverter" on `getAction` the serializer doesn't works because "$request->getContentType()" wasn't setted.

I didn't found how to use format_listener to set default format on my request

EventListener/BodyListener.php
AyrtonRicardo: I understood, but `$this->isDecodeable` doesn't supports "GET" method, causing error on deserialization.

Ener-Getick: Maybe adding something like
```php
if (!$request->getContent()) {
    return;
}
```
on top of the listener would help ?

Ener-Getick: BTW it looks wrong to use ``getRequestFormat`` (that's not up to you of course).
I think this is dedicated to the response format not the request content.

@xabbuh what do you think about only keeping the ``getContentType`` check ?

Ener-Getick:
@AyrtonRicardo don't worry about my first comment... I posted it too quickly.

I'm just wondering why would a tool not send the ``Content-Type`` header ? unless the request content is empty of course and we should add something to manage this case (if not already done?).
I'll see tomorrow what we can do to improve this listener a bit :)

AyrtonRicardo:
@Ener-Getick No problem xD 
I was using Postman to make requests and the default Request is empty. 

Anyway the tests that failed on Windows were fixed xD

EventListener/BodyListener.php
AyrtonRicardo: if `$request->getContentType()` stay empty the [deserialization](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Request/RequestBodyParamConverter.php#L95) will fail again xD

Ener-Getick: Yes but ``getRequestFormat`` has a confusing name, in fact it contains the response format not the request one.

AyrtonRicardo:
@Ener-Getick sorry about the delay. I agree with you, then the code was reverted.

Ener-Getick:
Thank you @AyrtonRicardo !

Issue #1535, opened 2016-07-08
ParamFetcher can't fetch file arrays


muffe:
I've got a REST function, where I have an image upload as a POST parameter, that looks something like this

``` php
 * @FileParam(name="image-data", strict=false)
 * @View()
 */
 public function postSubmitAction(Request $request, ParamFetcher $paramFetcher)
```

I'm uploading an array of images at the "image-data" key with the [] notation, therefore "image-data[]".
If i try to read the POST data with `$request->files->get('image-data')` I get an array of UploadedFile objects.

If I do the same with `$paramFetcher->get('image-data')` it returns null.

Ener-Getick:
We should probably move ``$map`` from ``AbstractScalarParam`` to ``AbstractParam``.
This would need a new test to ensure this works as expected. Can you work on that @muffe ?

muffe:
I just have taken a look at what the actual problem is.

The `getConstraints` method of the FileParam Annotation looks like this
```
if ($this->image) {
    $constraints[] = new Constraints\Image($options);
} else {
    $constraints[] = new Constraints\File($options);
}
```
In both cases the validation is obviously failing because we get an array as return value.

I just talked with @xabbuh about two possible ways of fixing this.

1. - my way.

We add a new configuration flag to the FileParam Annotation called `multiple`. If this is true we use the Collection constraint.

2. - the @xabbuh way.

In the configuration of the name field of the FileParam we append `[]` to the actual name (so image-data[] instead of image-data[]) to indicate that we want to receive multiple values. While retrieving the value from the actual FileBag we would have the remove the `[]`again.

In both cases we would have to filter the array in some way because we could have potential NULL values.

Ener-Getick:
Fixed in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1474

PR #1474, opened 2016-05-13
Add map option to FileParam to allow upload files in array

jorge07:

PR #1543, opened 2016-07-26
Added ability to upload an array of e.g. images

muffe:
Fixes https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1535

It is possible for the array to have NULL values if the REST function gets POSTed with e.g. empty file uploads in an HTML form.

I'm not really sure if it is reasonable/necessary to filter the array in that case.

### DIFF
--- a/Controller/Annotations/FileParam.php
+++ b/Controller/Annotations/FileParam.php
@@ -31,6 +31,8 @@ class FileParam extends AbstractParam
     public $requirements = null;
     /** @var bool */
     public $image = false;
+    /** @var bool */
+    public $multiple = false;
 
     /**
      * {@inheritdoc}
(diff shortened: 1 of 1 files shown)


Controller/Annotations/FileParam.php
```diff
@@ -43,7 +45,9 @@ public function getConstraints()
         }
 
         $options = is_array($this->requirements) ? $this->requirements : array();
-        if ($this->image) {
+        if($this->multiple) {
+            $constraints = new Constraints\Collection($options);
```
Ener-Getick: maybe ``\All`` would be better ?

muffe: Yeah makes sense, you'll probably want to treat them all the same.

Ener-Getick:
in fact this is a duplicate of https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1474.
I'm closing this one as I prefer https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1474.

Thanks !

PR #1547, opened 2016-08-03
Try to fix #1498

### LINKED ISSUE #1498: Symfony serializer not working on version >= 1.8
Trying version 1.8 break symfony serialization. Objects get serialized to [] .

magnetik:
Not sure if it's the right fix, but it works for me.

Bug #1498

### DIFF
--- a/Serializer/SymfonySerializerAdapter.php
+++ b/Serializer/SymfonySerializerAdapter.php
@@ -59,7 +59,9 @@ class SymfonySerializerAdapter implements Serializer
             $newContext[$key] = $value;
         }
 
-        $newContext['groups'] = $context->getGroups();
+        if (count($context->getGroups())) {
+            $newContext['groups'] = $context->getGroups();
+        }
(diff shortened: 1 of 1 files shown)


Serializer/SymfonySerializerAdapter.php
```diff
@@ -59,7 +59,9 @@ private function convertContext(Context $context)
             $newContext[$key] = $value;
         }
 
-        $newContext['groups'] = $context->getGroups();
+        if (count($context->getGroups())) {
```
xabbuh: I wonder if we should allow to set the groups in the context to `null` to make it possible to distinguish the case where someone builds their group dynamically and actually wants to force the serializer to exclude everything in that case.

magnetik: Is that what dunglas proposed in the bug comment (not sure if I understood well)?

magnetik:
See #1548

PR #1551, opened 2016-08-09
Fix custom message not used on access denied exception

lolmx:
When using `$this->createAccessDeniedException($customMessage)` in a controller, the response will always be:

    {
        code: 403,
        message: "You do not have the necessary permissions"
    }

This is caused by the AccessDeniedListener which is overriding the exception message. Here's a proposal to allow custom messages.

### DIFF
--- a/EventListener/AccessDeniedListener.php
+++ b/EventListener/AccessDeniedListener.php
@@ -70,13 +70,13 @@ class AccessDeniedListener implements EventSubscriberInterface
         $exception = $event->getException();
 
         if ($exception instanceof AccessDeniedException) {
-            $exception = new AccessDeniedHttpException(is_null($exception->getMessage())?'You do not have the necessary permissions':$exception->getMessagee(), $exception);
+            $exception = new AccessDeniedHttpException(is_null($exception->getMessage()) ? 'You do not have the necessary permissions' : $exception->getMessagee(), $exception);
             $event->setException($exception);
         } elseif ($exception instanceof AuthenticationException) {
             if ($this->challenge) {
-                $exception = new UnauthorizedHttpException($this->challenge, is_null($exception->getMessage())?'You are not authenticated':$exception->getMessage(), $exception);
+                $exception = new UnauthorizedHttpException($this->challenge, is_null($exception->getMessage()) ? 'You are not authenticated' : $exception->getMessage(), $exception);
             } else {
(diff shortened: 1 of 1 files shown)


qdequippe:
+1 to have custom messages

lolmx:
The Travis build fails because the tests explicitly expect the message `You are not authenticated` for  [UnauthorizedHttpException](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/EventListener/AccessDeniedListener.php#L77) and [HttpException](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/EventListener/AccessDeniedListener.php#L79).

Should we allow custom messages for those two exceptions too (need updating the tests [here](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Tests/EventListener/AccessDeniedListenerTest.php#L151)) or not?

Ener-Getick:
:+1: but do we really still need this listener ?
Using the exception listener, it's easy to use a different status code depending of the exception.

lolmx:
I don't know if the listener is still needed, but it's actually used.
[Here](https://github.com/lolmx/FOSRestBundle/blob/master/Resources/config/access_denied_listener.xml) it's declared as a service.

And because of the [priority](https://github.com/lolmx/FOSRestBundle/blob/master/EventListener/AccessDeniedListener.php#L90), the AccessDeniedListener will be called before the [ExceptionListener](https://github.com/lolmx/FOSRestBundle/blob/master/EventListener/ExceptionListener.php#L50). Once in the ExceptionListener, my custom message is lost yet, replaced by the `You do not have the necessary permissions`message.

Ener-Getick:
@lolmx you can disable it for now at least:
```yml
fos_rest:
    access_denied_listener: false

lolmx:
Indeed, I got my messages thanks :)

Should I close this PR?

Ener-Getick:
@lolmx i would prefer to have someone else point of view before :)

lsmith77:
to me its also a security question .. exception messages can leak internals. this is why in the exception controller we by default do not include the message in the response.

Ener-Getick:
@lsmith77 ok thanks :) then i think we should add a note in the docs saying that this listener should not be used to keep custom messages.

Thank you @lolmx !

PR #1553, opened 2016-08-19
same class name

tael:
fix this error.
-----
The class "FOS\RestBundle\View\View" is not annotated with @Annotation. Are you sure this class can be used as annotation? If so, then you need to add @Annotation to the _class_ doc comment of "FOS\RestBundle\View\View".

PR #1554, opened 2016-08-19
Patch 1

tael:
also error fixed.
--
The class "FOS\RestBundle\View\View" is not annotated with @Annotation. Are you sure this class can be used as annotation? If so, then you need to add @Annotation to the _class_ doc comment of "FOS\RestBundle\View\View".


tael:
@Ener-Getick can you review this also? 
that's same issue :D

Ener-Getick:
That's just a merge of #1553 in master right ? We're usually doing that ourselves after merging a or ;)

PR #1557, opened 2016-08-23
fix compatibility with jms serializer 1.3.

bendavies:
jms serializser 1.3 changed the default value of `Context::serializeNull`

ref: https://github.com/schmittjoh/serializer/pull/317

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,7 @@
 
     "conflict": {
         "symfony/validator": ">=2.5.0,<2.5.5",
-        "jms/serializer": "<0.13",
+        "jms/serializer": "<0.13 | 1.3.0",
         "jms/serializer-bundle": "<0.11",
         "sensio/framework-extra-bundle": ">=3.0.13"
     },


goetas:
it would be good to have a test for this

Ener-Getick:
I don't think that's the right fix. The problem is that we can't know now if the value has been manually defined or not.

bendavies:
@Ener-Getick not sure i follow. can you expand?

goetas:
> The problem is that we can't know now if the value has been manually defined or not.

does it matter? isnt just `serializeNulls|doNotSerializeNulls` ? does `FOSRestBundle` adds a third case when instead a boolean there is a `null` ?

Ener-Getick:
@goetas the global parameter should normally be used only when it isn't defined locally (eg. explicitelly in the controller).
But yes the safest option we have here imo is to use `||`.

stof:
@goetas the current behavior of FOSRestBundle is to preserve the existing value of the SerializationContext if you passed your own context with a configured value for this setting, and to configure it itself otherwise

goetas:
@stof  i see, ok, now the existing value is always a boolean, and by default is false.

bendavies:
I see what you mean. so how can we support both 1.2 and 1.3 of jms serializer?

goetas:
I'm not the best expert in how to deal with BC breaks, but does it make sense revert https://github.com/schmittjoh/serializer/pull/317/ and create a 1.3.1 release?

Ener-Getick:
That's a solution but it depends of @schmittjoh point of view.

goetas:
Created PR https://github.com/schmittjoh/serializer/pull/630

goetas:
https://github.com/schmittjoh/serializer/releases/tag/1.3.1 BC break reverted.

to be safe probably on composer can be added a conflict with the version 1.3.0

Ener-Getick:
@goetas thanks ! Can you update your pr @bendavies ? :)

bendavies:
Will do.

bendavies:
@Ener-Getick how's that

bendavies:
well according to https://getcomposer.org/doc/articles/versions.md
`"jms/serializer":                 "~0.13|~1.0, !=1.3.0",`
should work, but it doesn't.
(confirmed it doesn't work by trying to exclude 1.3.1, but 1.3.1 is still installed)

Any ideas?

Ener-Getick:
@bendavies did you remove your vendor directory ? Sometimes it fixes cache issues.

bendavies:
i did, removed vendor and composer.lock and reinstalled. can you try it?

Ener-Getick:
Only tomorrow, i'm on phone right now :/

bendavies:
@Seldaek sorry for the ping, any ideas?

bendavies:
seems like a bug in the solver to me
works:
`"jms/serializer":                 "~1.0 <>1.3.1",`
does not work:
`"jms/serializer":                 "~0.13|~1.0 <>1.3.1",`

bendavies:
fixed by not using the tilde operator and being verbose in the constraint.

bendavies:
still not fixed, can't get the conflict to work, do not merge.

bendavies:
https://github.com/composer/composer/issues/5616
https://github.com/composer/composer/issues/5617

bendavies:
@Ener-Getick think this is good for merge now.

Ener-Getick:
Shouldn't we use `|` for consistency?

bendavies:
sure

Ener-Getick:
Thank you @bendavies ! :)

goetas:
This should go in 2.0 branch I guess

Ener-Getick:
Yes, i'll merge it soon ;)

bendavies:
i'm not sure 2.0 suffers from the bug...need to check

PR #1559, opened 2016-08-30
Add condition to YAML routing loader, Add routing loader test

IndraGunawan:
Fix "Warning:  Invalid argument supplied for foreach()" if routes file is empty.
Add Routes loader test. all green for RestYamlCollectionLoaderTest


Tests/Routing/Loader/RestXmlCollectionLoaderTest.php
```diff
@@ -25,6 +25,25 @@
 class RestXmlCollectionLoaderTest extends LoaderTest
 {
     /**
+     * Test that XML files are invalid.
+     *
+     * @expectedException \InvalidArgumentException
+     * @dataProvider getPathsToInvalidFiles
+     */
+    public function testLoadThrowsExceptionWithInvalidFile($filePath)
```
Ener-Getick: Can you add an `@expectedExceptionMessage` please ?

Tests/Routing/Loader/RestYamlCollectionLoaderTest.php
```diff
@@ -27,6 +27,36 @@
...
+        $this->assertEquals([], $collection->all());
+    }
+
+    /**
+     * Test that YAML files are invalid.
+     *
+     * @expectedException \InvalidArgumentException
+     * @dataProvider getPathsToInvalidFiles
+     */
+    public function testLoadThrowsExceptionWithInvalidFile($filePath)
```
Ener-Getick: Same here

Tests/Routing/Loader/RestYamlCollectionLoaderTest.php
```diff
@@ -184,6 +214,36 @@ public function testControllerAsServiceWithClassName()
...
+            $this->assertEquals($params['path'], $route->getPath(), $name);
+            $this->assertEquals($params['method'], $methods[0], $name);
+            $this->assertContains($params['controller'], $route->getDefault('_controller'), $name);
+        }
+
+        $name = 'api_get_billing_payments';
+        $this->assertArrayNotHasKey($name, $etalonRoutes);
+    }
+
+    public function testRoutesWithPattern()
```
Ener-Getick: Is this related ?

Tests/Routing/Loader/RestYamlCollectionLoaderTest.php
indragunawan: i can't add `@expectedExceptionMessage` because of i'm using `@dataProvider` so the exception message may be difference for every tests file.
for example:
when exception catcby ParseException, the exception message is 'The file "/home/your/path/bad_format.yml" does not contain valid YAML' [RestYamlCollectionLoader.php](https://github.com/indragunawan/FOSRestBundle/blob/f2a557ac36284516c319e7925546907a6ab6eeee/Routing/Loader/RestYamlCollectionLoader.php#L67)
when the value of the routes file is not array, the exception message is 'The file "/home/your/path/nonvalid2.yml" must contain a YAML array.' [RestYamlCollectionLoader.php](https://github.com/indragunawan/FOSRestBundle/blob/f2a557ac36284516c319e7925546907a6ab6eeee/Routing/Loader/RestYamlCollectionLoader.php#L80)

i need to split the test to use `@expectedExceptionMessage`, how about this?

Tests/Routing/Loader/RestYamlCollectionLoaderTest.php
indragunawan: this test for [this](https://github.com/indragunawan/FOSRestBundle/blob/f2a557ac36284516c319e7925546907a6ab6eeee/Routing/Loader/RestYamlCollectionLoader.php#L126-L130)

Tests/Routing/Loader/RestYamlCollectionLoaderTest.php
Ener-Getick: Yes, should be better :)

Tests/Routing/Loader/RestYamlCollectionLoaderTest.php
Ener-Getick: It looks like this one should be deprecated (as it was deprecated in symfony) but ok for this test.

Tests/Routing/Loader/RestYamlCollectionLoaderTest.php
IndraGunawan: Ok. i will make seperate test for this. :D

Tests/Routing/Loader/RestYamlCollectionLoaderTest.php
IndraGunawan: how about if i add deprecated trigger_error to this section?

Ener-Getick: Should be done in another PR (targetting master), but fine for me :)

Routing/Loader/RestYamlCollectionLoader.php
```diff
@@ -60,11 +61,25 @@ public function load($file, $type = null)
...
         $collection->addResource(new FileResource($path));
 
+        // empty file
+        if (null === $config) {
+            return $collection;
+        }
+
+        // not an array
+        if (!is_array($config)) {
+            throw new \InvalidArgumentException(sprintf('The file "%s" must contain a YAML array.', $path));
```
Ener-Getick: it is called `mapping` in Yaml. maybe we can replace that by `a mapping (an array)`?

Tests/Routing/Loader/RestXmlCollectionLoaderTest.php
```diff
@@ -25,6 +25,28 @@
 class RestXmlCollectionLoaderTest extends LoaderTest
 {
     /**
+     * Test that route route not found.
+     *
+     * @expectedException \InvalidArgumentException
+     * @expectedExceptionMessageRegExp /Cannot find parent resource with name*./
```
Ener-Getick: No need for a regexp, it is already checked using `assertContains`

Ener-Getick:
Thank you @IndraGunawan :)

PR #1560, opened 2016-09-01
Fix available methods of View

IndraGunawan:


Resources/doc/2-the-view-layer.rst
```diff
@@ -143,14 +143,15 @@ There are several more methods on the ``View`` class, here is a list of all
...
 * ``setData($data)`` - Set the object graph or list of objects to serialize.
-* ``setTemplateData($templateData)`` - Set the template data array or anonymous function. Closure should return array.
+* ``setTemplateData($data)`` - Set the template data array or anonymous function. Closure should return array.
 * ``setHeader($name, $value)`` - Set a header to put on the HTTP response.
 * ``setHeaders(array $headers)`` - Set multiple headers to put on the HTTP response.
-* ``setSerializationContext($context)`` - Set the serialization context to use.
-* ``setTemplate($name)`` - Name of the template to use in case of HTML rendering.
-* ``setTemplateVar($name)`` - Name of the variable the data is in, when passed
+* ``setStatusCode($code)`` - Set the HTTP status code.
+* ``setContext(Context $context)`` - Set the serialization context to use.
```
Ener-Getick: This method should not be used imo (maybe we will deprecate it), using `getContext` is preferred.

IndraGunawan: i just add available method of FOS\RestBundle\View\View.

its better to remove from doc?

Ener-Getick: The section title says that those methods are the most important so I think we should only include the most important :)

IndraGunawan: it would be * ``getContext()`` - The serialization context to use.
or you have another opinion?

Ener-Getick: Fine for me :+1:

Ener-Getick:
Thank you @IndraGunawan ! :)

PR #1562, opened 2016-09-01
Handle invalid param exception when invalid value is not scalar

### LINKED ISSUE #1546: Query param array validation: array to string conversion error
Hello,
I followed the conversation about array to map changes and some logic to validate types for array keys.
What I want to achieve is really similar to what is in the docs:
```
     * @QueryParam(map=true, name="filters", requirements=@MyComplexConstraint, description="List of complex filters")
     * In this example, the ParamFetcher will validate each value of the array with the constraint, returning the
     * default value if you are in safe mode or throw a BadRequestHttpResponse containing the constraint violation
     * messages in the message.
```
http://symfony.com/doc/master/bundles/FOSRestBundle/param_fetcher_listener.html
But without the `map=true` to get parameters in validator once, as an array with keys

My setup is like this:
```
     * @QueryParam(
     *     name="page",
     *     strict=true,
     *     nullable=true,
     *     requirements=@MyCustomValidatorConstraint
     *     )
```
Following comment:
https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1192#issuecomment-156176789
I get an array with keys in custom Validator.
Works fine until I add any violation:
  `$this->context->addViolation('Error message');`
or following Symfony docs way:
```
$this->context->buildViolation('error message')
                ->setParameter('%string%', $value)
                ->addViolation();
``` 
Adding violations in that custom validator ends up with:
```
[2016-07-29 17:18:53] request.CRITICAL: Uncaught PHP Exception Symfony\Component\Debug\Exception\ContextErrorException: "Notice: Array to string conversion" at /{...}/vendor/friendsofsymfony/rest-bundle/Exception/InvalidParameterException.php line 47 {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\ContextErrorException(code: 0): Notice: Array to string conversion at /{...}/vendor/friendsofsymfony/rest-bundle/Exception/InvalidParameterException.php:47)"} []
```
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Exception/InvalidParameterException.php#L45
`$violation->getInvalidValue()` return array of filters, not a string.

Do I miss anything and should configure something to make this work?
Looks I need to set invalid value just don't see a way to do it.

PS.
fos-rest bundle 2.0.0

For now I use not great hacks to make it working.
I use map=true and, based on property path in context inside custom validator, try to find which key I am checking now to know which validation rule to apply.

senkal:
Fixes #1546

### DIFF
--- a/Exception/InvalidParameterException.php
+++ b/Exception/InvalidParameterException.php
@@ -38,11 +38,13 @@ class InvalidParameterException extends BadRequestHttpException
             if ($key > 0) {
                 $message .= "\n";
             }
+            
+            $invalidValue = $violation->getInvalidValue();
 
             $message .= sprintf(
                 'Parameter "%s" of value "%s" violated a constraint "%s"',
                 $parameter->getName(),
-                $violation->getInvalidValue(),
+                is_scalar($invalidValue) ? $invalidValue : var_export($invalidValue, true),
                 $violation->getMessage()
             );
         }
--- a/Exception/InvalidParameterException.php
+++ b/Exception/InvalidParameterException.php
@@ -38,7 +38,7 @@ class InvalidParameterException extends BadRequestHttpException
             if ($key > 0) {
                 $message .= "\n";
             }
-            
+
             $invalidValue = $violation->getInvalidValue();
 
             $message .= sprintf(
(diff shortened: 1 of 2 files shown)


Ener-Getick:
Could you fix cs please?

senkal:
@Ener-Getick I am afraid I need some help with it.
I am not sure what to run.
I downloaded php-cs-fixer, since you have the configuration for it in your project, and run fix.
It used your configuration file(`.php_cs`) and fix the file when there is an issue.
But still it is not enough.

Which exactly tool + configuration should I run to be on pair with your settings?

Ener-Getick:
Great thanks @senkal ! :)

Issue #1082, opened 2015-07-08
RestBundle routes doesnt support hostname matching


brpaz:
In standard Symfony routing, you can say that a specific route will only match if the hostname matches a defined parameter. See http://symfony.com/doc/current/components/routing/hostname_pattern.html

FosRestBunlde seems to npt support this behaviour.

In RestYamlCollectionLoader "host" parameter is not read from configs. I think its a simple fix. I can try to make a PR

lsmith77:
sure .. I think this feature was added later on .. please make sure to also update the tests (and docs if necessary)

pedroresende:
@lsmith77 This is a possible PR for this

xabbuh:
This was fixed by #1564.

PR #1564, opened 2016-09-03
- Added feature hostname matching - RestBundle routes doesnt support hostname matching #1082

pedroresende:


lsmith77:
looks good to me from a quick inspection .. @Ener-Getick wdyt?

Routing/Loader/RestXmlCollectionLoader.php
```diff
@@ -70,6 +70,7 @@ protected function parseNode(RouteCollection $collection, \DOMElement $node, $pa
                 $namePrefix = (string) $node->getAttribute('name-prefix');
                 $parent = (string) $node->getAttribute('parent');
                 $type = (string) $node->getAttribute('type');
+                $host = isset($config['host']) ? $config['host'] : null;
```
stloyd: Assign this to local variable is useless, as this is used only once.

pedroresende: @stloyd I've assigned to a local variable use the same structure that is currently present

Resources/doc/5-automatic-route-generation_single-restful-controller.rst
```diff
@@ -31,6 +31,7 @@ are explained in the next section.
     # app/config/routing.yml
     users:
         type:     rest
+        host:     m.example.com
```
Ener-Getick: I don't think we should include this in of our first examples, it would only confuse people as it's something very common.

Resources/doc/6-automatic-route-generation_multiple-restful-controllers.rst
```diff
@@ -15,10 +15,12 @@ XML collection:
     # src/Acme/HelloBundle/Resources/config/users_routes.yml
     users:
         type:     rest
+        host:     hostname.example.com
```
Ener-Getick: this one is managed by symfony anyway so it's useless imo.

Routing/Loader/RestXmlCollectionLoader.php
Ener-Getick: I agree with @stloyd I don't think we need this variable. Maybe we should remove the other variables as well.

Ener-Getick:
BTW is it really useful to keep supporting XML and Yaml? This is more or less a copy-paste of the one of symfony except some features but there are not worth it imo (automatically add format prefix, naming prefix (this one can be useful but it doesn't belong to this bundle)).
WDYT?

Ener-Getick:
Thank you @pedroresende !

PR #1566, opened 2016-09-06
Skipping php files without classes for the route generation

parnas:
In my case, I use Controller Traits, and that leads to "Symfony\Component\Config\Exception\FileLoaderLoadException: Cannot load resource "false"". Just checking if there is a class in file, before trying to load it, solves the problem.


Routing/Loader/DirectoryRouteLoader.php
```diff
@@ -50,8 +50,10 @@ public function load($resource, $type = null)
         $finder = new Finder();
 
         foreach ($finder->in($resource)->name('*.php')->files() as $file) {
-            $imported = $this->processor->importResource($this, ClassUtils::findClassInFile($file), array(), null, null, 'rest');
-            $collection->addCollection($imported);
+            if($class = ClassUtils::findClassInFile($file)) {
```
Ener-Getick: Maybe we should use a stricter comparison in case there class name is similar to false `false !==`

parnas: I'm really wondering in what cases that might happen, but sure, it would be, technically, more correct :)

What do you prefer, assign in a separate line then check or "&& $class !== false"?

Ener-Getick: I was thinking about a name such as `0` but i realized that's not allowed so it's fine for me as is :)

Ener-Getick:
Thank you @parnas !

PR #1570, opened 2016-09-10
#1569: Added suffix "Action" for getUsers() method as it is required …

### LINKED ISSUE #1569: It is not possible to define route manually without "Action" suffix in controller method name
Actually, it is normal behaviour for `RestActionReader`, but [**documentation**](http://symfony.com/doc/current/bundles/FOSRestBundle/7-manual-route-definition.html#method-name-prefix) in this case confuses developers to use manual definition of routes. Example demonstrates following snippet: `public function getUsers() { /** */ }`, but without `Action` suffix it will not work according to this condition [Routing/Loader/Reader/RestActionReader.php#L430](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Routing/Loader/Reader/RestActionReader.php#L430)

How to update documentation for this component?

warlock39:
…precondition to define route manually

Fixes #1569

### DIFF
--- a/Resources/doc/7-manual-route-definition.rst
+++ b/Resources/doc/7-manual-route-definition.rst
@@ -82,7 +82,7 @@ Example:
     * @Get("/users/foo", name="_foo")
     * @Get("/users/bar", name="_bar")
     */
-    public function getUsers() { /** */ }
+    public function getUsersAction() { /** */ }
 
 
 Result:
@@ -107,7 +107,7 @@ Example:
     * @Get("/users/foo", name="get_foo", options={ "method_prefix" = false })
     * @Get("/users/bar", name="get_bar", options={ "method_prefix" = false })
     */
-    public function getUsers() { /** */ }
+    public function getUsersAction() { /** */ }
 
 
 Result:


Ener-Getick:
Why this change? The Action prefix is not required for manual routing it is only for automatic routing.

warlock39:
@Ener-Getick , thank for your feedback.

According to description of referenced issue: #1569 , the condition in [Routing/Loader/Reader/RestActionReader.php#L430](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Routing/Loader/Reader/RestActionReader.php#L430) filters any class methods without `Action` suffix. So there are two options to get manual routing work:

1. Register additional routing resource (example: `resource: '@AppBundle/Controller/'` with `type: annotation`) in `routing.yml`
2. Add `Action` suffix to method name with custom route annotation

[Documentation](http://symfony.com/doc/master/bundles/FOSRestBundle/7-manual-route-definition.html) says nothing about it.
If i overlooked something, please let me know.

Ener-Getick:
Ok got it.
I have to look deeper into the implementation but imo we should instead remove this restriction when using custom routes.

lsmith77:
I am not 100% sure but I think this restriction was for security reasons.

Ener-Getick:
@lsmith77 I understand for automatic routes (eg we don't want all methods to be converted to routes)
```
// should not be converted to a route
public function getUser(): UserInterface
{
}
```

but if the user explicitly requires a route what security problems could it cause?
```
/**
 * @Get('/user')
 */
public function getUser(): UserInterface
{
}

lsmith77:
that should be fine indeed.

Ener-Getick:
Let's merge this for now :)

Issue #1573, opened 2016-09-13
BodyListener should check for supported content-type before checking if request is decodable


JornEngelbart:
The body listener first checks whether a request is decodable and then checks whether or not the given format derived from the content-type is supported. Because the decodable check returns ```false``` whenever a request is a form request, form requests are by definition always supported. This is inconvenient behaviour because for a lot of REST api's it is required to disallow these request to prevent CSRF attacks using web forms.

An easy solution would be to add a very simple form decoder and move the isFormRequest to check if a form decoder is supported in the decoderProvider.

JornEngelbart:
Okay, after some research I think it would be better to have a configuration option to accept or not accept form content types. Will create pull request for my current solution

PR #1574, opened 2016-09-13
#1573 Add config option to disallow form content types in body listener

JornEngelbart:


EventListener/BodyListener.php
```diff
@@ -41,17 +41,20 @@ class BodyListener
...
      * @param DecoderProviderInterface $decoderProvider
      * @param bool                     $throwExceptionOnUnsupportedContentType
+     * @param bool                     $acceptFormContentType
      * @param ArrayNormalizerInterface $arrayNormalizer
      * @param bool                     $normalizeForms
      */
     public function __construct(
         DecoderProviderInterface $decoderProvider,
         $throwExceptionOnUnsupportedContentType = false,
+        $acceptFormContentType = true,
```
Ener-Getick: That's a bc break

JornEngelbart:
I also fixed this for the 1.7 branch. Should I create another pull request for this?

Ener-Getick:
Why don't you want to support forms?

JornEngelbart:
@Ener-Getick Because there is a csrf vulnerability when doing a POST request to an endpoint that does not require any request body. When a user is already logged in a malicious form could cause the user to perform actions that he did not intent to do.

Ener-Getick:
@JornEngelbart i'm not sure this belongs to this bundle as this part is managed by php and symfony themselves.
IMO you should create your own listener throwing an exception when having a form.

Wdyt @lsmith77 @xabbuh?

JornEngelbart:
@Ener-Getick Seems inconsistent to me. There is an option to throw exceptions when a content-type is not supported. But at the same time two "special" content-types, namely the forms, are always supported.

Ener-Getick:
Because forms are always supported (they are managed by php itself). 
What's the problem of having your own listener ? It's just a matter of a few lines.

JornEngelbart:
@Ener-Getick Implementing an own listener is no problem. But IMO if you give an option to throw an exception on unsupported content-types it would be rather logical to be able to support or not support any content-type including forms (whether or not they are managed by PHP, these are content-types). Right?

Ener-Getick:
I have no strong opinion about it, i'm just afraid it could bring more maintainance without bringing a real value imo.
I remembered [NelmioCorsBundle](https://github.com/nelmio/NelmioCorsBundle), why not using it ? It should be safer for csrf vulnérabilités.

xabbuh:
Wouldn't it be better to then let the decoder provider decide whether or not support this content type and let the already existing exception mechanism step into?

xabbuh:

PR #1576, opened 2016-09-18
Update the annotation doc to change array to map.

ChadSikorra:
I noticed this while doing annotations based off the docs. It looks like the `array` property for the annotation `QueryParam` and `RequestParam` was changed to `map` a while ago.


Ener-Getick:
Thank you @ChadSikorra !

Issue #1578, opened 2016-09-21
escape malformed UTF-8 characters


franek:
On one of our project, we are using FOSRestBundle.
We have defined a query param : 
```
* @QueryParam(name="terms", requirements={"rule" = ".*", "error_message" = "Terms must be a valid string."}, strict=true, nullable=true, description="...")
```

Behind the scene, our API returns to user error message. It is serialized by Symfony serializer.

Some of our users try our endpoint and send us malformed UTF-8 characters. In this edge case, Serializer throws an exception : 
```
Symfony\Component\Serializer\Exception\UnexpectedValueException: Malformed UTF-8 characters, possibly incorrectly encoded
https://github.com/FriendsOfSymfony/FOSRestBundle/commit/b4a08cd6c84535bfe9e18eea52e78a9f5b2ad29d

It throws an exception because InvalidParameterException displays the invalid character whithout escaping it.

I have added a test to demonstrate the error https://github.com/FriendsOfSymfony/FOSRestBundle/commit/14371a979c727fc2ac74523655263244b7d30b43.

I don't really know how to fix that. I have tried to encode the character with utf8_encode. It seems to fix my problem but I think it is not a good solution. It may have some FOSRest users that does not use UTF-8 : https://github.com/FriendsOfSymfony/FOSRestBundle/commit/b4a08cd6c84535bfe9e18eea52e78a9f5b2ad29d

PR #1579, opened 2016-09-21
Escape malformed utf 8 characters

franek:
related to https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1578

### DIFF
--- a/Exception/InvalidParameterException.php
+++ b/Exception/InvalidParameterException.php
@@ -44,7 +44,7 @@ class InvalidParameterException extends BadRequestHttpException
             $message .= sprintf(
                 'Parameter "%s" of value "%s" violated a constraint "%s"',
                 $parameter->getName(),
-                is_scalar($invalidValue) ? $invalidValue : var_export($invalidValue, true),
+                is_scalar($invalidValue) ? utf8_encode($invalidValue) : var_export($invalidValue, true),
                 $violation->getMessage()
             );
         }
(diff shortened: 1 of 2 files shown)


Exception/InvalidParameterException.php
```diff
@@ -44,7 +44,7 @@ public static function withViolations(ParamInterface $parameter, ConstraintViola
             $message .= sprintf(
                 'Parameter "%s" of value "%s" violated a constraint "%s"',
                 $parameter->getName(),
-                is_scalar($invalidValue) ? $invalidValue : var_export($invalidValue, true),
+                is_scalar($invalidValue) ? utf8_encode($invalidValue) : var_export($invalidValue, true),
```
Tobion: `utf8_encode` requires the xml writer extension. it should be available most of the time so I guess this is fine.
But this will just change the original value so the error message becomes useless because the allegedly wrong value is not the real wrong value anymore. What I would propose, is to check if the string is valid utf8 and if not, hex encode the binary data and output the hex values of the wrong value.

Tests/Request/ParamFetcherTest.php
```diff
@@ -262,6 +264,54 @@ public function testValidationException()
...
+        list($fetcher, $method) = $this->getFetcherToCheckValidation($param);
+
+        $stringInvalidValue = urldecode('%E3%81');
+        $stringViolation = $this->getMockBuilder(ConstraintViolationInterface::class)
+            ->getMock();
+        $stringViolation->method('getInvalidValue')
+            ->willReturn($stringInvalidValue);
+
+        $arrayInvalidValue = ['page' => 'abcde'];
+        $arrayViolation = $this->getMockBuilder(ConstraintViolationInterface::class)
```
Tobion: please rebase the PR and use `$this->createMock`

xabbuh:
I am closing here as the PR seems to be stalled.

Issue #1536, opened 2016-07-08
Not possible to use a custom exclusion strategy with JMSSerializer


TobiasXy:
In my project I wrote a custom exclusion strategy for the JMSSerializer.
In version 1.7 I added it like this:

```
$context = SerializationContext::create()
            ->addExclusionStrategy($this->get('my_exclusion_strategy'));
$view->setSerializationContext($context);
```

Now in version 2.0 there is the new (generic) Context class, but it seems there is no way to set a custom exclusion strategy. Also there is apparently no way to hook into the JMSSerializerAdapter, (which is creating the actual SerializationContext of the JMSSerializer) to add the exclusion strategy there.

Do I really have to override the JMSSerializerAdapter to add a custom exclusion strategy or did I miss something obvious?

Ener-Getick:
What about using a new attribute ``exclusion_strategies`` ?

cc @xabbuh

TobiasXy:
Exclusion strategies are specific to the JMSSerializer, right?
Adding a property `exclusion_strategies` to the generic Context class would make it somehow specific again or did I misunderstand you there?

Ener-Getick:
It would only be used by the jms serializer but i don't see another solution.
Except maybe making the adapters extandable.

xabbuh:
Sounds like a good idea. A comment may be needed to explain that this might not be supported by adapters (or we could make this comment even more generic so that it applies to the `Context` class as a whole).

alanhartless:
In the same boat here - ended up just downgrading to 1.8 after realizing this would be an issue as I'm also using a JMS custom exclusion strategy.

Ener-Getick:
Can someone submit a PR for this one ? I already have a lot of PRs to manage :/

podorozhny:
+1

bblue:
Is anyone working on a PR for this? I'll give it a shot if not...

Ener-Getick:
I don't think so. I think updating the jms adapter should be enough.

xabbuh:
This was implemented in #1583.

PR #1583, opened 2016-09-30
#1536 Added support for JMS exclusion strategies

bblue:
Fix for issue #1536. I don't think it needs more magic than this.

### DIFF
--- a/Context/Context.php
+++ b/Context/Context.php
@@ -263,20 +263,6 @@ final class Context
     }
 
     /**
-     * Sets the exclusion strategies.
-     *
-     * Warning: This only applies to the JMS serializer adapter.
-     *
-     * @see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1536
-     *
-     * @param array $exclusionStrategies
-     */
-    public function setExclusionStrategies($exclusionStrategies)
-    {
-        $this->exclusionStrategies = $exclusionStrategies;
-    }
-
-    /**
      * Gets an array of exclusion strategies.
      *
      * Warning: This only applies to the JMS serializer adapter.
--- a/Context/Context.php
+++ b/Context/Context.php
@@ -42,6 +42,10 @@ final class Context
      * @var bool
      */
     private $serializeNull;
+    /**
+     * @var array
+     */
+    private $exclusionStrategies = array();
 
     /**
      * Sets an attribute.
@@ -256,5 +260,44 @@ final class Context
     public function getSerializeNull()
     {
         return $this->serializeNull;
+    }
+
+    /**
+     * Sets the exclusion strategies
+     *
+     * Warning: This only applies to the JMS serializer adapter.
+     * @see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1536
+     *
+     * @param array $exclusionStrategies
+     */
+    public function setExclusionStrategies($exclusionStrategies)
+    {
+        $this->exclusionStrategies = $exclusionStrategies;
+    }
+
+    /**
+     * Gets an array of exclusion strategies
+     *
+     * Warning: This only applies to the JMS serializer adapter.
+     * @see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1536
+     *
+     * @return array
+     */
+    public function getExclusionStrategies()
+    {
+        return $this->exclusionStrategies;
+    }
+
+    /**
+     * Adds an exclusion strategy
+     *
+     * Warning: This only applies to the JMS serializer adapter.
+     * @see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1536
+     *
+     * @param $exclusionStrategy
+     */
+    public function addExclusionStrategy($exclusionStrategy)
+    {
+        $this->exclusionStrategies[] = $exclusionStrategy;
     }
 }
--- a/Context/Context.php
+++ b/Context/Context.php
@@ -11,7 +11,6 @@
 
 namespace FOS\RestBundle\Context;
 
-
 use JMS\Serializer\Exclusion\ExclusionStrategyInterface;
 
 /**
--- a/Context/Context.php
+++ b/Context/Context.php
@@ -267,8 +267,6 @@ final class Context
      *
      * Warning: This only applies to the JMS serializer adapter.
      *
-     * @see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1536
-     *
      * @return array
      */
     public function getExclusionStrategies()
@@ -280,8 +278,6 @@ final class Context
      * Adds an exclusion strategy.
      *
      * Warning: This only applies to the JMS serializer adapter.
-     *
-     * @see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1536
      *
      * @param $exclusionStrategy
      */
--- a/Context/Context.php
+++ b/Context/Context.php
@@ -266,9 +266,9 @@ final class Context
     }
 
     /**
-     * Gets an array of exclusion strategies.
+     * Gets the array of exclusion strategies.
      *
-     * Warning: This only applies to the JMS serializer adapter.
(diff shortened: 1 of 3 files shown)


lsmith77:
fine for me

Serializer/JMSSerializerAdapter.php
```diff
@@ -96,6 +96,11 @@ private function convertContext(Context $context, $direction)
         if (null !== $context->getSerializeNull()) {
             $jmsContext->setSerializeNull($context->getSerializeNull());
         }
+        if (is_array($strategies = $context->getExclusionStrategies())) {
```
Ener-Getick: why not just always make it an array ?

yagobski:
Please release this commit soon ! Thanks

Serializer/JMSSerializerAdapter.php
bblue: Sure can. I will update the PR when I get home today. The `is_array` should move to the `setExclusionStrategies()` method instead then. I suppose the same method should throw an exception if the parameter is not an array, so it does not fail silently (as the current approach also would).

bblue: On second thought `is_array ` and an exception is not needed, any invalid data will clearly make itself known when the foreach fails. Pushed a simplied version just now.

bblue:
I am not really sure why the last commit failed the Travis test...

Context/Context.php
```diff
@@ -257,4 +261,46 @@ public function getSerializeNull()
...
+    /**
+     * Sets the exclusion strategies.
+     *
+     * Warning: This only applies to the JMS serializer adapter.
+     *
+     * @see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1536
+     *
+     * @param array $exclusionStrategies
+     */
+    public function setExclusionStrategies($exclusionStrategies)
```
Ener-Getick: Do we really need a setter? Isn't an adder enough?

Context/Context.php
```diff
@@ -257,4 +261,32 @@ public function getSerializeNull()
     {
         return $this->serializeNull;
     }
+
+    /**
+     * Gets an array of exclusion strategies.
+     *
+     * Warning: This only applies to the JMS serializer adapter.
+     *
+     * @see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1536
```
Ener-Getick: I don't think we need to link this issue.

Context/Context.php
```diff
@@ -257,4 +261,32 @@ public function getSerializeNull()
...
+    /**
+     * Adds an exclusion strategy.
+     *
+     * Warning: This only applies to the JMS serializer adapter.
+     *
+     * @see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1536
+     *
+     * @param $exclusionStrategy
+     */
+    public function addExclusionStrategy($exclusionStrategy)
```
Ener-Getick: can't we type hint on an interface?

Ener-Getick:
LGTM :+1: btw could you add some tests to avoid regressions?

Tests/Context/ContextTest.php
```diff
@@ -98,4 +98,20 @@ public function testSerializeNull()
...
+    {
+        $strategy1 = $this->createMock('JMS\Serializer\Exclusion\ExclusionStrategyInterface');
+        $strategy2 = $this->createMock('JMS\Serializer\Exclusion\ExclusionStrategyInterface');
+
+        $this->context->addExclusionStrategy($strategy1);
+        $this->context->addExclusionStrategy($strategy2);
+
+        $strategies = $this->context->getExclusionStrategies();
+
+        $this->assertEquals(true, is_array($strategies));
```
Ener-Getick: This looks useless

Tests/Context/ContextTest.php
```diff
@@ -98,4 +98,20 @@ public function testSerializeNull()
...
+        $this->assertEquals(true, !empty($strategies));
```
Ener-Getick: The next assertion is enough

Tests/Context/ContextTest.php
```diff
@@ -98,4 +98,20 @@ public function testSerializeNull()
...
+        $this->assertEquals(true, in_array($strategy1, $strategies, true));
+        $this->assertEquals(true, in_array($strategy2, $strategies, true));
```
Ener-Getick: Can be replaced by assertEquals([$s1, $s2], $strategies);

bblue:
Done. That was close to my first ever test, so don't ask me to create another unless you are prepared to tear it apart :)

Tests/Context/ContextTest.php
```diff
@@ -98,4 +98,17 @@ public function testSerializeNull()
 
         $this->assertEquals(true, $this->context->getSerializeNull());
     }
+
+    public function testExclusionStrategy()
+    {
+        $strategy1 = $this->getMock('JMS\Serializer\Exclusion\ExclusionStrategyInterface');
+        $strategy2 = $this->getMock('JMS\Serializer\Exclusion\ExclusionStrategyInterface');
```
Ener-Getick: This is deprecated :/ you can replace it by `getMockBuilder(...)->getMock()` and also use the `::class` constant.

Ener-Getick:
@bblue no problem, my review is here for that :)
Thanks for taking care of this issue !

Ener-Getick:
Thank you @bblue !

xabbuh:
The `Context` class is part of our abstraction layer. Cluttering it with implementation-specific (JMS-specific here) code goes against this abstraction. I think a proper solution would be to ship our own `ExclusionStrategy` interface and let the concrete adapters deal with the implementation-specific stuff.

This way we could also easily update the Symfony Serializer integration if at some time in the future the Symfony Serializer component ships with a similar feature.

Ener-Getick:
@xabbuh i'm affraid it would be confusing and harder to use for no gain.
What's the problem of depending on an external interface as long as we can still use the class in other contexts? At worst if we really don't want a jms interface what about using an attribute?

xabbuh:
Imagine the Symfony Serializer got exclusion strategies in the future too? Due to the JMS type hint we would have to add new methods for the Symfony implementation serving the same purpose but with a different signature. Or what do we do when we want to support more implementations (for example, [ivory/serializer](https://github.com/egeloen/ivory-serializer/)? This would lead in the same direction. If we went this way, the `Context` class could become a huge mess.

Ener-Getick:
@xabbuh we can update the class signature (it is final) so we would only have to replace the type hint by a less stricter one.
BTW what do you think about recentrating the Context class around the attributes concept? It looks weird to me that `setAttribute('groups')` does not the same as `setGroups()`.

xabbuh:
But what would the less stricter type hint? There is no common base interface that we could use. Which means that we would then be forced to remove the type hint at all.

What exactly do you mean with the attributes concept? I mean what do you envision?

Ener-Getick:
If symfony adds exclusion strategies support, we could add its interface to our type hinting, but i don't think it's worth having a common class.

For the attributes, i just would like something like:
```php
public function getGroups() {
    return $this->getAttribute('groups');
}
```
That would be a small bc break, but the class behavior would be more expectable.

xabbuh:
We couldn't add the interface to the type hint (except if there was a common base interface for both implementations which doesn't seem very likely). So effectively we would have to remove the type hint. I am not convinced if that were the better option. At least it would make it harder to find some bugs in userland code.

Speaking about the attributes: Making this change would be compatible with how the Symfony serializer handles groups and attributes. So it would make the behaviour more consistent with this implementation (you couldn't have a custom `groups` attribute here anyway). However, we would introduce a small BC break for the JMS serializer like you said. I am actually not completely sure if we should go that way (it may confuse people who are used to the JMS serializer that there are some attributes that have a special meaning).

GuilhemN:
Sorry i forgot to answer you...
I don't think the symfony serializer will ever support exclusion strategies so imo it's fine to couple this feature to the jms serializer.

PR #1584, opened 2016-10-07
Option to skip RequestBody validation

merk:
I have a requirement that the RequestBodyParamConverter doesnt perform the validation, so my controller action can calculate which validation groups are required and do it itself.

Seems to me like its logical that this is configurable on ParamConverter.

### DIFF
--- a/Request/RequestBodyParamConverter.php
+++ b/Request/RequestBodyParamConverter.php
@@ -105,7 +105,7 @@ class RequestBodyParamConverter implements ParamConverterInterface
 
         $request->attributes->set($configuration->getName(), $object);
 
-        if (null !== $this->validator) {
+        if (null !== $this->validator && (!isset($options['skipValidation']) || !$options['skipValidation'])) {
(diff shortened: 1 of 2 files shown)


Request/RequestBodyParamConverter.php
```diff
@@ -105,7 +105,7 @@ public function apply(Request $request, ParamConverter $configuration)
 
         $request->attributes->set($configuration->getName(), $object);
 
-        if (null !== $this->validator) {
+        if (null !== $this->validator && (!isset($options['skipValidation']) || !$options['skipValidation'])) {
```
Ener-Getick: What about `validate`?

Ener-Getick: It looks very verbose, let's keep something short :)

merk: Changed to `validate`

merk:
@GuilhemN is there anything stopping this from being merged?

GuilhemN:
Thank you @merk!

Issue #1571, opened 2016-09-12
fos_rest.exception.codes configuration bug?


tael:
Hi, I'm testing `Exception Handling` and stopped by a little problem belows.

I'm not sure this is a bug or NOT. But if not, I wish someone helps me! :D

Symfony\Component\HttpFoundation\Response::HTTP_* constant does not works.

from the manual ( http://symfony.com/doc/master/bundles/FOSRestBundle/4-exception-controller-support.html )
> ...
> To map Exception classes to HTTP response status codes an exception map may be configured, where the keys match a fully qualified class name and the values are either an integer HTTP response status code or a string matching a class constant of the Symfony\Component\HttpFoundation\Response class:
> ...

This Configuration does works, well
```
fos_rest.exception.codes
            'Symfony\Component\Security\Core\Exception\AccessDeniedException': 403
```
But, This configuration does not works

```
fos_rest.exception.codes
            'Symfony\Component\Security\Core\Exception\AccessDeniedException': HTTP_FORBIDDEN
```

tael:
I reviewed source code,
1.5 looks work maybe,
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/1.5/Util/Codes.php
but 2.0 ? 404 not found this class
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/2.0/Util/Codes.php
is it deprecated?

mvrhov:
The symfony does contain all the codes necessary and thus that class is gone in 2.0

tael:
yes I agree.  but HTTP_FORBIDDEN is not works. 
this feature is gone also?

edefimov:
@tael No, it's not gone, it looks like a bug in the bundle. [This PR](https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1585) fixes this issue

xabbuh:
This was fixed by #1585.

PR #1585, opened 2016-10-07
#1571 Fixed using Response constants in exception codes configuration

edefimov:
According to [configuration](http://symfony.com/doc/master/bundles/FOSRestBundle/4-exception-controller-support.html) we can use constants from Symfony Response class as a value for `exception.codes` configuration section.
However when the bundle is configured in such a way, the constant is not converted into corresponding http status code what leads to error during run time. This PR fixes incorrect behavior and converts constant to value during configuration normalization.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -14,6 +14,7 @@ namespace FOS\RestBundle\DependencyInjection;
 use Symfony\Component\Config\Definition\Builder\ArrayNodeDefinition;
 use Symfony\Component\Config\Definition\Builder\TreeBuilder;
 use Symfony\Component\Config\Definition\ConfigurationInterface;
+use Symfony\Component\Config\Definition\Exception\InvalidConfigurationException;
 use Symfony\Component\HttpFoundation\Response;
 
 /**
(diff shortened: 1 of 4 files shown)


Ener-Getick:
Could you add tests for that please?

Tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -0,0 +1,138 @@
...
+     */
+    private $processor;
+
+    /**
+     * testAcceptsIntegers.
+     */
+    public function testAcceptsIntegers()
+    {
+        $expectedConfig = [
+            md5(microtime()) => mt_rand(200, 599),
```
Ener-Getick: I think something static would be better :)

Tests/DependencyInjection/ConfigurationTest.php
```diff
@@ -0,0 +1,138 @@
...
+     * testThatResponseConstantsConvertedToCodes.
+     */
+    public function testThatResponseConstantsConvertedToCodes()
+    {
+        $ref = new \ReflectionClass(Response::class);
+        $expectedResult = [];
+        $config = [];
+
+        foreach ($ref->getConstants() as $constantName => $value) {
+            if (strpos($constantName, 'HTTP_') !== 0) {
```
Ener-Getick: Same here, no need to test all the constants ;)

Tests/DependencyInjection/ConfigurationTest.php
edefimov: It could be a very relyable test :)

Ener-Getick:
I went through our extension, and in fact it looks like this is already managed [here](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/DependencyInjection/FOSRestExtension.php#L342-L347). Are you sure this doesn't work?

edefimov:
Yes. Well, the whole picture is that on the lines [329-332](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/2.1.0/DependencyInjection/FOSRestExtension.php#L329-L332) the configuration is injected into the service as is and only after that the constants are actually replaced with their values. So, I think the complete fix should remove the substitution logic from extension and move exception class validation from extension into configuration.

Ener-Getick:
No problem for me :) We just need to remove these lines of the extension then.

DependencyInjection/Configuration.php
```diff
@@ -389,15 +390,51 @@ private function addExceptionSection(ArrayNodeDefinition $rootNode)
...
+                                            );
+                                        }
+
+                                        $item = constant('Symfony\Component\HttpFoundation\Response::'.$item);
+                                    }
+
+                                    return $items;
+                                })
                             ->end()
                             ->prototype('scalar')->end()
```
Ener-Getick: Can be changed to `int` .

Ener-Getick:
Except my last comment, LGTM :+1:

edefimov:
Done

Ener-Getick:
Thank you @edefimov !

PR #1587, opened 2016-10-11
Add support \Twig_Template on set template.

liverbool:
Add support \Twig_Template on set template.


stof:
what is the use case for passing the Twig_Template instance rather than the template name ?

liverbool:
@stof when we need to create template from string.

```php
$template = $twig->createTemplate("Example {{ string }}");
$view = View::create();
$view->setTemplate($template);
// ...
```

Ener-Getick:
Can you add tests for that please?

Ener-Getick:
Can you also add a functional test? I'm not sure it will work nicely with the rest of the `ViewHandler`.

Tobion:
Closing as the namespaces twig classes should be used now. So this is outdated in its current form.

PR #1589, opened 2016-10-11
explicitly list all required packages

xabbuh:
Instead of relying on the Symfony FrameworkBundle to pull in all our
needed dependencies, we should explicitly list all of them. This will
ensure that nothing breaks when the FrameworkBundle stops depending on
some optional dependencies (which was done recently).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -23,9 +23,17 @@
     "require": {
         "php":                            "^5.5.9|~7.0",
         "psr/log":                        "^1.0",
-        "symfony/framework-bundle":       "^2.7|^3.0",
+        "symfony/config":                 "^2.7|^3.0",
+        "symfony/debug":                  "^2.7|^3.0",
+        "symfony/dependency-injection":   "^2.7|^3.0",
+        "symfony/event-dispatcher":       "^2.7|^3.0",
         "symfony/finder":                 "^2.7|^3.0",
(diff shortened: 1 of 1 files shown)


xabbuh:
see symfony/symfony#15748 and the related pull requests for some context

composer.json
```diff
@@ -23,9 +23,17 @@
...
+        "symfony/config":                 "^2.7|^3.0",
+        "symfony/debug":                  "^2.7|^3.0",
+        "symfony/dependency-injection":   "^2.7|^3.0",
+        "symfony/event-dispatcher":       "^2.7|^3.0",
         "symfony/finder":                 "^2.7|^3.0",
+        "symfony/framework-bundle":       "^2.7|^3.0",
+        "symfony/http-foundation":        "^2.7|^3.0",
+        "symfony/http-kernel":            "^2.7|^3.0",
         "symfony/routing":                "^2.7|^3.0",
+        "symfony/security-core":          "^2.7|^3.0",
```
Ener-Getick: Do we depend on this?

xabbuh: Yes, the `AccessDeniedListener` makes use of the `AccessDeniedException` and `AuthenticationException` classes.

Ener-Getick: Ok, thanks :)

Ener-Getick:
LGTM :+1:

PR #1590, opened 2016-10-11
drop the build job using Symfony 3.0

xabbuh:
Symfony 3.0 is not forward-compatible with Twig 2.0 (it references Twig
extensions by their name instead of the extension's FQCN) and won't be
updated anymore as it already reached end of maintenance. This currently
lets our Travis CI builds fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,8 +25,6 @@ matrix:
     - php: 7.0
       env: SYMFONY_VERSION='2.8.*'
     - php: 7.0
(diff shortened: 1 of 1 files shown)


Ener-Getick:
BTW maybe we should move our constraints to 3.1 if it is not maintened anymore?

xabbuh:
Well, the code will still work. So I think it's not needed to intentionally break people's apps. But as they won't get any Symfony support anymore it's imo okay to not focus on this release too much.

PR #1591, opened 2016-10-12
the Regex constraint is resolvable

### LINKED ISSUE #1575: System parameters cannot be used as requirements on a QueryParam annotation
The following code does not work when a 'my_regex' parameter is listed in the parameters.yml file. Stepping through the code, this seems to be because the code sees this as class Fos\RestBundle\Validator\Constraints\Regex, which does not implement the ResolvableConstraintInterface. This happens even when my_regex is set to a simple string like 'test'

`* @QueryParam(name="departureTime", requirements="%my_regex%", strict=true, nullable=false)`

xabbuh:
This fixes #1575.

### DIFF
--- a/Tests/Functional/Bundle/TestBundle/Controller/ParamFetcherController.php
+++ b/Tests/Functional/Bundle/TestBundle/Controller/ParamFetcherController.php
@@ -11,12 +11,10 @@
 
 namespace FOS\RestBundle\Tests\Functional\Bundle\TestBundle\Controller;
 
-use FOS\RestBundle\Controller\Annotations\FileParam;
 use FOS\RestBundle\Controller\FOSRestController;
 use FOS\RestBundle\Controller\Annotations\QueryParam;
 use FOS\RestBundle\Controller\Annotations\RequestParam;
 use FOS\RestBundle\Request\ParamFetcherInterface;
-use Symfony\Component\HttpFoundation\File\UploadedFile;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpKernel\HttpKernelInterface;
--- a/Tests/Functional/Bundle/TestBundle/Controller/ParamFetcherController.php
+++ b/Tests/Functional/Bundle/TestBundle/Controller/ParamFetcherController.php
@@ -11,10 +11,12 @@
 
 namespace FOS\RestBundle\Tests\Functional\Bundle\TestBundle\Controller;
 
+use FOS\RestBundle\Controller\Annotations\FileParam;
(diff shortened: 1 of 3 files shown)


Ener-Getick:
LGTM :+1:
Could you add a requirement [here](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Tests/Functional/Bundle/TestBundle/Controller/ParamFetcherController.php#L39) using parameter injection at the same time? It should ensure the annotation actually works.

xabbuh:
@Ener-Getick For now, I failed to create a proper test (admittedly I didn't invest too much time in it). Do you have time to take over?

Ener-Getick:
Not tonight but i should be able to do it tomorrow :)

Ener-Getick:
Test case added.

Tests/Functional/Bundle/TestBundle/Controller/ParamFetcherController.php
```diff
@@ -11,10 +11,12 @@
 
 namespace FOS\RestBundle\Tests\Functional\Bundle\TestBundle\Controller;
 
+use FOS\RestBundle\Controller\Annotations\FileParam;
```
xabbuh: The added annotations are not needed, right?

Ener-Getick: Indeed, i forgot to close my opened files before changing branch.

Tests/Functional/Bundle/TestBundle/Controller/ParamFetcherController.php
```diff
@@ -23,13 +23,13 @@
 class ParamFetcherController extends FOSRestController
 {
     /**
-     * @RequestParam(name="raw", requirements=@IdenticalTo({"foo"="raw", "bar"="foo"}), default="invalid")
-     * @RequestParam(name="map", map=true, requirements=@IdenticalTo({"foo"="map", "foobar"="foo"}), default="%invalid2% %%")
-     * @RequestParam(name="bar", map=true, requirements="%bar% foo", strict=true)
+     * @RequestParam(name="raw", requirements=@IdenticalTo({"foo"="raw", "bar"="foo"}), default="invalid", strict=false)
+     * @RequestParam(name="map", map=true, requirements=@IdenticalTo({"foo"="map", "foobar"="foo"}), default="%invalid2% %%", strict=false)
+     * @RequestParam(name="bar", nullable=true, requirements="%bar%\ foo")
```
Ener-Getick: Btw this is really not expected. It is due to the `x` option and is not convenient imo, unfortunately it's gonna be hard to change...

GuilhemN:
Thank you @xabbuh !

PR #1597, opened 2016-10-19
Ignoring PSR7 request typehint in route loader

xabbuh:


Ener-Getick:
In fact didn't it break because of the use of reflection? As we removed it, i think we can safelly readd the interface directly.

xabbuh:
@Ener-Getick You are right. I removed my commit and kept only the one from #1229.

Ener-Getick:
Thank you @xabbuh

claudiobizzotto:
Awesome! Thanks :thumbsup:

Issue #1604, opened 2016-11-10
Versioning with path parameter doesn't set exclusion strategy version


TobiasXy:
Currently I'm setting up versioning for my Rest API which is already using this bundle (and the Symfony Serializer).

I need the requested version in my serializer, because the Controller only returns an entity, but depending on the requested version it should be normalized with a different representation.

If I provide the version through a query string parameter or a header it works fine and I can access the version within the normalizer using `$context['version']`. If I set the version as a path parameter (like `GET /v1/some/entity/123`) it isn't available in the $context variable (i.e. it is the default version), but it is set in the `$request->attributes` parameter bag as expected.

Now after reading a bit through the code the reason seems to be:
For versioning using a query string parameter or a header, the VersionListener

1.  uses a VersionResolver to get the requested version from the request
2.  sets the "version" request attribute
3.  calls the `setExclusionStrategyVersion` method of the ViewHandler

Now for a version parameter in the path, there is no VersionResolver. This is not a problem if you only need the version in the controller, since symfony itself sets the "version" request attribute, but in the Serializer you only get the default version, because of that.

If this is considered to be working as intended for some reason just close this issue and I would write a custom EventListener to do this for me. Otherwise I could create a pull request to fix that.

xabbuh:
This indeed looks like an issue to me. I suggest to split the `VersionListener` to fix this bug. The existing `VersionListener` should only be responsible to set the `version` request attribute when necessary. We can then introduce a new event listener that is executed after the `VersionListener` and sets the view handler's exclusion strategy version based on the request's version attribute (if set).

GuilhemN:
@xabbuh what about adding a `RequestAttributeVersionResolver` instead?

xabbuh:
@GuilhemN But we actually do not need to resolve the attribute anymore. We get that for free thanks to the Symfony routing system. So I don't think we should add this resolver at all.

GuilhemN:
Then maybe we could do in the same listener (at the end):
```php
if ($request->attributes->has('version') && $this->viewHandler instanceof ConfigurableViewHandlerInterface) {
    $this->viewHandler->setExclusionStrategyVersion($request->attributes->get('version'));
}
```
?

xabbuh:
That would work too.

GuilhemN:
It's just I think having a new listener is overkill but if you think it's worth it, we can create it.

PR #1605, opened 2016-11-11
Support path versioning for rendering responses

GuilhemN:
see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1604

### DIFF
--- a/EventListener/VersionListener.php
+++ b/EventListener/VersionListener.php
@@ -25,23 +25,12 @@ class VersionListener
     private $viewHandler;
     private $versionResolver;
     private $defaultVersion;
-    private $version = false;
 
     public function __construct(ViewHandlerInterface $viewHandler, VersionResolverInterface $versionResolver, $defaultVersion = null)
     {
         $this->viewHandler = $viewHandler;
         $this->versionResolver = $versionResolver;
         $this->defaultVersion = $defaultVersion;
-    }
-
-    /**
-     * Gets the version.
-     *
-     * @return mixed
-     */
-    public function getVersion()
-    {
-        return $this->version;
     }
 
     public function onKernelRequest(GetResponseEvent $event)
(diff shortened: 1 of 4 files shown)


EventListener/VersionListener.php
```diff
@@ -53,16 +53,18 @@ public function onKernelRequest(GetResponseEvent $event)
         }
 
         $this->version = $this->versionResolver->resolve($request);
-        if (false === $this->version && null !== $this->defaultVersion) {
+        if (false === $this->version && null !== $this->defaultVersion && !$request->attributes->has('version')) {
             $this->version = $this->defaultVersion;
         }
```
xabbuh: What about changing this to:

```php
if (false === $this->version && $request->attributes->has('version')) {
    $this->version = $request->attributes->get('version');
} elseif (false === $this->version && null !== $this->defaultVersion) {
    $this->version = $this->defaultVersion;
}
```

xabbuh: The code below could then be simplified a bit.

xabbuh: By the way, what was the reason for the `getVersion()` method which requires us to make the event listener stateful?

GuilhemN: In fact we could remove it as we made listeners `@internal`. WDYT?

xabbuh: Oh, I missed the fact that we made it internal in 1.8. So yes, I would update it accordingly. But we can do that in a separate PR.

EventListener/VersionListener.php
```diff
@@ -53,16 +53,24 @@ public function onKernelRequest(GetResponseEvent $event)
...
         $this->version = $this->versionResolver->resolve($request);
+        if (false === $this->version && $request->attributes->has('version')) {
+            $this->version = $request->attributes->get('version');
+        } elseif (false === $this->version && null !== $this->defaultVersion) {
+            $this->version = $this->defaultVersion;
+        }
+
         if (false === $this->version && null !== $this->defaultVersion) {
             $this->version = $this->defaultVersion;
         }
```
xabbuh: This block should be removed now.

EventListener/VersionListener.php
```diff
@@ -53,16 +53,24 @@ public function onKernelRequest(GetResponseEvent $event)
...
+        // Override request attribute
         if (false !== $this->version) {
             $request->attributes->set('version', $this->version);
+        }
 
-            if ($this->viewHandler instanceof ConfigurableViewHandlerInterface) {
-                $this->viewHandler->setExclusionStrategyVersion($this->version);
-            }
+        // Use the resolved version when rendering the response
+        if ($request->attributes->has('version') && $this->viewHandler instanceof ConfigurableViewHandlerInterface) {
```
xabbuh: We can now save a method call here by checking for `$this->version` instead of the request attributes.

EventListener/VersionListener.php
GuilhemN: That's alright, it's not much to do and I won't forget it if I do it now :stuck_out_tongue:

GuilhemN:
Do you see something else to change, or can I merge this? :)

xabbuh:
👍 failures do not seem to be related

Tests/Functional/VersionTest.php
```diff
@@ -79,4 +79,16 @@ public function testDefaultVersion()
...
+    public function testVersionInPath()
+    {
+        $this->client->request(
+            'GET',
+            '/version/2.3',
+            [],
+            [],
+            ['HTTP_Accept' => 'application/json']
+        );
+        $this->assertEquals('{"version":"2.3"}', $this->client->getResponse()->getContent());
```
GuilhemN: I agree, will do it.

PR #1609, opened 2016-11-18
fix parameter incompatibles attribute

### LINKED ISSUE #1608: QueryParam 'incompatibles' attribute does not work as expected
According to the official [symfony docs](http://symfony.com/doc/current/bundles/FOSRestBundle/param_fetcher_listener.html), using the 'incompatibles' attribute it is possible to prevent clients requesting n parameters simultaneously:

> @RequestParam(name="search", requirements="[a-z]+", description="search")
     * @RequestParam(name="byauthor", requirements="[a-z]+", description="by author", incompatibles={"search"})
     * Imagine you have an api for a blog with to get Articles with two ways of filtering
     *   1 by filtering the text
     *   2 by filtering the author
     * and you don't have yet implemented the possibility to filter by both at the same time.
     * In order to prevent clients from doing a request with both (which will produce not the expected
     * resut and is likely to be considered as a bug) you can precise the parameters can't be present
     * at the same time by doing

> 

This is my set up:

```
/**
   * @Annotations\Get(path="/test", options={"expose" = true})
   * @Annotations\QueryParam(
   *    name="search",
   *    requirements="[a-z]+",
   *    description="search"
   * )
   * @Annotations\QueryParam(
   *    name="byauthor",
   *    requirements="[a-z]+",
   *    description="by author",
   *    incompatibles={"search"}
   * )
   */
```

If I call the following url:
```
/test?search=text&byauthor=author
```
as expected, I receive a 400 Bad Request with the following error message:

> 'byauthor' param is incompatible with search param. 

But if i call the following url:
```
/test?search=text
```
I receive a 400 Bad Request with  message:

> 'byauthor' param is incompatible with search param.

IndraGunawan:
fix #1608

### DIFF
--- a/Request/ParamFetcher.php
+++ b/Request/ParamFetcher.php
@@ -195,23 +195,25 @@ class ParamFetcher implements ParamFetcherInterface
      */
     protected function checkNotIncompatibleParams(ParamInterface $param)
     {
+        if (null === $param->getValue($this->getRequest(), null)) {
+            return;
+        }
+
         $params = $this->getParams();
-        if (null !== $param->getValue($this->getRequest(), null)) {
-            foreach ($param->getIncompatibilities() as $incompatibleParamName) {
-                if (!array_key_exists($incompatibleParamName, $params)) {
-                    throw new \InvalidArgumentException(sprintf("No @ParamInterface configuration for parameter '%s'.", $incompatibleParamName));
-                }
-                $incompatibleParam = $params[$incompatibleParamName];
+        foreach ($param->getIncompatibilities() as $incompatibleParamName) {
+            if (!array_key_exists($incompatibleParamName, $params)) {
+                throw new \InvalidArgumentException(sprintf("No @ParamInterface configuration for parameter '%s'.", $incompatibleParamName));
+            }
+            $incompatibleParam = $params[$incompatibleParamName];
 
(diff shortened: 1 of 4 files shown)


Request/ParamFetcher.php
```diff
@@ -196,20 +196,22 @@ public function all($strict = null)
...
-
-            if ($incompatibleParam->getValue($this->getRequest(), null) !== null) {
-                $exceptionMessage = sprintf(
-                    "'%s' param is incompatible with %s param.",
-                    $param->getName(),
-                    $incompatibleParam->getName()
-                );
-
-                throw new BadRequestHttpException($exceptionMessage);
+        if (null !== $param->getValue($this->getRequest(), null)) {
```
IndraGunawan: ok. i will change it

GuilhemN:
It looks like tests are broken.

GuilhemN:
LGTM :+1: 

@xabbuh @lsmith77 is it ok for you as a bug fix or should it be considered a bc break?

xabbuh:
I would say this is a bugfix. Question is how likely is it that someone relied on the previous behaviour?

GuilhemN:
@xabbuh I think the expected behavior is the one of this PR. I don't use this feature myself but I would say it's unlikely that someone relied on the previous behavior.

GuilhemN:
Thank you @IndraGunawan !

Issue #614, opened 2013-11-24
Cannot overwrite rules configuration


jmcclell:
Currently, the TreeBuilder for FOSRestBundle's configuration has the cannotBeOverwritten() function applied. Why is this necessary?

It prevents me from having different rules in dev and production environments. I wanted to have my production configuration allow JSON only while in development also allow HTML with a custom view listener which showed the JSON response along with the Symfony toolbar and other debug information. I am not sure how best to go about doing that without being able to overwrite the configuration in my dev config file.

PR #1613, opened 2016-11-21
Allow overwriting rules in configuration

alcaeus:
This was previously mentioned in #614. To quote from that ticket:

> the issue is that the order matters and its not possible to ensure that configuration is merged between environments in a way that matches the users expectations

The same effect can be achieved by using `performNoDeepMerging`, which allows overwriting the configuration (e.g. in config_dev.yml for testing purposes) but still keeps the order of the rules as specified.


GuilhemN:
That's possible but that's not safe (people may not realize that they did something that won't count).
You should instead put your prod config in `config_prod.yml` and your dev config in `config_dev.yml`.

lsmith77:
I guess we should close this PR? @alceus do you think we should expand the docs on this topic?

alcaeus:
TBH, I don't understand why you shouldn't be able to overwrite them. Merging them would not make sense, but not allowing them to be overwritten at all seems weird. Note that this doesn't just affect the default prod/dev config files, but any configuration that includes another config file (for whatever reason that may be).

alcaeus:
@lsmith77 @GuilhemN any chance this will get merged or would you rather document why configuration shouldn't be overwritten or merged?

GuilhemN:
Rethinking about it, this method is used [in the `SecurityBundle`](https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/SecurityBundle/DependencyInjection/MainConfiguration.php) and we have quite the same priorities issues. So I'm :+1: for merging this.

alcaeus:
Looks like I missed both this pull requests' birthday as well as the 2.3.0 release. Is this something that could make it into 2.3.1?

alcaeus:
@GuilhemN any plans to merge this or do you want me to just close it?

xabbuh:
@alcaeus Can you rebase on the latest `master` branch?

xabbuh:
never mind, it was easy to do, see #1973

PR #1611, opened 2016-11-18
Update `SerializerConfigurationPass::process()` in order to respect serializer service inference

phansys:
|Q            |A     |
|---          |---   |
|Branch       |master|
|Bug fix?     |yes   |
|New feature? |no    |
|BC breaks?   |no    |
|Deprecations?|no    |
|Tests pass?  |yes   |
|Fixed tickets|n/a   |
|License      |MIT   |
|Doc PR       |n/a   |

The alias `serializer` for `jms_serializer.serializer ` service can be disabled via `enable_short_alias` setting. See https://github.com/schmittjoh/JMSSerializerBundle/commit/fec96d9a2ab7260382cbaa43e5744488ee330769.
Without this fix, if this flag is configured as `false`, the expected service inference is not respecting the desired order:
1. The one you configured using fos_rest.service.serializer (if you did).
2. The JMS serializer, if the JMSSerializerBundle is available (and registered).
3. The Symfony Serializer if it's enabled (or any service called serializer).

PR #1622, opened 2016-12-02
Fix serializers configuration priority

GuilhemN:
See https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1611

### DIFF
--- a/DependencyInjection/Compiler/SerializerConfigurationPass.php
+++ b/DependencyInjection/Compiler/SerializerConfigurationPass.php
@@ -41,31 +41,26 @@ final class SerializerConfigurationPass implements CompilerPassInterface
             throw new \InvalidArgumentException('Neither a service called "jms_serializer.serializer" nor "serializer" is available and no serializer is explicitly configured. You must either enable the JMSSerializerBundle, enable the FrameworkBundle serializer or configure a custom serializer.');
         }
 
-        if ($container->has('serializer')) {
-            $class = $container->getParameterBag()->resolveValue(
-                $container->findDefinition('serializer')->getClass()
-            );
-
-            if (is_subclass_of($class, 'Symfony\Component\Serializer\SerializerInterface')) {
-                $container->setAlias('fos_rest.serializer', 'fos_rest.serializer.symfony');
-                $container->removeDefinition('fos_rest.serializer.exception_wrapper_serialize_handler');
-            } elseif (is_subclass_of($class, 'JMS\Serializer\SerializerInterface')) {
-                $container->setAlias('fos_rest.serializer', 'fos_rest.serializer.jms');
-            } elseif (is_subclass_of($class, 'FOS\RestBundle\Serializer\Serializer')) {
-                $container->setAlias('fos_rest.serializer', 'serializer');
-            } else {
-                throw new \InvalidArgumentException(sprintf('"fos_rest.serializer" must implement FOS\RestBundle\Serializer\Serializer (instance of "%s" given).', $class));
-            }
-
-            return;
-        } else {
-            $container->removeDefinition('fos_rest.serializer.exception_wrapper_normalizer');
-        }
-
         if ($container->has('jms_serializer.serializer')) {
             $container->setAlias('fos_rest.serializer', 'fos_rest.serializer.jms');
+
+            return;
+        }
+
+        // Remove the exception normalizer linked to the jms normalizer
+        $container->removeDefinition('fos_rest.serializer.exception_normalizer.jms');
+
+        // As there is no `jms_serializer.serializer` service, there is a `serializer` service
+        $class = $container->getParameterBag()->resolveValue(
+            $container->findDefinition('serializer')->getClass()
+        );
+
+        if (is_subclass_of($class, 'Symfony\Component\Serializer\SerializerInterface')) {
+            $container->setAlias('fos_rest.serializer', 'fos_rest.serializer.symfony');
+        } elseif (is_subclass_of($class, 'FOS\RestBundle\Serializer\Serializer')) {
+            $container->setAlias('fos_rest.serializer', 'serializer');
         } else {
-            $container->removeDefinition('fos_rest.serializer.exception_wrapper_serialize_handler');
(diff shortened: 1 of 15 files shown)


Tests/DependencyInjection/Compiler/SerializerConfigurationPassTest.php
```diff
@@ -62,24 +62,24 @@ public function testShouldConfigureJMSSerializer()
...
+    public function testJmsSerializerServiceSupersedesSerializerService()
     {
         $this->container->register('jms_serializer.serializer', 'JMS\Serializer\Serializer');
         $this->container->register('serializer', 'Symfony\Component\Serializer\Serializer');
 
         $compiler = new SerializerConfigurationPass();
         $compiler->process($this->container);
 
-        $this->assertSame('fos_rest.serializer.symfony', (string) $this->container->getAlias('fos_rest.serializer'));
+        $this->assertSame('fos_rest.serializer.jms', (string) $this->container->getAlias('fos_rest.serializer'));
```
GuilhemN: @xabbuh this was wrong according to the docs but it is technically a bc break. Is it fine for you anyway?

lyrixx: This is a huge BC break unfortunately. But acutally, I prefer this versio of the code... ;)

.styleci.yml
```diff
@@ -1,3 +1,4 @@
 preset: symfony
 disabled:
   - braces
+  - php_unit_fqcn_annotation
```
phansys: This change doesn't seem related to the subject.

DependencyInjection/Configuration.php
```diff
@@ -59,12 +59,14 @@ public function getConfigTreeBuilder()
                 ->arrayNode('access_denied_listener')
                     ->canBeEnabled()
                     ->beforeNormalization()
-                        ->ifArray()->then(function ($v) { if (!empty($v) && empty($v['formats'])) {
-     unset($v['enabled']);
-     $v = ['enabled' => true, 'formats' => $v];
- }
+                        ->ifArray()->then(function ($v) {
+                            if (!empty($v) && empty($v['formats'])) {
```
phansys: Why not simply `$v && !$v['formats']`?

DependencyInjection/Configuration.php
```diff
@@ -59,12 +59,14 @@ public function getConfigTreeBuilder()
...
+                                unset($v['enabled']);
```
phansys: `$v['enabled']` is overridden at the next line, which is the reason for this `unset()` call?

DependencyInjection/Configuration.php
```diff
@@ -59,12 +59,14 @@ public function getConfigTreeBuilder()
                 ->arrayNode('access_denied_listener')
                     ->canBeEnabled()
                     ->beforeNormalization()
-                        ->ifArray()->then(function ($v) { if (!empty($v) && empty($v['formats'])) {
-     unset($v['enabled']);
-     $v = ['enabled' => true, 'formats' => $v];
- }
+                        ->ifArray()->then(function ($v) {
```
phansys: `$v` argument could be declared as `array`.

.styleci.yml
GuilhemN: styleci failed without it

DependencyInjection/Configuration.php
GuilhemN: that's not related, just a cs fix

DependencyInjection/Configuration.php
GuilhemN: that's not related, please open another issue

xabbuh:
@GuilhemN I would prefer if we could split this PR into two commits. One for the code style changes and one for the actual fix which would make it easier to use git blame in the future.

GuilhemN:
@xabbuh sure, it's done.

DependencyInjection/Compiler/SerializerConfigurationPass.php
```diff
@@ -41,31 +41,26 @@ public function process(ContainerBuilder $container)
...
         }
 
-        if ($container->has('jms_serializer.serializer')) {
-            $container->setAlias('fos_rest.serializer', 'fos_rest.serializer.jms');
+        // Remove the exception normalizer linked to the jms normalizer
+        $container->removeDefinition('fos_rest.serializer.exception_normalizer.jms');
+
+        // As there is no `jms_serializer.serializer` service, there is a `serializer` service
+        $class = $container->getParameterBag()->resolveValue(
+            $container->findDefinition('serializer')->getClass()
```
GuilhemN: This is already managed [here](https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1622/files/8d8f4452e049c0cf2e9f59b2ef8a99486447868a..717cab75b4a5567942cd024ca5e86468fd2b113a#diff-b4c2bb665099019e5209acdfe6371d2aR41).

DependencyInjection/Compiler/SerializerConfigurationPass.php
```diff
@@ -41,31 +41,26 @@ public function process(ContainerBuilder $container)
...
+        );
+
+        if (is_subclass_of($class, 'Symfony\Component\Serializer\SerializerInterface')) {
+            $container->setAlias('fos_rest.serializer', 'fos_rest.serializer.symfony');
+        } elseif (is_subclass_of($class, 'FOS\RestBundle\Serializer\Serializer')) {
+            $container->setAlias('fos_rest.serializer', 'serializer');
```
xabbuh: What about the case of `JMS\Serializer\SerializerInterface`?

GuilhemN: It is already supported under `jms_serializer.serializer`. I don't think we need to support the case where the jms serializer is only configured under `serializer`. Wdyt?

GuilhemN: I don't think it has ever been supported, what's supported is [removing the `serializer` alias](https://github.com/schmittjoh/JMSSerializerBundle/blob/master/DependencyInjection/Configuration.php#L45).

xabbuh: 👍 then for me

DependencyInjection/Configuration.php
GuilhemN: Good question, it should probably be removed.

GuilhemN:
Thanks for your reviews !

PR #1625, opened 2016-12-03
change method to check subclass

IndraGunawan:
change `is_subclass_of` to `ReflectionClass::implementsInterface` reported by Scrutinizer (https://scrutinizer-ci.com/g/FriendsOfSymfony/FOSRestBundle/inspections/6de3a7c5-9dea-436d-8bcb-bef87bb7f044/issues/files/DependencyInjection/Compiler/SerializerConfigurationPass.php?status=new&orderField=path&order=asc&honorSelectedPaths=0)


GuilhemN:
[This bug](https://bugs.php.net/bug.php?id=53727) was fixed a while ago in php 5.3 and `is_subclass_of` is faster, so there's no need to change it.

PR #1629, opened 2016-12-06
Fixed typo

lyrixx:
See: https://github.com/FriendsOfSymfony/FOSRestBundle/blob/1.5/DependencyInjection/Compiler/SerializerConfigurationPass.php#L24

### DIFF
--- a/Resources/doc/1-setting_up_the_bundle.rst
+++ b/Resources/doc/1-setting_up_the_bundle.rst
@@ -43,7 +43,7 @@ This bundle needs a serializer to work correctly. In most cases,
 you'll need to enable a serializer or install one. This bundle tries
 the following (in the given order) to determine the serializer to use:
 
-#. The one you configured using ``fos_rest.services.serializer`` (if you did).
+#. The one you configured using ``fos_rest.serializer`` (if you did).
 #. The JMS serializer, if the `JMSSerializerBundle`_ is available (and registered).
 #. The `Symfony Serializer`_ if it's enabled (or any service called ``serializer``).
 


GuilhemN:
Actually it is talking about the config here.

lyrixx:
> Actually it is talking about the config here.

I did not understand your comment. What do you mean?

lyrixx:
And BTW, the order is wrong (at least in for 2.0)

GuilhemN:
I'm refering to the [Configuration](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/DependencyInjection/Configuration.php#L119). But there's still a typo, it should be `fos_rest.service.serializer`.

lyrixx:
I detected this issue, because in my project I have the SF serializer **and** the JMS one. But, for legacy reason, fos rest should use the JMS one. And right now, it's not possible with ease. The only solution I see is to create a compiler pass, that should pass **after** your to update the alias.

(I'm using 2.0)

GuilhemN:
https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1622 fixes it :)

lyrixx:
I don't understand the code here. There is a configuration available, but it can not find where it is used.

Then, in the [pass](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/DependencyInjection/Compiler/SerializerConfigurationPass.php#L29) the code is looking for a service with id `fos_rest.serializer`.

GuilhemN:
It is used [here](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/DependencyInjection/FOSRestExtension.php#L66-L70) and creates this services.

xabbuh:
Though I think that we could indeed improve this sentence to become something like this:

The one you configured using the `fos_rest.service.serializer` option (if you did).

Beware that you will probably have to rebase your changes on the `2.0` branch (the service/services mismatch is already solved on that branch).

PR #1631, opened 2016-12-07
[Doc] Fix invalid usuage of annitation

mimol91:
Using View for both creating View and as annotations cause error 
`The class \"FOS\\RestBundle\\View\\View\" is not annotated with @Annotation`

Fixed to use `View` for creating object and `ViewAnnotation` as annotation


xabbuh:
👍

PR #1640, opened 2017-01-01
move license to project root and add years

xabbuh:
Moving the file allows GitHub to automatically detect the license. The years are required for a valid MIT license afaik.


GuilhemN:
Thank you @xabbuh !

Issue #1507, opened 2016-06-01
Disable 'method_prefix' globally


JHGitty:
Is it possible to disable `method_prefix` globally?
I don't want to set it to `false` on each action.

http://symfony.com/doc/current/bundles/FOSRestBundle/7-manual-route-definition.html#method-name-prefix

Ener-Getick:
This is not possible for now but it could be a great improvement :+1:

xabbuh:
closing here as implemented in #1643

PR #1643, opened 2017-01-05
Disable 'method_prefix' globally

jorge07:
Add support for https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1507

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -103,7 +103,7 @@ final class Configuration implements ConfigurationInterface
                     ->addDefaultsIfNotSet()
                     ->children()
                         ->scalarNode('default_format')->defaultNull()->end()
-                        ->scalarNode('method_prefix')->defaultTrue()->end()
+                        ->scalarNode('prefix_methods')->defaultTrue()->end()
                         ->scalarNode('include_format')->defaultTrue()->end()
                     ->end()
                 ->end()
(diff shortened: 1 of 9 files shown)


DependencyInjection/Configuration.php
```diff
@@ -103,6 +103,7 @@ public function getConfigTreeBuilder()
                     ->addDefaultsIfNotSet()
                     ->children()
                         ->scalarNode('default_format')->defaultNull()->end()
+                        ->scalarNode('method_prefix')->defaultTrue()->end()
```
xabbuh: By just looking at the name of the option I would expect to be able to pass a string that will be the prefix. To have an option acting as a toggle to turn on/off the feature I would rather rename it to something like `prefix_methods`.

jorge07: I just followed the same behavior as the existing option that appears in the doc http://symfony.com/doc/current/bundles/FOSRestBundle/7-manual-route-definition.html#method-name-prefix

But maybe we can add that option too

jorge07: Ping @xabbuh . Should I change it?

xabbuh: Yeah, I think naming that option `method_name_prefix` was not the best idea. Let's not make the same mistake here again. Can you please rename it? :)

jorge07: Done

PR #1645, opened 2017-01-05
Fixed regular expression pattern used to detect custom routes in RestRouteCollection.

JProffitt71:
The regex pattern in RestRouteCollection::all() was including the whitespace and the comment, causing it to fail against all route names (thinking they are all custom). This broke the sorting that allowed routes like `get_user` to be defined before routes like `new_user` and `remove_user` without breaking them. 

This bug was introduced between 1.8 and 2.0, when the pattern was pulled out of the foreach loop.

### DIFF
--- a/Routing/RestRouteCollection.php
+++ b/Routing/RestRouteCollection.php
@@ -76,9 +76,9 @@ class RestRouteCollection extends RouteCollection
      */
     public function all()
     {
-        $regex = '/' .
-            '(_|^)' .
-            '(get|post|put|delete|patch|head|options|mkcol|propfind|proppatch|lock|unlock|move|copy|link|unlink)_' . // allowed http methods
+        $regex = '/'.
+            '(_|^)'.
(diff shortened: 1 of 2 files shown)


lsmith77:
can you add a test please?

JProffitt71:
Forgive me if I'm being dense, but shouldn't the testCustomActionRoutesOrder() method in Routing/Loader/RestRouteLoaderTest.php already test this? For that matter, shouldn't that be asserting already? I have not worked with tests yet, so my knowledge on them is limited.

Actually looking at that test, it looks like it is backwards. It will only fail if get_bar is dumped after new_bar (the final desired behavior we actually want).

JProffitt71:
My apologies, I misunderstood how PHPUnit worked. I learned more about that, and figured out why unit tests were not catching this issue.

The OrderController fixture had the actions defined in the correct order (`['new_foos', 'get_foos', 'new_bars', 'get_bars_custom', 'get_bars']`), which would pass unit tests even if custom route sorting wasn't working. I swapped `new_foos` and `get_foos` so it would properly test this issue.

PR #1647, opened 2017-01-09
add Tests to autoload-dev

IndraGunawan:
add Tests directory to autoload-dev

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,6 +27,11 @@
             "Tests/"
         ]
     },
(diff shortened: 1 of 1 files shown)


xabbuh:
Thank you for this PR. However, this change is not necessary as it is already covered by the `autoload` section. Did you experience any issue this would solve?

IndraGunawan:
sorry, i'm not double-check about this. :)

PR #1651, opened 2017-01-15
Document correct order with JMSSerializerBundle

### LINKED ISSUE #1600: `jms_serializer.form_error_handler.class` parameter from `serializer.xml` is not used
In `Resources/config/serializer.xml` of this bundle, the parameter `jms_serializer.form_error_handler.class` is overridden to `FOS\RestBundle\Serializer\Normalizer\FormErrorHandler`. However, when I test with returning a form with errors from a controller, instead the serialized output from the default `JMS\Serializer\Handler\FormErrorHandler` class is seen in the HTTP response body. It seems the override of the parameter in `serializer.xml` has no effect.

I can only get the desired result by putting this in my app's `parameters.yml`:

    parameters:
        jms_serializer.form_error_handler.class: FOS\RestBundle\Serializer\Normalizer\FormErrorHandler

After I define the above parameter, I get the desired result, e.g. I see the form errors are serialized by `FOS\RestBundle\Serializer\Normalizer\FormErrorHandler` with `code`, `message` and `errors` keys, as seen in the [docs](https://symfony.com/doc/master/bundles/FOSRestBundle/2-the-view-layer.html#forms-and-views).

Am I missing something or is this a bug? Any help is appreciated. Thanks! :)

Note: I'm using version `2.1.0` of `friendsofsymfony/rest-bundle`, and version `1.1.0` of `jms/serializer-bundle`.

ddeboer:
Fix #1600.

### DIFF
--- a/Resources/doc/1-setting_up_the_bundle.rst
+++ b/Resources/doc/1-setting_up_the_bundle.rst
@@ -27,10 +27,30 @@ file of your project:
     {
         public function registerBundles()
         {
-            $bundles = array(
+            $bundles = [
                 // ...
                 new FOS\RestBundle\FOSRestBundle(),
-            );
+            ];
+
+            // ...
+        }
+    }
+    
+If you use JMSSerializerBundle (optional), make sure to include it *before*
+FOSRestBundle:
+   
+.. code-block:: php
+
+    // app/AppKernel.php
+    class AppKernel extends Kernel
+    {
+        public function registerBundles()
+        {
+            $bundles = [
+                // ...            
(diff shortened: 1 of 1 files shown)


xabbuh:
What do you think if we set the parameter in a compiler pass instead to completely prevent those WTF moments?

ddeboer:
@xabbuh Even better, of course!

dbu:
i am merging this as its at least a help. if somebody wants to provide a compiler pass to make things more reliable, please do...

Issue #1652, opened 2017-01-19
Non-existing template engine gives DI error


c410-f3r:
Since [this PR](https://github.com/symfony/symfony/pull/19120), templating can be disabled with the following configuration:

```
framework:
    templating: false
```
 
Using this causes the following error to arise:

```
[Symfony\Component\DependencyInjection\Exception\InvalidArgumentException]          
Unable to replace alias "fos_rest.templating" with actual definition "templating". 

[Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]  
You have requested a non-existent service "templating".
```

GuilhemN:
I thought https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1655 fixed this but in fact it does not, we can't easily fix it as the `ViewHandler` depends on the Templating component and we would need to adapt it to make it work without it.
Feel free to submit a PR if you're willing to do it :)

sogetimaitral:
Hello,

New user here (on both symfony and FOS Rest Bundle).
Should I understand that we need to keep and get twig running to be able to run FOS Rest Bundle?
We wanted to disable it to avoid using it, since we only want a REST API using JSON only...
Thanks for the help.

xabbuh:
We should solve the issue somehow. But we need to deeper think about a proper solution.

c410-f3r:
@GuilhemN Thanks for the initiative for trying to fix the problem. I would like the send a PR but right now I don't have time and I don't know how FOSRestBundle works, maybe in the near future.
@sogetimaitral Twig is useful for development when using the web profiler.

sogetimaitral:
Interesting, Twig is used by the profiler front-end. I note that, thanks for the hint :)

GuilhemN:
see https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1675

c410-f3r:
Used the same configuration with latest commit (66b66dd5f2589f4c4453a1734f0e765aa3502fc0) and the problem still persist.

Tried several configurations:

```
framework:
    templating: false
```

```
framework:
    templating:
        enabled: false
```

```
framework:
    templating:
        enabled: false
        engines: []
```

```
framework:
    templating:
        enabled: false
        engines: ['php']
```

GuilhemN:
@c410-f3r https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1684 should be the last fix...

PR #1655, opened 2017-01-20
Make aliases private

GuilhemN:
Fix https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1652 by making aliases private (this way they are automatically removed).


xabbuh:
Should we really apply this latch so broadly? It could break applications where people rely on the aliases being public.

GuilhemN:
I don't think people depend on this aliases,  they're not documented anywhere nor useful for the end user (why would someone use `fos_rest.validator` instead of `validator`?).

xabbuh:
I am not sure. They shouldn't, but that doesn't it's not happening. But we could agree on this being an unsupported use case and make the changes you suggested nonetheless.

GuilhemN:
@xabbuh imo this is not worth a bc layer :) No one should rely on that.

xabbuh:
:+1: let's do it then :)

GuilhemN:
Ok, thanks ☺

GuilhemN:
Well after rethinking about it we should just revert this PR, it did only bring unwanted effects without solving https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1652. Sorry for that...

Issue #1661, opened 2017-02-03
Unexpected Content-Type on Unrecognized Extensions w/ prefer_extension


JProffitt71:
I'm having an issue with counter-intuitive content-types for requests that have unrecognized extensions. Here is the minimal configuration to create this issue:

```
fos_rest:
    format_listener:
        rules:
            - { path: '^/', priorities: ['html', 'json', 'xml'], prefer_extension: true}
```

And a sample problematic request:
Location: /js/main.min.js.map
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8

The format negotiator cannot resolve .map to a mime-type, so it just gives the last mime-type a q of 2.0:
text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8; q=2.0

This causes it to weight anything not listed in the Accept header higher, and in this case it returns a Content-Type of application/json (the first format in my priorities thats not listed in Accept header).

This is extremely counter-intuitive. I would expect when it couldn't determine the mime-type from the extension, for it to not modify the Accept header at all and do content negotiation like prefer_extension was off.

PR #1668, opened 2017-02-15
Fixed prefer_extension content negotiation

JProffitt71:
Fixes issue #1661 

When prefer_extension is set, the Accept header will be prepended to instead of appended to. This ensures that the Accept header cannot override the qvalue of the extension mime-type, like in this scenario:

```
Location: http://example.com/file.json
Accept: text/html, application/json; q=0.1
```

If no mime-type can be found for the extension, the Accept header will not be modified.

Added unit tests for prefer_extension to ensure content negotiation remains predictable.

### DIFF
--- a/Negotiation/FormatNegotiator.php
+++ b/Negotiation/FormatNegotiator.php
@@ -77,7 +77,7 @@ class FormatNegotiator extends BaseNegotiator
                     $extensionHeader = $request->getMimeType($extension);
 
                     if ($extensionHeader) {
-                        $header = $extensionHeader.'; q='.$options['prefer_extension'].','.$header;
+                        $header = $extensionHeader.'; q='.$options['prefer_extension'].($header ? ','.$header : '');
                     }
                 }
             }
--- a/Negotiation/FormatNegotiator.php
+++ b/Negotiation/FormatNegotiator.php
@@ -75,10 +75,10 @@ class FormatNegotiator extends BaseNegotiator
                 if (!empty($extension)) {
                     // $extensionHeader will now be either a non empty string or an empty string
                     $extensionHeader = $request->getMimeType($extension);
-                    if ($header && $extensionHeader) {
-                        $header .= ',';
+
+                    if ($extensionHeader) {
+                        $header = $extensionHeader.'; q='.$options['prefer_extension'].','.$header;
                     }
-                    $header .= $extensionHeader.'; q='.$options['prefer_extension'];
                 }
             }
 
--- a/Tests/Negotiation/FormatNegotiatorTest.php
+++ b/Tests/Negotiation/FormatNegotiatorTest.php
@@ -109,18 +109,18 @@ class FormatNegotiatorTest extends \PHPUnit_Framework_TestCase
         $this->addRequestMatcher(true, ['priorities' => $priorities, 'prefer_extension' => '2.0']);
 
         $reflectionClass = new \ReflectionClass(get_class($this->request));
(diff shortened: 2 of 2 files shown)


Tests/Negotiation/FormatNegotiatorTest.php
```diff
@@ -103,6 +103,45 @@ public function testGetBestWithFormat()
         $this->assertEquals(new Accept('application/json;version=1.0'), $this->negotiator->getBest(''));
     }
 
+    public function testGetBestWithPreferExtension()
+    {
+        $priorities = ['text/html', 'application/json'];
+        $this->addRequestMatcher(true, ['priorities' => $priorities, 'prefer_extension' => '2.0']);
+
+        $reflectionClass = new \ReflectionClass(get_class($this->request));
```
GuilhemN: Why not using the `X_ORIGINAL_URL` header?

Negotiation/FormatNegotiator.php
```diff
@@ -75,10 +75,10 @@ public function getBest($header, array $priorities = [])
                 if (!empty($extension)) {
                     // $extensionHeader will now be either a non empty string or an empty string
                     $extensionHeader = $request->getMimeType($extension);
-                    if ($header && $extensionHeader) {
-                        $header .= ',';
+
+                    if ($extensionHeader) {
+                        $header = $extensionHeader.'; q='.$options['prefer_extension'].','.$header;
```
GuilhemN: what if `$header` is empty? is `foo;q=10,` valid?

Tests/Negotiation/FormatNegotiatorTest.php
```diff
@@ -103,6 +103,45 @@ public function testGetBestWithFormat()
...
+        $reflectionProperty->setValue($this->request, 'http://example.com/file.json');
+
+        // Without extension mime-type in Accept header
+
+        $this->request->headers->set('Accept', 'text/html');
+        $this->assertEquals(new Accept('application/json'), $this->negotiator->getBest(''));
+
+        // With low q extension mime-type in Accept header
+
+        $this->request->headers->set('Accept', 'text/html, application/json; q=0.1');
```
GuilhemN: Does not look very useful.

Tests/Negotiation/FormatNegotiatorTest.php
JProffitt71: I wanted to do that, but Request->getRequestUri() caches the result, so any tests afterwards would erroneously use the first request uri set. Actually it should be using "pathInfo" because getPathInfo also caches its result and that is what is used by FormatNegotiator. I'll push an update to fix that.

Tests/Negotiation/FormatNegotiatorTest.php
JProffitt71: This test caught an issue before I made my modifications. Prior to my modification of FormatNegotiator, the mime-type in the Accept header would override the one prefer_extension appends. So if you requested `file.json` with the header in that test, the final header would be:

`Accept: text/html, application/json; q=0.1, application/json; q=2.0`

It would ignore the second instance of application/json and prefer text/html, failing that test.

This test guards against the case where a contributor refactors FormatNegotiator to append the mime-type again. However if there are other guards against that, I agree we can scrap this test.

Negotiation/FormatNegotiator.php
JProffitt71: Good point, this has been fixed.

Issue #1659, opened 2017-01-30
Using body_listener is conflicting with the symfony user argument resolver, resulting in a 400


norbert-yoimo:
If you have the body_listener, enabled, and try to use the [$user argument resolver](http://symfony.com/blog/new-in-symfony-3-2-user-value-resolver-for-controllers) in symfony, like this:

```php
namespace MyBundle\Controller;
use FOS\RestBundle\Controller\FOSRestController;
use FOS\RestBundle\Controller\Annotations\View as FOSView;
use FOS\RestBundle\Controller\Annotations\Get;
use Symfony\Component\Security\Core\User\UserInterface;

class MyController extends FOSRestController {

    /**
     * @FOSView
     * @Get("/team/X")
     */
    public function getXAction(UserInterface $user) {
    }
}
```

You will always get a `400 Bad Request` response from this route, with the body:

```
{
    "code": 400,
    "message": "Could not decode JSON, syntax error - malformed JSON."
}
```

Looking into the stacktrace, there is a call to:

```
at Serializer ->deserialize ('', 'Symfony\\Component\\Security\\Core\\User\\UserInterface', 'json', object(DeserializationContext)) 
in vendor/friendsofsymfony/rest-bundle/Serializer/JMSSerializerAdapter.php at line 60
```

Which makes me believe that the FOSRestBundle body_listener is trying to deserialize an empty string as the user object, instead of letting the symfony argument resolver take care of it.

GuilhemN:
Can you check that https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1669 fixes your issue?

norbert-yoimo:
Sorry I haven't yet had time to take a look, I'm still keeping it on my task list

Preclowski:
Problem still exists

anonim1133:
It's still pain in the ass. #1669 fixes it, but does not exist in newer branches.

xabbuh:
fixed in #1669

PR #1669, opened 2017-02-17
[RequestBodyParamConverter] Force being explicit

GuilhemN:
Fixes https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1659

When a parameter is not a class registered in doctrine, the `RequestBodyParamConverter` can be reached without being explicitly asked which is not expected (imo at least).

### DIFF
--- a/Request/RequestBodyParamConverter.php
+++ b/Request/RequestBodyParamConverter.php
@@ -124,7 +124,7 @@ class RequestBodyParamConverter implements ParamConverterInterface
      */
     public function supports(ParamConverter $configuration)
     {
-        return null !== $configuration->getClass();
+        return null !== $configuration->getClass() && 'fos_rest.request_body' === $configuration->getConverter();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


Request/RequestBodyParamConverter.php
```diff
@@ -124,7 +124,7 @@ public function apply(Request $request, ParamConverter $configuration)
      */
     public function supports(ParamConverter $configuration)
     {
-        return null !== $configuration->getClass();
+        return null !== $configuration->getClass() && null !== $configuration->getConverter();
```
xabbuh: Can't we also check the returned value?

GuilhemN: Sure, done.

Preclowski:
Hey guys, problem did back. Argument resolver does some problems with fos rest. This fix is solution, but it was reverted in 2.2 (or earlier).

PR #1670, opened 2017-02-18
View::setSerializationContext inexistant

clavier-souris:
Replaced View::setSerializationContext call in example with View::setContext as previous method does not exist anymore


GuilhemN:
Imo using `setContext` should be discouraged (maybe we should deprecate it?): I think we should instead promote using `getContext`.

clavier-souris:
This is just about documentation.
I have been mistaken by this error in a Symfony 3 project.

GuilhemN:
Thank you @clavier-souris !

PR #1672, opened 2017-02-23
Bad mime type

sebastianlp:
This is an image so the correct mime type would be "image/jpeg" or "image/something".


GuilhemN:
That's true, thanks!

PR #1676, opened 2017-02-28
argument "map" has changed to "array"

carlcasbolt:


GuilhemN:
That's the opposite: now `map` is used. https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Controller/Annotations/AbstractScalarParam.php

carlcasbolt:
OK, honest mistake, I must have an old version of the code (I started using it last year). I'll see if I can upgrade my version and use `map=true`

greg0ire:
lol, I should have checked before approving

PR #1679, opened 2017-03-10
Update versioning.rst

vpassapera:
Clarify mime_type ordering actually matters since the VersionListener takes the first result matching.

e.g.

```yaml
fos_rest:
    param_fetcher_listener: true
    body_listener: true
    format_listener:
        rules:
            - { path: ^/, priorities: [ 'html', 'json', 'application/javascript', 'text/css', '*/*'], fallback_format: html, prefer_extension: true }
    body_converter:
        enabled: true
        validate: true
    view:
        mime_types:
            json:
              - 'application/json;version=1.0'
              - 'application/json'
```

If application/json is listed first, and you're using the accept header as your versioning strategy, it will not work.

### DIFF
--- a/Resources/doc/versioning.rst
+++ b/Resources/doc/versioning.rst
@@ -94,7 +94,9 @@ See :doc:`Format Listener <format_listener>`
     fos_rest:
         view:
             mime_types:
-                json: ['application/json', 'application/json;version=1.0', 'application/json;version=1.1']
+                json: ['application/json;version=1.0', 'application/json;version=1.1', 'application/json']
+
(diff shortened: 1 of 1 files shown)


vpassapera:
Noticed this is actually a bug within the request object so I will open a PR against that.

There's a line that's hardcoding the top element in the config and returning that always.

```php
        return isset(static::$formats[$format]) ? static::$formats[$format][0] : null;
```

vpassapera:
Opened a PR against http foundation.

IF this is incorrect and there's another way of addressing this issue, I would be happy to take some direction.

https://github.com/symfony/http-foundation/pull/34

PR #1681, opened 2017-03-14
Add file line and trace info to exception

phoenixgao:
Add file line and trace info to jms exception serializer if debug is enabled

### DIFF
--- a/Serializer/Normalizer/AbstractExceptionNormalizer.php
+++ b/Serializer/Normalizer/AbstractExceptionNormalizer.php
@@ -29,7 +29,7 @@ class AbstractExceptionNormalizer
     /**
      * @var bool
      */
-    private $debug;
+    protected $debug;
 
     /**
      * @param array $messagesMap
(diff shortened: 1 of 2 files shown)


phoenixgao:
@GuilhemN It's quite easy to write a custom exception normalizer for symfony serializer like in the documentation, but it's so difficult to have one for jms serializer, since from 2.0 the exception wrapper handler was deprecated, and jms exception handler service depends on other private service.

Is it possible to have some examples in the documentation that how to customize jms exception handler? then I would be glad to close this PR :p

GuilhemN:
An example for json could be:
```php
use JMS\Serializer\Handler\SubscribingHandlerInterface;
use JMS\Serializer\GraphNavigator;
use JMS\Serializer\JsonSerializationVisitor;
use JMS\Serializer\Context;

class MyExceptionHandler implements SubscribingHandlerInterface
{
    public static function getSubscribingMethods()
    {
        return array(
            array(
                'direction' => GraphNavigator::DIRECTION_SERIALIZATION,
                'format' => 'json',
                'type' => 'App\Exception\MyException',
                'method' => 'serializeMyExceptionToJson',
            ),
        );
    }

    public function serializeMyExceptionToJson(JsonSerializationVisitor $visitor, App\Exception\MyException $exception, array $type, Context $context)
    {
        return $visitor->visitArray([
            'message' => $exception->getMessage(),
        ]);
    }
}
```

xabbuh:

Issue #1538, opened 2016-07-09
Symfony\Component\Security\Http\Firewall\ExceptionListener overrides AccessDeniedListener


bblue:
The logic in the `AccessDeniedListener` is effectively overriden by `Symfony\Component\Security\Http\Firewall\ExceptionListener`

The priority of the Symfony listener is lower than that of FOSRest, thus the following method in Symfony's listener overrides the `$event->setException($e);`

```
private function handleAuthenticationException(GetResponseForExceptionEvent $event, AuthenticationException $exception)
{
    if (null !== $this->logger) {
        $this->logger->info('An AuthenticationException was thrown; redirecting to authentication entry point.', array('exception' => $exception));
    }

    try {
        $event->setResponse($this->startAuthentication($event->getRequest(), $exception));
    } catch (\Exception $e) {
        $event->setException($e);
    }
}
```

This will trigger a redirect to the login page, and in some cases (my case) this will trigger yet another authentication exception (due to faulty JWT token for instance), but this new exception will be caught in the above method, the `$event->exception` will be overridden, and hence no conversion to the http exceptions are done. I will still get JSON data, but it will come with a 500 error.

geoffreytran:
Running into this same issue here

pppdns:
Same here, although I get HTML data (coming from [RedirectResponse#L85](https://github.com/symfony/http-foundation/blob/2c8a1e1/RedirectResponse.php#L85)))

lsmith77:
FYI I created a PR but I am not sure yet what the side effects will be https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1692

scott-r-lindsey:
Any movement on this?

bblue:
Just realized I had been referenced in this thread a while ago. Haven't worked on that project for a loooong time, and its dependencies and symfony has not been updated since 2017. 

I recall initally making a compiler pass to override the FOSRest listener, but was not happy with that solution. I ended up instead solving the root cause that threw my initial exception (not linked to fosrest), thus I really just side-steped the real bug, but I then would not have to override any FOSRest classes.

So to summarize: you can probably avoid a particular use-case error with a compiler pass, but it would be fragile. Unfortunately I have deleted my compiler pass object, so I can't tell you what I actually wrote in it... I probably just rethrew the exception.

I'm guessing the PR by @lsmith77 is the way to go, but have not looked at it.

xabbuh:
[...]

PR #1692, opened 2017-04-06
fix listener priorities for the access denied listener and the exception listener

lsmith77:
@xabbuh why did you set the listener priority to `-100` https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1410/files#diff-324b1abf8a181b12c840bf5baefb86f3R50 ?

fixes https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1538

### DIFF
--- a/EventListener/ExceptionListener.php
+++ b/EventListener/ExceptionListener.php
@@ -47,7 +47,7 @@ class ExceptionListener extends HttpKernelExceptionListener
     public static function getSubscribedEvents()
     {
         return array(
-            KernelEvents::EXCEPTION => array('onKernelException', -100),
+            KernelEvents::EXCEPTION => array('onKernelException', 3),
         );
     }
 


xabbuh:
@lsmith77 I think there was no special reason for this choice (the only important thing was to let the FOSRestBundle listener have a higher priority than the one from Symfony core which has a priority of -128).

Maybe @jovobe or @GuilhemN remember more details.

jovobe:
Sorry, I don't remember the context.

xabbuh:
I guess making this change should be safe. But we should add a test that enforces the listener to be registered with this priority.

lsmith77:
I was also pondering if we should in fact disable the relevant core listener and extend it .. rather than relying on priorities. It increases the coupling .. but in a way the coupling exists already but is left implicit causing the issues to begin with.

xabbuh:
We used to disable the core listener until #1410 and changed it there to let the core listener step in outside of rest zones. Do we really want to change that again?

xabbuh:
What we could maybe do is to decorate the core listener and forward the exception handling to the decorated listener in case we are outside of rest zones. This way we would also not have to deal with priorities at all.

lsmith77:
@xabbuh decorate is fine for me too

lsmith77:
haven't found time to work on a decorator :-/

darkbluesun:
What's preventing this being done?

lsmith77:
rather than using priorities, we wanted to look into decorating the core listener (i guess the exception listener). i am just not finding the time to work on this. but this PR works for me in a project I am about to wrap up. but its hard to know for sure that it does not have unwanted side effects.

lsmith77:
@darkbluesun do you have time to work on this?

darkbluesun:
Sorry @lsmith77 I don't. I had thought that the PR was helpful having tried it, but it had some unfortunate side effects. What you suggest is probably better, but I don't have the time or expertise for it, sorry.

zerkms:
Btw, this PR does not seem to change the behaviour: the `Symfony\Component\Security\Http\Firewall\ExceptionListener` listener still handles the exception. And the FosRestBundle's ExceptionListener is not even registered.

zerkms:
My apologies, this PR totally works, thanks :-)

Any chance to merge it any time soon?

unglud:
Any plans to merge it? 'cuz I need to change priority from -100 to 3 in `FOS\RestBundle\EventListener\ExceptionListener` to be able everything work.

xabbuh:
see https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1692#issuecomment-304780693

lsmith77:
@xabbuh just stumbled over this one again. I guess https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1692#issuecomment-293210594 is still the way to go.

xabbuh:
Closing here in favour of #2109 and #2161

Issue #1695, opened 2017-04-12
Non `ClassResourceInterface` controller re-uses route name for `cget` and `get` actions.


kherge:
I am currently in the process of creating a re-usable controller for handling route actions. Unfortunately I am experience a strange issue with how `cgetAction` and `getAction` route names are being handled.

## Problem

When generating route names for `cgetAction` and `getAction` without implementing `ClassResourceInterface`, the route names conflict and the last route definition wins.

## Solution

Instead of using `get` as the name of the route for `cgetAction`, it should be changed to `list` (or something like it). Depending on the implementation, this may introduce a backwards compatibility break.

## Proposal

This change should be gated behind a new bundle setting in order to prevent a backwards compatibility break. For example, a boolean setting called `fos_rest.routing_loader.cget_list` controls whether the [`RestActionReader`](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/d62a6c0f4bc699f899865d7e7bc7a4186aef9a86/Routing/Loader/Reader/RestActionReader.php#L521-L538) class uses `get` or `list`. By default, this setting will be set to `false`. This will allow all existing configurations to continue working as before without any configuration changes, but also allow the new route name to be used when desired.

## Notes

- [Route name generator.](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/d62a6c0f4bc699f899865d7e7bc7a4186aef9a86/Routing/Loader/Reader/RestActionReader.php#L521-L538)
- [Route name prefixing.](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/d62a6c0f4bc699f899865d7e7bc7a4186aef9a86/Routing/Loader/Reader/RestActionReader.php#L278)

kherge:
I created a pull request that implements the proposed solution.

kherge:
Abandoning in favor of using a custom route loader.

PR #1696, opened 2017-04-12
Supporting "list" as alternative route name for "cgetAction".

kherge:
See #1695.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -102,6 +102,7 @@ final class Configuration implements ConfigurationInterface
                 ->arrayNode('routing_loader')
                     ->addDefaultsIfNotSet()
                     ->children()
+                        ->booleanNode('cget_list')->defaultFalse()->end()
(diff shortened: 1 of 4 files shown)


kherge:
I've also considered making the actual `list` prefix configurable through the `fos_rest.routing.cget_list` setting. Maybe rename the setting to `fos_rest.routing.cget` and default it to `cget`?

I'm open to suggestions.

kherge:
Hi @GuilhemN, please let me know if there is anything I can do to help move this along!

lsmith77:
Generally the idea would be to have `get_users` vs. `get_user` for such route names, but not all subjects will nicely pluralize.

So this sounds like a sensible solution to me.
Maybe we should even deprecate the old behavior?

kherge:
I believe that the original behavior should be deprecated.

The original behavior would break route generation when used without `ClassResourceInterface`.

kherge:
Abandoning in favor of using a custom route loader.

PR #1701, opened 2017-05-08
Use JMS Serializer Context Factories to create new contextes

goetas:
As of jms/serializer 1.4.0 (and jms/serializer-bundle 1.2.0) is possible to configure the default context via services ([doc](http://jmsyst.com/libs/serializer/master/configuration#setting-a-default-serializationcontext-factory)).

This PR improves the integration with FOS Rest and JMS serializer

(I have also added some extra tests for the `JMSSerializerAdapter` class)

### DIFF
--- a/Serializer/JMSSerializerAdapter.php
+++ b/Serializer/JMSSerializerAdapter.php
@@ -43,8 +43,7 @@ class JMSSerializerAdapter implements Serializer
         SerializerInterface $serializer,
         SerializationContextFactoryInterface $serializationContextFactory = null,
         DeserializationContextFactoryInterface $deserializationContextFactory = null
-    )
-    {
+    ) {
         $this->serializer = $serializer;
         $this->serializationContextFactory = $serializationContextFactory;
         $this->deserializationContextFactory = $deserializationContextFactory;
--- a/Serializer/JMSSerializerAdapter.php
+++ b/Serializer/JMSSerializerAdapter.php
@@ -13,8 +13,10 @@ namespace FOS\RestBundle\Serializer;
 
 use FOS\RestBundle\Context\Context;
 use JMS\Serializer\Context as JMSContext;
-use JMS\Serializer\DeserializationContext as JMSDeserializationContext;
-use JMS\Serializer\SerializationContext as JMSSerializationContext;
+use JMS\Serializer\ContextFactory\DeserializationContextFactoryInterface;
+use JMS\Serializer\ContextFactory\SerializationContextFactoryInterface;
+use JMS\Serializer\DeserializationContext;
+use JMS\Serializer\SerializationContext;
 use JMS\Serializer\SerializerInterface;
 
 /**
@@ -34,10 +36,24 @@ class JMSSerializerAdapter implements Serializer
     const DESERIALIZATION = 1;
 
     private $serializer;
+    /**
+     * @var SerializationContextFactoryInterface
+     */
+    private $serializationContextFactory;
+    /**
+     * @var DeserializationContextFactoryInterface
+     */
+    private $deserializationContextFactory;
 
-    public function __construct(SerializerInterface $serializer)
+    public function __construct(
+        SerializerInterface $serializer,
+        SerializationContextFactoryInterface $serializationContextFactory = null,
+        DeserializationContextFactoryInterface $deserializationContextFactory = null
+    )
     {
         $this->serializer = $serializer;
+        $this->serializationContextFactory = $serializationContextFactory;
+        $this->deserializationContextFactory = $deserializationContextFactory;
     }
 
     /**
@@ -69,9 +85,13 @@ class JMSSerializerAdapter implements Serializer
     private function convertContext(Context $context, $direction)
     {
         if ($direction === self::SERIALIZATION) {
-            $jmsContext = JMSSerializationContext::create();
+            $jmsContext = $this->serializationContextFactory
+                ? $this->serializationContextFactory->createSerializationContext()
+                : SerializationContext::create();
         } else {
-            $jmsContext = JMSDeserializationContext::create();
+            $jmsContext = $this->deserializationContextFactory
+                ? $this->deserializationContextFactory->createDeserializationContext()
+                : DeserializationContext::create();
             $maxDepth = $context->getMaxDepth(false);
             if (null !== $maxDepth) {
                 for ($i = 0; $i < $maxDepth; ++$i) {
@@ -97,7 +117,7 @@ class JMSSerializerAdapter implements Serializer
             $jmsContext->setSerializeNull($context->getSerializeNull());
         }
 
-        foreach($context->getExclusionStrategies() as $strategy) {
+        foreach ($context->getExclusionStrategies() as $strategy) {
             $jmsContext->addExclusionStrategy($strategy);
         }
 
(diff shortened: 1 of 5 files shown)


Serializer/JMSSerializerAdapter.php
```diff
@@ -34,10 +34,24 @@ class JMSSerializerAdapter implements Serializer
     const DESERIALIZATION = 1;
 
     private $serializer;
+    /**
+     * @var SerializationContextFactoryInterface
+     */
```
xabbuh: the class property docblocks are not needed

goetas: do you want me to remove them or is it ok as it is?

xabbuh: It's not a blocker for me. But I would prefer to remove them.

xabbuh:
Oh, can you add an entry to the changelog?

goetas:
> Oh, can you add an entry to the changelog?

there is no section for the 2.0 branch in https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/CHANGELOG.md and the file looks almost one year old.

Serializer/JMSSerializerAdapter.php
goetas: done

goetas:
The commit https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1701/commits/dbdf1f93296ee9c1732c0127669e59b63dfcab08 is becasue today I have released the JMS serializer bundle 2.0. The new version is almost 100% backward compatible with the 1.x, so no changes are required from the user point of view.
The changes are mainly about changing some defaults to optimize performance and DX, [here a full list](https://github.com/schmittjoh/JMSSerializerBundle/blob/master/UPGRADING.md)

Serializer/JMSSerializerAdapter.php
```diff
@@ -34,10 +34,18 @@ class JMSSerializerAdapter implements Serializer
     const DESERIALIZATION = 1;
 
     private $serializer;
+    private $serializationContextFactory;
+    private $deserializationContextFactory;
 
-    public function __construct(SerializerInterface $serializer)
+    public function __construct(
+        SerializerInterface $serializer,
+        SerializationContextFactoryInterface $serializationContextFactory,
```
GuilhemN: Not sure this bc break is acceptable in a patch, should we release this in 2.3?

goetas:
@GuilhemN 
>  It sounds a bit weird to me to have context factories but why not

there are some usecases where users want to specify at application level if serialize nulls or enable max-depth... so IMO make sense to have a place with per-app configs

> I'm not sure what to do about the bc break.

Imo this is not  bug fix so should not go with 2.2.something, but more with 2.3.0... but I do not know if you have plans to release it soon..

GuilhemN:
> there are some usecases where users want to specify at application level if serialize nulls or enable max-depth... so IMO make sense to have a place with per-app configs

Sure but imo this should be configured directly on the serializer not by forcing to instantiate contexts through a factory all the time. Anyway this is not the place to discuss your choices :)

> Imo this is not bug fix so should not go with 2.2.something, but more with 2.3.0... 

I agree

> Imo this is not bug fix so should not go with 2.2.something, but more with 2.3.0... but I do not know if you have plans to release it soon..

That's the problem, it depends how much PRs we get, we try to release minors only when we have enough changes. I'll submit some changes soon, if they're big enough then we'll probably release a minor version soon.

goetas:
> That's the problem, it depends how much PRs we get, we try to release minors only when we have enough change

Do not see the problem. Releases (except of major) are cheap. Some projects release minor and bugfix releases even daily (if there are changes to release obviously).

Releasing often allows to ship earlier new features and bugfixes... by semver, updating from 1.x to 1.y, from the user point of is compatible and safe.

Anyway, is not a big deal waiting for some time.
The main advantage of the serializer-bundle 2.0 are some doctrine related performance improvements that are enabled by default, while in 1.5 is necessary to update them manually. 

Said that, I will agree with whatever decision will be taken :smile:

Serializer/JMSSerializerAdapter.php
xabbuh: Oh indeed, I missed that. We could default the argument to `null` and fall back to the old code then, right?

goetas: done

Tests/DependencyInjection/FOSRestExtensionTest.php
```diff
@@ -358,6 +358,20 @@ public function testConfigLoad()
         );
     }
 
+    public function testJMSAdapterSerializerServiceDefinition()
```
GuilhemN: Doesn't seem very useful to me. It can be removed imo.

Tests/Serializer/JMSSerializerAdapterTest.php
```diff
@@ -0,0 +1,114 @@
...
+        $this->serializer->expects($this->once())->method('serialize')->with('foo', 'json', $jmsContext);
+
+        $adapter->serialize('foo', 'json', new Context());
+    }
+
+    public function testBasicDeSerializeAdapterWithoutContextFactories()
+    {
+        $jmsContext = DeserializationContext::create();
+        $adapter = new JMSSerializerAdapter($this->serializer);
+        $this->serializer->expects($this->once())->method('deserialize')->with('foo', 'json', $jmsContext);
```
GuilhemN: missing `'string'` argument here

goetas: hmmm... curios why this test was green then... :confused:

goetas: ah, it did not https://travis-ci.org/FriendsOfSymfony/FOSRestBundle/jobs/235667634 :confused:  :confused:  :confused:  :confused:

goetas:
@GuilhemN  rebased and modified as asked by you

Serializer/JMSSerializerAdapter.php
```diff
@@ -34,10 +36,18 @@ class JMSSerializerAdapter implements Serializer
...
     private $serializer;
+    private $serializationContextFactory;
+    private $deserializationContextFactory;
 
-    public function __construct(SerializerInterface $serializer)
+    public function __construct(
+        SerializerInterface $serializer,
+        SerializationContextFactoryInterface $serializationContextFactory = null,
+        DeserializationContextFactoryInterface $deserializationContextFactory = null
+    )
```
Aliance: > When the argument list is split across multiple lines, the closing parenthesis and opening brace MUST be placed together on their own line with one space between them.

http://www.php-fig.org/psr/psr-2/#method-arguments

goetas: fixed, thanks

goetas:
@xabbuh  @GuilhemN `"minimum-stability": "dev"` in composer.json is a bad idea IMO... it downloads unstable packages when not necessary....
As i can understand, was introduced to allow sf 4.0, why not set it "at-runtime" in travis.yml?

goetas:
As of 5.7.2, phpunit will clone all objects provided by a data provider (introduced with https://github.com/sebastianbergmann/phpunit/commit/55d21483aad5cc0cfb63075b53f4cc5169ec855d)

This will cause a call to`Symfony\Component\Form\Form::__clone` mock, where `$this->children` is null and will throw `Error: __clone method called on non-object`

lsmith77:
can we merge this? looks ok from a quick inspection.

GuilhemN:
Thank you @goetas !

goetas:
thanks to you! :)

PR #1706, opened 2017-05-15
Actualize RouteRedirectView class name.

Aliance:
There is no `RouteRedirect` class, `RouteRedirectView` is correct class name.


GuilhemN:
Thank you @Aliance!

PR #1707, opened 2017-05-17
Remove redundant unset.

Aliance:
`unset($v['enabled']);` is redundant here, because key `enabled` is overriden at the next line.

https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1622#discussion_r91139867

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -62,7 +62,6 @@ final class Configuration implements ConfigurationInterface
                     ->beforeNormalization()
                         ->ifArray()->then(function ($v) {
                             if (!empty($v) && empty($v['formats'])) {
-                                unset($v['enabled']);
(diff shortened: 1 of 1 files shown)


Aliance:
Any ideas why the `enable` key was missed before?
Should I just add it to expected array?

GuilhemN:
Actually it's not redundant, it's here to support

Aliance:
Can't understand you idea. As I saw, there are two steps:
1. unset key `enabled` from array `$v`*
2. set array `$v` to `['enabled' => ...]`

So the first step seems to be redundant, because of second step always add `enabled` key. Where am I wrong?

Close pr?

GuilhemN:
The second step reuse the original `$v` (`$v = ['formats' => $v, ...];`, so `$v['enabled']` must be removed.
This should indeed be closed, thanks!

PR #1708, opened 2017-05-18
allow Symfony 4.0 requirements

xabbuh:


GuilhemN:
Shouldn't we run tests on it?

xabbuh:
done

composer.json
```diff
@@ -30,40 +30,41 @@
...
+        "symfony/twig-bundle": "^2.7|^3.0|^4.0",
+        "symfony/browser-kit": "^2.7|^3.0|^4.0",
+        "symfony/dependency-injection": "^2.7|^3.0|^4.0",
+        "symfony/expression-language": "~2.7|^3.0|^4.0",
+        "symfony/css-selector": "^2.7|^3.0|^4.0",
         "phpoption/phpoption": "^1.1",
         "jms/serializer-bundle": "^1.0",
         "psr/http-message": "^1.0"
     },
+    "minimum-stability": "dev",
```
GuilhemN: Can't we just require `4.0.*@dev` instead ?

xabbuh: We can if you prefer that. :)

GuilhemN: It's fine for me as long as we don't have problems, just tests are more likely to break with this setting.
Let's keep it this way for now, we'll see soon enough if it causes issues :)

GuilhemN:
Thank you @xabbuh!

composer.json
xabbuh: see #1800

PR #1710, opened 2017-05-29
check if templateData contains the field status_code

dbu:
without this check, some of our error handling fails with the invalid access to the non-existing array key.

### DIFF
--- a/Serializer/Normalizer/ExceptionHandler.php
+++ b/Serializer/Normalizer/ExceptionHandler.php
@@ -42,7 +42,7 @@ class ExceptionHandler extends AbstractExceptionNormalizer implements Subscribin
 
     /**
      * @param JsonSerializationVisitor $visitor
-     * @param Exception                $exception
+     * @param \Exception               $exception
      * @param array                    $type
      * @param Context                  $context
      *
@@ -61,7 +61,7 @@ class ExceptionHandler extends AbstractExceptionNormalizer implements Subscribin
 
     /**
      * @param XmlSerializationVisitor $visitor
-     * @param Exception               $exception
+     * @param \Exception              $exception
      * @param array                   $type
      * @param Context                 $context
      */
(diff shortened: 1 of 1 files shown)


dbu:
huh, what does this failure mean? can it be that the bug was hiding that we output stack traces?

dbu:
it seems the problem happens when we pass an exception as $data in `FOS\RestBundle\Controller\ControllerTrait::view`

Serializer/Normalizer/ExceptionHandler.php
```diff
@@ -102,7 +102,10 @@ protected function convertToArray(\Exception $exception, Context $context)
 
         $templateData = $context->attributes->get('template_data');
         if ($templateData->isDefined()) {
-            $data['code'] = $statusCode = $templateData->get()['status_code'];
+            $data = $templateData->get();
```
GuilhemN: You shouldn't use `$data` here, we voluntarily include only some keys.

GuilhemN:
We indeed didn't think about the case where you would pass an exception manually to the view handler, so except my minor comment :+1:

Serializer/Normalizer/ExceptionHandler.php
dbu: ah, i guess thats also the reason for the failing tests... was not looking close enough to see that $data is already used here.

lsmith77:
damn .. its seems like the failure is caused by an issue we already have in master :(

dbu:
for this issue: is passing an exception to the view layer intended? and should that trigger the exception handler or should we log some warning that something weird has been done when no status_code was provided?

GuilhemN:
> is passing an exception to the view layer intended

I would instead throw the exception but I guess we can also accept passing it directly to the view layer as it's just a data object.

> and should that trigger the exception handler or should we log some warning that something weird has been done when no status_code was provided?

I think it should be managed just like any other value.

dbu:
the build is now green. ok to merge @GuilhemN ?

GuilhemN:
Sure, sorry I forgot your PR.
Thank you @dbu!

PR #1721, opened 2017-06-21
be compatible with FQCN services in Symfony 3.3

xabbuh:
This *could*  be an alternative to symfony/symfony#23239.

### DIFF
--- a/Routing/Loader/RestRouteLoader.php
+++ b/Routing/Loader/RestRouteLoader.php
@@ -126,6 +126,10 @@ class RestRouteLoader extends Loader
             if ($useScope) {
                 $this->container->leaveScope('request');
             }
+
+            if ($controller === $class) {
+                $prefix .= ':';
+            }
         } elseif (class_exists($controller)) {
             // full class name
             $class = $controller;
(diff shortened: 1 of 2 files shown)


stof:
if you want to use the `::` notation, meaning **not** using the controller as a service, the right solution would be to change the order of conditions in the if/elseif chain instead. This will also avoid getting the service from the container to get the class.

xabbuh:
But the goal is to use the controller as a service. This is just a workaround around some inconsistency in FrameworkBundle's `ControllerResolver` implementation. And I am not really convinced that this is the right fix. Probably it's just wrong that `setContainer()` is called for a controller being registered as a service, but I am not sure we should revert that now.

stof:
@xabbuh but your code now is equivalent to witching the order, except for the case where you have a service id equal to a class name, but where the actual service class is not equal to the id (in which case, I'm not sure your code behaves the expected way either)

xabbuh:
We should probably only do this when the id equals the class.

Routing/Loader/RestRouteLoader.php
```diff
@@ -126,6 +126,10 @@ private function getControllerLocator($controller)
             if ($useScope) {
                 $this->container->leaveScope('request');
             }
+
+            if ($controller === $class && class_exists($controller)) {
```
GuilhemN: I don't think we need the class existence check as $class comes from php itself.

xabbuh: indeed

GuilhemN:
It seems the tests need to be updated :)

xabbuh:
It seems tests are passing now. :)

xabbuh:
closing as symfony/symfony#23239 has been merged

PR #1732, opened 2017-07-12
Easier Custom Form Error Handler

phptuts:
I am suggesting we change private methods to protected so that someone can easily change the response of a form error.  Here an example:

https://github.com/phptuts/symfonystarter/blob/f67743e1398b986206d6f721ca2b85a556fb9bd3/src/CoreBundle/Handler/FormErrorHandler.php#L16-L16

Right now I have to over ride to 2 function in order to get what I need done when it would be easier if   it was protected and I just had to over ride one function.  First time doing this so if I am breaking a rule let me know.

Thank you

Noah

### DIFF
--- a/Serializer/Normalizer/FormErrorHandler.php
+++ b/Serializer/Normalizer/FormErrorHandler.php
@@ -86,7 +86,7 @@ class FormErrorHandler extends JMSFormErrorHandler
         return $result;
     }
 
-    private function adaptFormArray(\ArrayObject $serializedForm, Context $context = null)
+    protected function adaptFormArray(\ArrayObject $serializedForm, Context $context = null)
     {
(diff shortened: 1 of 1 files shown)


phptuts:
It makes it easier to wrap the form in a standard response which helps clients like iphone and android.  Here an example of what it could like.

`{
    "meta": {
        "type": "form_errors",
        "paginated": false
    },
    "data": {
        "children": {
            "email": {
                "errors": [
                    "This value is not a valid email address."
                ]
            },
            "plainPassword": {
                "errors": [
                    "This value is too short. It should have 3 characters or more."
                ]
            }
        }
    }
}`

Tobion:
Closing as this PR has conflicts and in general it's better to keep potential extensions point as low as possible for easier maintainance.

PR #1733, opened 2017-07-12
Use `controller.service_arguments` with FOS

louvelmathieu:
By using `controller.service_arguments` with sf 3.3 (http://symfony.com/doc/master/service_container/3.3-di-changes.html#controllers-are-registered-as-services) FOS route generator failed and add dependency to your URL so you need to exclude somes classes.

Exemple : 

    # app/config/services.yml
    services:
        # ...
        # controllers are imported separately to make sure they're public
        # and have a tag that allows actions to type-hint services
        AppBundle\Controller\:
            resource: '../../src/AppBundle/Controller'
            public: true
            tags: ['controller.service_arguments']

Your controller will be :

    use Psr\Log\LoggerInterface;

    class InvoiceController extends Controller
    {
        public function getListAction(LoggerInterface $logger)
        {
            $logger->info('A new way to access services!');
        }
    }

So you need to add class to your FOS configuration to get route `GET /list` and not `GET /list/{logger}` :

    # app/config/config.yml
    fos_rest:
        routing_loader:
            routing_ignore_classes:
                - 'Psr\Log\LoggerInterface'

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -105,6 +105,7 @@ final class Configuration implements ConfigurationInterface
                         ->scalarNode('default_format')->defaultNull()->end()
                         ->scalarNode('prefix_methods')->defaultTrue()->end()
                         ->scalarNode('include_format')->defaultTrue()->end()
+                        ->scalarNode('routing_ignore_classes')->defaultValue([])->end()
                     ->end()
                 ->end()
                 ->arrayNode('body_converter')
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -106,6 +106,7 @@ final class Configuration implements ConfigurationInterface
                         ->scalarNode('prefix_methods')->defaultTrue()->end()
                         ->scalarNode('include_format')->defaultTrue()->end()
                         ->scalarNode('routing_ignore_classes')->defaultValue([])->end()
+                        ->scalarNode('routing_ignore_types')->defaultValue([])->end()
                     ->end()
                 ->end()
                 ->arrayNode('body_converter')
--- a/Resources/doc/5-automatic-route-generation_single-restful-controller.rst
+++ b/Resources/doc/5-automatic-route-generation_single-restful-controller.rst
@@ -423,6 +423,18 @@ Your controller will be :
                 - 'Psr\Log\LoggerInterface'
 
 
+If you are using php7, You can also exclude a type of variable.
+To exclude all object from your routing, you have to add :
+
+.. code-block:: yaml
+
+    # app/config/config.yml
+    fos_rest:
+        routing_loader:
+            routing_ignore_types:
+                - 'object'
+
+
 
 Changing pluralization in generated routes
 ------------------------------------------
--- a/Resources/doc/5-automatic-route-generation_single-restful-controller.rst
+++ b/Resources/doc/5-automatic-route-generation_single-restful-controller.rst
@@ -380,6 +380,50 @@ to limit or add a custom format, you can do so by overriding it with the
         // ...
     }
 
+Use `controller.service_arguments` with FOS
+---------------------------
+
+By using `controller.service_arguments` tags you need to exclude classes.
+
+.. code-block:: yaml
+
+    # app/config/services.yml
+    services:
+        # ...
+
+        # controllers are imported separately to make sure they're public
+        # and have a tag that allows actions to type-hint services
+        AppBundle\Controller\:
+            resource: '../../src/AppBundle/Controller'
+            public: true
+            tags: ['controller.service_arguments']
+
+
+Your controller will be :
+
+.. code-block:: php
+
+    use Psr\Log\LoggerInterface;
+
+    class InvoiceController extends Controller
+    {
+        public function listAction(LoggerInterface $logger)
+        {
+            $logger->info('A new way to access services!');
+        }
+    }
+
+
+.. code-block:: yaml
+
+    # app/config/config.yml
+    fos_rest:
+        routing_loader:
+            routing_ignore_classes:
+                - 'Psr\Log\LoggerInterface'
+
+
+
 Changing pluralization in generated routes
 ------------------------------------------
 
--- a/Routing/Loader/Reader/RestActionReader.php
+++ b/Routing/Loader/Reader/RestActionReader.php
@@ -111,6 +111,12 @@ class RestActionReader
     private $hasMethodPrefix;
 
     /**
+     * Ignore custom classes form route generator
+     * @var array
+     */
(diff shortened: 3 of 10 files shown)


Resources/doc/5-automatic-route-generation_single-restful-controller.rst
```diff
@@ -380,6 +380,50 @@ to limit or add a custom format, you can do so by overriding it with the
...
+    }
+
+
+.. code-block:: yaml
+
+    # app/config/config.yml
+    fos_rest:
+        routing_loader:
+            routing_ignore_classes:
+                - 'Psr\Log\LoggerInterface'
```
GuilhemN: This interface can be added to the hardcoded list :)

DependencyInjection/Configuration.php
```diff
@@ -105,6 +105,7 @@ public function getConfigTreeBuilder()
                         ->scalarNode('default_format')->defaultNull()->end()
                         ->scalarNode('prefix_methods')->defaultTrue()->end()
                         ->scalarNode('include_format')->defaultTrue()->end()
+                        ->scalarNode('routing_ignore_classes')->defaultValue([])->end()
```
GuilhemN: What about `ignored_types`?

Resources/doc/5-automatic-route-generation_single-restful-controller.rst
louvelmathieu: Yes, it's done

DependencyInjection/Configuration.php
louvelmathieu: I add it but it's a feature only for php7 : https://secure.php.net/manual/fr/reflectionparameter.gettype.php

louvelmathieu:
@GuilhemN : Have you take a look of the change ?

DependencyInjection/Configuration.php
GuilhemN: Sorry I meant renaming the option to `ignored_types` because it also accepts interfaces and traits FQCN.

Routing/Loader/Reader/RestActionReader.php
```diff
@@ -500,6 +518,18 @@ private function getMethodArguments(\ReflectionMethod $method)
                 }
             }
 
+            // Dont break php5 compatibility
+            if (method_exists($argument, 'getType')) {
```
GuilhemN: This feature doesn't seem very useful to me

Soullivaneuh:
:-1: For this way. For me this is only a workaround. Plus, the logger example is not really relevant, this is a common service usually injected on controller constructors.

On a huge project with a lot of specific service, this way will be overkill with a huge configuration file.

Instead, we should find a way to make it working without any workaround.

Soullivaneuh:
Or, at least, this should be improved with a compiler pass fetching existing services class to add them to the ignore list. Still a workaround for me, but it avoids filling a big configuration file. :confused:

Soullivaneuh:
BTW, I try updating manually the ignoreClass variable on current stable:

```php
$ignoreClasses = [
    \Symfony\Component\HttpFoundation\Request::class,
    \FOS\RestBundle\Request\ParamFetcherInterface::class,
    \Symfony\Component\Validator\ConstraintViolationListInterface::class,
    \Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter::class,
    MessageInterface::class,
    LoggerInterface::class,
    Logger::class,
];
```

But this change nothing, with this controller:

```php
public function indexAction(LoggerInterface $logger)
{
    dump($logger);
```

I still got this error:

```
The format "" is not supported for deserialization.

Symfony\Component\HttpKernel\Exception\
UnsupportedMediaTypeHttpException
in vendor/friendsofsymfony/rest-bundle/Request/RequestBodyParamConverter.php (line 99)
```

Or maybe this is an another issue?

xorik:
I agree with @Soullivaneuh this isn't convenient.

Imagine that I have 30 controllers with many different services and repositories, passed as arguments. In this case I need to write down 50-100 exclusions?

At least we need something like preg (something like `(Service|Repository)`). But it's also far from perfect

xorik:
Its much better to make a whitelist of classes, than blacklist.So I prefer something like this:
```
fos_rest:
    routing_loader:
        routing_allowed_types:
            - '^AppBundle\\(Document|Entity)'
```

mvrhov:
Or just stop using it like that. Inject everything in constructor, or generate one class per action and again inject everything in constructor.

Laizerox:
@xorik I've written patch which does exactly what you want. The patch is based on conversation in this PR.

It is still WIP since it is missing tests and documentation update but should be done in coming days.

See here: laizerox/FOSRestBundle@919b7cf4584d4f28589db7696e4750cd27dcd21f

Whitelisting is a new option under `routing_loader`. Simply give one string or array under `allowed_types`.

Example:

```
fos_rest:
    routing_loader:
        allowed_types:
            - 'AppBundle\Entity\Class'
            - '^AppBundle\\(Entity|Document)'
```

As you can see it still accepts either FQCN including subclasses but as a new feature regex.c

The code logic already wraps strings into regex.

I hope you will give me more feedback in upcoming PR or directly into commit before PR.

Thank you!

Resources/doc/5-automatic-route-generation_single-restful-controller.rst
```diff
@@ -423,6 +423,18 @@ Your controller will be :
...
+If you are using php7, You can also exclude a type of variable.
+To exclude all object from your routing, you have to add :
+
+.. code-block:: yaml
+
+    # app/config/config.yml
+    fos_rest:
+        routing_loader:
+            routing_ignore_types:
+                - 'object'
```
stof: that's broken. `$argument->getType()` will never return `object`

Routing/Loader/Reader/RestActionReader.php
```diff
@@ -508,6 +518,18 @@ private function getMethodArguments(\ReflectionMethod $method)
...
+                $argumentType = $argument->getType();
+                if ($argumentType) {
+                    foreach ($this->ignoreTypes as $type) {
+                        if ($argumentType === $type) {
```
stof: that's broken. The return value of `$argument->getType()` will be a ReflectionType object, so there is no way to have the same instance in the ignoreTypes property.

bilel-noomene:
Any news on this PR?
Otherwise, what is the best way to inject the services in the FOSRest controllers?

mvrhov:
@bilel-noomene read my post from 6th December

Tobion:
Closing as the current implementation is not working and in general `controller.service_arguments` is mostly a bad idea. Adding a workaround for this is not needed because there is already a workaround as @mvrhov explained.

PR #1734, opened 2017-07-13
ExceptionListener decorate ProfileListener avoiding 302 redirects instead of 403

silverbackdan:
See PR and comments here:
https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1692

I've made an attempt to decorate ProfileListener which has a higher priority than the current exception listener configuration. This is instead of just changing the priority of the FOSRest listener.

This now responds with a 403 instead of a redirect (in my case when using @Security annotations

I'm not as experienced with bigger bundles like this so these changes would need a keen eye to look over and tell me if anything is done badly or should be improved.

Are you able to have a review of this @lsmith77 and @xabbuh 

Not sure of side-effects. I can see at the moment when I use this I get a much bigger exception response than expected - I'm not sure right now if that'd automatically be reduced in production environment as it's extending the ProfilerListener class.

One of the changes which had a cascading effect was the Exception type hinting. Because it no longer extends HttpKernelExceptionListener the duplicateRequest method couldn't be overwritten. It seemed as though the only difference was assigning the exception directly rather than creating a FlattenException object.

Really keen to have thoughts and if I can make improvements I will certainly try!


Controller/ExceptionController.php
```diff
@@ -82,7 +82,7 @@ public function showAction(Request $request, $exception, DebugLoggerInterface $l
      *
      * @return View
      */
-    protected function createView(\Exception $exception, $code, array $templateData, Request $request, $showException)
+    protected function createView(FlattenException $exception, $code, array $templateData, Request $request, $showException)
```
xabbuh: What is the reason for this change?

silverbackdan:
Because I rushed this solution :P The modified listener no longer extends `HttpKernelExceptionListener` - when it did the `duplicateRequest` method was overwritten to not be an Exception but a `FlattenedException`.

Of course other classes down the chain will require that to be an `Exception` to be compatible with other calls. I'll revert all the places I changed `Exception` to `FlattenedException` and create a new service that decorates `HttpKernelExceptionListener` overriding the `duplicateRequest` as it was before. Won't be too long.

Cheers.

silverbackdan:
There's the JSON response I was expected to see now and hopefully these changes shouldn't have knock-on consequences. Do let me know though - Thanks.

silverbackdan:
The failing phpunit tests are mostly the param `profiler_listener.only_exceptions` not being defined. When I'm using this update, it does appear to be setup in `vendor/symfony/framework-bundle/DependencyInjection/FrameworkExtension::registerProfilerConfiguration`

The 1st PHPUnit test fail was a simple fix - I hadn't defined logger as null by default as it should have been. I will continue to see if I can work out why the param can't be found in the tests though.

silverbackdan:
OK, so I've done this all wrong - I've extended the ProfileListener which is only active in dev and test env. Back to the drawing board I think. Please do comment and let me know if you get time to let me know the correct class to be decorating to bring the priority up if you can. But I'll keep looking too - quite keen to fix this one if I can. Cheers.

PR #1736, opened 2017-07-14
Decorate core security bundle exception listener to return 403...

silverbackdan:
...instead of 302 redirect

Attempt 2, this time decorating `Symfony\Component\Security\Http\Firewall\ExceptionListener`

Original PR with suggestion: https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1692

Again as the two in the origin conversation I'd very much appreciate if @lsmith77 and/or @xabbuh can take a look through - I'll have a look at the results of the tests here as well if further adjustments may need to be made.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -103,6 +103,7 @@ final class Configuration implements ConfigurationInterface
                     ->addDefaultsIfNotSet()
                     ->children()
                         ->scalarNode('default_format')->defaultNull()->end()
+                        ->scalarNode('prefix_methods')->defaultTrue()->end()
                         ->scalarNode('include_format')->defaultTrue()->end()
                     ->end()
                 ->end()
--- a/DependencyInjection/FOSRestExtension.php
+++ b/DependencyInjection/FOSRestExtension.php
@@ -58,6 +58,7 @@ class FOSRestExtension extends Extension
         $container->getDefinition('fos_rest.routing.loader.yaml_collection')->replaceArgument(2, $config['routing_loader']['include_format']);
         $container->getDefinition('fos_rest.routing.loader.xml_collection')->replaceArgument(2, $config['routing_loader']['include_format']);
         $container->getDefinition('fos_rest.routing.loader.reader.action')->replaceArgument(3, $config['routing_loader']['include_format']);
+        $container->getDefinition('fos_rest.routing.loader.reader.action')->replaceArgument(5, $config['routing_loader']['prefix_methods']);
 
         foreach ($config['service'] as $key => $service) {
             if ('validator' === $service && empty($config['body_converter']['validate'])) {
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+Copyright (c) 2011-2017 FriendsOfSymfony <http://friendsofsymfony.github.com/>
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
--- a/README.md
+++ b/README.md
@@ -38,6 +38,4 @@ All the installation instructions are located in the [documentation](http://symf
 License
 -------
 
-This bundle is under the MIT license. See the complete license in the bundle:
-
-    Resources/meta/LICENSE
+This bundle is under the MIT license. See the complete license [in the bundle](LICENSE).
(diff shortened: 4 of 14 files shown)


silverbackdan:
The Symfony 4 build error is related to https://symfony.com/blog/new-in-symfony-3-3-dependency-injection-deprecations#deprecated-the-definitiondecorator-class where`Symfony\Component\DependencyInjection\DefinitionDecorator` has been moved to `Symfony\Component\DependencyInjection\ChildDefinition` - I don't think that's to do with this PR but it may just be the asserts that need updating to check for either class name.

The PHP5.5 and Symfony 2.8 tests - I'm not quite sure which bundle config is being loaded in, but I think the `SecurityBundle` just needs adding in wherever necessary.

If I can do anything else though, please do comment, I'll be happy to do what I can do get this issue fixed.

silverbackdan:
I can't quite work out why the --prefer-lowest flag is showing errors on this PR but not other PRs. It looks to me as though a form object is being passed to the Symfony serializer correctly and coming back incorrectly in the `FOS\RestBundle\Serializer\SymfonySerializerAdapter`. Have you seen this happen before by any chance?

I've worked through to get all other tests passing now though and updated where `DefinitionDecorator` is used.

lsmith77:
great stuff .. I will try to have a look over the weekend and also test it with an app where we currently rely on the hack in #1692

composer.json
```diff
@@ -30,6 +30,7 @@
     "require": {
         "php": "^5.5.9|~7.0",
         "psr/log": "^1.0",
+        "symfony/symfony": "2.8.*",
```
lsmith77: I guess this should be removed

silverbackdan: Ah yes of course, that changed as I was testing on early Symfony versions. I'll sort that now.

silverbackdan: That was what was causing the error in Travis-CI checks too it seems :)

lsmith77:
looks good to me!
@GuilhemN @xabbuh wdyt?

Resources/config/exception_listener.xml
```diff
@@ -5,13 +5,22 @@
            xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
 
     <services>
-        <service id="fos_rest.exception_listener" class="FOS\RestBundle\EventListener\ExceptionListener">
-            <tag name="kernel.event_subscriber" />
+        <service id="security.exception_listener" class="FOS\RestBundle\EventListener\ExceptionListener">
```
GuilhemN: We should decorate the core listener instead imo (see https://symfony.com/doc/current/service_container/service_decoration.html). Here in case the FrameworkBundle is registered after the FOSRestBundle, it won't work as expected.

silverbackdan: OK - could you perhaps give me the service ID or class which is the one you'd like decorated? (Possibly a quick technique for finding that out if you'd like me to?)

or is this request more about just the configuration in the link you've provided?

GuilhemN: `security.exception_listener` should be fine to decorate, you should just inject it in the constructor of the fosrest listener instead of instantiating it by hands.

silverbackdan: Ah ok no problem - I'll see if I can get this updated later today. Thanks for the feedback

GuilhemN: Ok, thanks for working on this :)

silverbackdan: No worries, slight issue though `security.exception_listener` is abstract so I can't appear to decorate it like this or pass it into a service as an argument.

silverbackdan: Looking into this with the help of Stof and this issue I found https://github.com/symfony/symfony/issues/11926
https://github.com/symfony/symfony/blob/05c720734bae7f96edb5a3fb9644d360f6301336/src/Symfony/Bundle/SecurityBundle/DependencyInjection/Security/Factory/AbstractFactory.php#L171-184

silverbackdan: OK, well I'm a little stuck now how to move forwards with this one. I'll await feedback because I'm not 100% sure if this is the best way or if there should just be the option to change the priority of the listener currently implemented via the config. Maybe the default value could be 0 or 1 instead.

GuilhemN:
Oops sorry I didn't pushed the right branch after rebasing...
Please run `git pull https://github.com/friendsofsymfony/FOSRestBundle.git silverbackdan-master` and re-push :)

GuilhemN:
I didn't really understand the original issue so I can't tell if this is the right solution. I will just follow @lsmith77 point of view :)

Issue #1737, opened 2017-07-17
Cannot use JMS group overriding in Controller


Rewieer:
I need to use the group overriding functionnality from JMS (http://jmsyst.com/libs/serializer/master/cookbook/exclusion_strategies) in my controller. Yet it doesn't work.
That is because in the Context file, when adding groups, the function `addGroup` doesn't add the key.
So the following
```php
@Rest\View(serializerGroups={"foo", "bar"={"baz"}})
```

Will be added as following in the group list : 
```php
$data = [
    "foo",
    ["baz"]
]
```

All the Context needs to do is to have an optional "key" parameters to allow adding the key from the "addGroups" function.

PR #1716, opened 2017-06-07
Improve Context for "Overriding Groups of Deeper Branches of the Graph"

mbremeersch:
[Exclusion Strategies](http://jmsyst.com/libs/serializer/master/cookbook/exclusion_strategies#overriding-groups-of-deeper-branches-of-the-graph)
My PR is just for having the possibility to put in annotation ```@View()``` the full serialization context example:
```
use JMS\Serializer\SerializationContext;

$context = SerializationContext::create()->setGroups(array(
    'Default', // Serialize John's name
    'manager_group', // Serialize John's manager
    'friends_group', // Serialize John's friends

    'manager' => array( // Override the groups for the manager of John
        'Default', // Serialize John manager's name
        'friends_group', // Serialize John manager's friends. If you do not override the groups for the friends, it will default to Default.
    ),

    'friends' => array( // Override the groups for the friends of John
        'manager_group' // Serialize John friends' managers.

        'manager' => array( // Override the groups for the John friends' manager
            'Default', // This would be the default if you did not override the groups of the manager property.
        ),
    ),
));
$serializer->serialize($john, 'json', $context);
```

```
<?php
//MyController.php
use FOS\RestBundle\Controller\Annotations as Rest;
/**
 * @Rest\View(serializerGroups={
 *    "Default",
 *    "manager_group",
 *    "friends_group",
 *    "manager":{
 *                 "Default",
 *                 "friends_group"
 *              },
 *    "friends":{
 *                 "manager_group",
 *                 "manager":{
 *                               "Default"
 *                           }
 *
 *              },
 *    "manager":{"Default","friends_group"}
 *   })
 *
 *
 *
 */
 public function exampleAction()
 {
    //...
 }

```

PR #1738, opened 2017-07-17
handling key for groups

Rewieer:
This is a PR to solve #1737

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ vendor
 composer.lock
 composer.phar
 .php_cs.cache
+.idea
--- a/Context/Context.php
+++ b/Context/Context.php
@@ -128,16 +128,21 @@ final class Context
      * Adds a normalization group.
      *
      * @param string $group
+     * @param string $key
      *
      * @return self
      */
-    public function addGroup($group)
+    public function addGroup($group, $key = null)
     {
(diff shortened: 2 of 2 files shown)


GuilhemN:
Duplicate of https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1716

I did not have the time to look at it but at first sight I don't really like it, I'd prefer having a `setGroups` method and maybe even deprecate the adders.

Rewieer:
This method is quite cheap I must admit, but a setGroups method requires more work, and this feature is quite important. Can't we put it "as is" for now (it doesn't really decrease code readibility or so) and refactor the group part later ?

GuilhemN:
Rethinking about it, I don't think that's the right solution, I would rather prefer using `setGroups` with `array_merge` in our code.

Rewieer:
So you want me to remove addGroup and add a setGroups function. I will try to.

GuilhemN:
There is already a `setGroups` function, we just need to use it. See https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1740 :)

GuilhemN:
Closing as https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1740 is merged.

PR #1739, opened 2017-07-21
use ChildDefinition when possible

xabbuh:
`ChildDefinition` was introduced in Symfony 3.2 and replaces the `DefinitionDecorator` class which will be removed in Symfony 4.0.

### DIFF
--- a/DependencyInjection/Compiler/FormatListenerRulesPass.php
+++ b/DependencyInjection/Compiler/FormatListenerRulesPass.php
@@ -11,6 +11,7 @@
 
 namespace FOS\RestBundle\DependencyInjection\Compiler;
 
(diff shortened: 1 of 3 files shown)


GuilhemN:
instead of duplicating code, can we do 
```php
$childDefinitionClass = class_exists(ChildDefinition::class) ? ChildDefinition::class : DefinitionDecorator::class;

$definition = new $childDefinitionClass(...);

$this->assertInstanceOf($childDefinitionClass, ...);
```
?

xabbuh:
We could. Do you find that more readable?

GuilhemN:
Yes and more maintainable as here we have two update two lines instead of one.

PR #1741, opened 2017-07-24
Update view_response_listener.rst

dsuurlant:
It took me a while to figure out what was going wrong in my configuration - the fact that the setting needs be false for the View Response Listener to work is not actually mentioned, only implied.

### DIFF
--- a/Resources/doc/view_response_listener.rst
+++ b/Resources/doc/view_response_listener.rst
@@ -5,7 +5,7 @@ The view response listener makes it possible to simply return a ``View``
 instance from action controllers. The final output will then automatically be
 processed via the listener by the ``fos_rest.view_handler`` service.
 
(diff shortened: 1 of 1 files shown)


GuilhemN:
Well in fact it's the opposite, it must be set to `true` as specified a few lines later 

> This requires adding the SensioFrameworkExtraBundle to your vendors and making sure sensio_framework_extra.view.annotations is set to false.

dsuurlant:
That's exactly where the confusion comes from.

- If you want to use the View object as a response: `return new View($data)` then `sensio_framework_extra.view.annotations` must be set to `false`.

- If you want to  return data and have it processed further by what you specify in the annotations (serializer groups, status codes, etc) `sensio_framework_extra.view.annotations` must be set to `true`.

PR #1742, opened 2017-07-24
Remove need to register the JMS serializer before the FOSRestBundle

GuilhemN:
Fix https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1600 in a nicer way


GijsL:
I have a custom FormErrorHander, which i set with

```
parameters:
```

I think the CompilerPass now will overwrite my parameter? Should I create a bug report for this?

xabbuh:
@GijsL Yes, I think it's better to move the discussion into a new issue.

PR #1744, opened 2017-07-28
Make ExceptionController::getAndCleanOutputBuffering() protected

jhuh-illuminateed:
Allow getAndCleanOutputBuffering call when the class is extended

### DIFF
--- a/Controller/ExceptionController.php
+++ b/Controller/ExceptionController.php
@@ -143,7 +143,7 @@ class ExceptionController
      *
      * @return string
      */
-    private function getAndCleanOutputBuffering($startObLevel)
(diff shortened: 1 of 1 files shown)


GuilhemN:
I'm not sure about this change, if it becomes `protected` it ends up in our public api while it looks `internal` to me.

dlpetrie:
The code refers to the TwigBundle ExceptionController as an example.  The Twig ExceptionController is protected, so why wouldn't this?

GuilhemN:
@dlpetrie maybe they can no longer make it private because it is in their public api :)
Imo if you are reusing this method you are basically rewriting the controller and I don't think we should support this case.

I am closing for now, feel free to reopen if you think this should be reconsidered.

PR #1745, opened 2017-08-03
add fos_rest.exception.service to configuration

egils:
I noticed that for exception.service configuration option was not added with [this commit](https://github.com/FriendsOfSymfony/FOSRestBundle/commit/a3ee6d7c4427d569d577e717ab8dccc1a857abbd). Without this option [FOSRestExtension.php#L328](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/DependencyInjection/FOSRestExtension.php#L328) is unreachable and default ExceptionListener is always registered as an event subscriber.
If there is an actual reason why this listener should always be active, code block in FOSRestExtension should be removed.

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -406,6 +406,7 @@ final class Configuration implements ConfigurationInterface
                     ->canBeEnabled()
                     ->children()
                         ->scalarNode('exception_controller')->defaultNull()->end()
+                        ->scalarNode('service')->defaultNull()->end()
                         ->arrayNode('codes')
                             ->useAttributeAsKey('name')
                             ->beforeNormalization()
--- a/Resources/doc/configuration-reference.rst
+++ b/Resources/doc/configuration-reference.rst
@@ -76,6 +76,7 @@ Full default configuration
         exception:
             enabled:              false
             exception_controller:  null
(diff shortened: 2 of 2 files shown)


GuilhemN:
I don't really get why we allow to override our listeners service... For me it's the same as just disabling them.

egils:
@GuilhemN it is disabling the listener. If fos_rest.exception.service (any) is defined in configuration, [this code](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/DependencyInjection/FOSRestExtension.php#L328) disables a default ExceptionListener. And all other listeners are disabled in same fashion after the commit that is linked in the description.

GuilhemN:
Yes but to disable the listener you can as well do `fos_rest: exception: false`, right?

egils:
Yes, but that would also disable a default controller. In my case I want to keep the default controller and have a custom ExceptionListener. But that's not the main reason I opened this PR. At the current state there is a piece of unreachable code. To fix it I see two obvious ways:
* add fos_rest.exception.service configuration option to override a default listener (as this PR does);
* remove unreachable code and force users handle exception responses fully on their own.

GuilhemN:
Fair point, thank you @egils!

PR #1749, opened 2017-08-11
Skip abstract classes in route generation

mjpvandenberg:
When scanning a directory for controllers to auto-generate routes for, abstract controller classes should be skipped. HTTP requests to those routes cause the controller resolver to try instantiate an abstract class.

### DIFF
--- a/Routing/Loader/ClassUtils.php
+++ b/Routing/Loader/ClassUtils.php
@@ -36,7 +36,11 @@ class ClassUtils
             }
 
             if (true === $class && T_STRING === $token[0]) {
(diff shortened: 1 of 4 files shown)


xabbuh:
Can you add a test case to prevent future regressions?

mjpvandenberg:
Moved the check into `ClassUtils::findClassInFile()`, so that RestRouteLoader will throw an exception if an abstract controller was configured.

Tobion:
Closing as automatic route generation is deprecated: #1954

PR #1759, opened 2017-09-04
Add prefix to route name

ThisIsAreku:
When having multiple controller with same action name, routes are overwritten. This set a prefix using the controller's name to ensure uniqueness.

EDIT: I forgot to edit unit tests. Feel free to discuss about this PR, I'll edit them later.


Tobion:
While I think this the current naming is indeed broken because of easily conflicting route names, see #1954,, changing it like this is probably a BC break as it changes the route names.

xabbuh:
I am closing here as I agree with @Tobion.

PR #1760, opened 2017-09-04
Fix enumerated list

Landerstraeten:
Hardcode the list instead of auto-enumeration

### DIFF
--- a/Resources/doc/versioning.rst
+++ b/Resources/doc/versioning.rst
@@ -79,64 +79,64 @@ If you want to version by Accept header, you will need to do the following:
 
 #. The format listener must be enabled
 
-See :doc:`Format Listener <format_listener>`
+   See :doc:`Format Listener <format_listener>`
 
 #. The client must pass the requested version in his header like this :
 
-.. code-block:: yaml
+   .. code-block:: yaml
 
-    Accept:application/json;version=1.0
+       Accept:application/json;version=1.0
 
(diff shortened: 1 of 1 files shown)


xabbuh:
I don't see why we would want to do that. Not enumerating items statically has the advantage than we can delete/add items in the middle without having to adjust all other items.

Landerstraeten:
@xabbuh, check http://symfony.com/doc/master/bundles/FOSRestBundle/versioning.html
It's 3 times '1'...

Resources/doc/versioning.rst
```diff
@@ -77,17 +77,17 @@ automatically in the ``ViewHandler``.
 
 If you want to version by Accept header, you will need to do the following:
 
-#. The format listener must be enabled
+1. The format listener must be enabled
 
 See :doc:`Format Listener <format_listener>`
```
xabbuh: We need to indent the content of the list items by three spaces each. That should fix the rendering of the list.

Landerstraeten: Oh ok. I'll update this.

xabbuh:
I see. Thanks for the hint. Actually, we currently have three lists with one item each. Indenting the other paragraphs of the items will solve that (see my inline comment above).

Landerstraeten:
@xabbuh, are you happy with the change?

PR #1763, opened 2017-09-19
add ViewHandlerInterface service alias

### LINKED ISSUE #1724: Deprecation warning on Symfony 3.3
On a fresh install with Symfony 3.3.2, i get the following deprecation notice in my logs:

    Autowiring services based on the types they implement is deprecated since 
    Symfony 3.3 and won't be supported in version 4.0. You should rename (or alias) the
    "fos_rest.view_handler.default" service to "FOS\RestBundle\View\ViewHandlerInterface" instead.

The only change I made besides installing FOSRestBundle and JMSSerializerBundle was to have the DefaultController extend FOSRestController like so:

    class DefaultController extends FOSRestController
    {
        /**
         * @Route("/", name="homepage")
         */
        public function indexAction(Request $request)
        {
            // replace this example code with whatever you need
            return $this->render('default/index.html.twig', [
                'base_dir' => realpath($this->getParameter('kernel.project_dir')).DIRECTORY_SEPARATOR,
            ]);
        }
    }

I wasn't able to find anything in the issues or PRs for this, is there a workaround available?

xabbuh:
This should fix #1724.

### DIFF
--- a/Resources/config/view.xml
+++ b/Resources/config/view.xml
@@ -19,6 +19,8 @@
             <argument /> <!-- default engine -->
         </service>
 
+        <service id="FOS\RestBundle\View\ViewHandlerInterface" alias="fos_rest.view_handler.default" />
+
         <service id="fos_rest.view_handler.jsonp" class="FOS\RestBundle\View\JsonpHandler" public="false">
             <argument /> <!-- JSONP callback parameter -->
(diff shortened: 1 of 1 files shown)


zbyyna:
Works! Thank you!

PR #1764, opened 2017-09-21
new "parents" option in @RouteResource annotation to defines parents route resources parts

bnowak:
With that option we can define parents of route resource directly in @FOS\RestBundle\Controller\Annotations\RouteResource annotation. So we don't have to define routing in yml or xml files when we want to use seperate controllers for subresources.

I figured out that this case can be also cover by:
```php
use FOS\RestBundle\Controller\FOSRestController;

/**
 * @Rest\RouteResource("User")
 */
class UserController extends FOSRestController
{
    public function getAction($id)
    {} // "get_user" [GET] /users/{id}
}

/**
 * @Rest\RouteResource("User")
 */
class UserCommentsController extends FOSRestController
{
    public function cgetCommentsAction($userId)
    {} // "get_user_comments" [GET] /users/{userId}/comments
}
```

but when we have two or more words in subresource name route loader break every word into separate route resources:

```php
/**
 * @Rest\RouteResource("User")
 */
class PhoneNumberController extends FOSRestController
{
    public function cgetPhoneNumbersAction($userId)
    {} // "get_user_phone_numbers" [GET] /users/{userId}/phone/numbers
}
```

### DIFF
--- a/Tests/Fixtures/Controller/AnnotatedManyParentsArticleController.php
+++ b/Tests/Fixtures/Controller/AnnotatedManyParentsArticleController.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\RestBundle\Tests\Fixtures\Controller;
+
+use FOS\RestBundle\Controller\Annotations as Rest;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+
+/**
+ * @author Bartłomiej Nowak <barteknowak90@gmail.com>
+ * @Rest\RouteResource("Comment", parents={"Article", "Media"})
+ */
+class AnnotatedManyParentsArticleController extends Controller
+{
+    /**
+     * "get_article_media_comment" [GET] /articles/{slug}/media/{mediaId}/comments/{id}.{_format}
+     */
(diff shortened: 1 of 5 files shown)


smokarz:
totally agree with @bnowak , this feature would help a lot in development

gcmikolaj:
+1

Tests/Fixtures/Controller/AnnotatedManyParentsArticleController.php
```diff
@@ -0,0 +1,36 @@
...
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\RestBundle\Tests\Fixtures\Controller;
+
+use FOS\RestBundle\Controller\Annotations as Rest;
+use Symfony\Bundle\FrameworkBundle\Controller\Controller;
```
Tobion: this class is deprecated. you don't need any base controller class here at all.

Tobion:
I think route creation based on the controller/action names is a broken design. See #1954
So I'd prefer to not add more of those strange things. As the people commenting on this PR seem to be using this feature, I'd like to ask you what do you think? Why would you prefer creating routes like this and adding a comment in the action to explain what it does, instead of directly creating a route like `@Get("/articles/{slug}/media/{mediaId}/comments/{id}.{_format}")`? That seems more self-explaining to me than `@Rest\RouteResource("Comment", parents={"Article", "Media"})`

bnowak:
I agree with you,
It's old PR and at that time I thought it was a good solution but not now. The idea was to not create redundant YML files with routes.

I'm closing this PR.

PR #1769, opened 2017-10-03
Fix problems with map option

roukmoute:
Currently we have some problems with `map` options.
In fact it is with `map` and anothers options like `nullable` or `notBlank`.

### DIFF
--- a/Controller/Annotations/AbstractScalarParam.php
+++ b/Controller/Annotations/AbstractScalarParam.php
@@ -67,6 +67,9 @@ abstract class AbstractScalarParam extends AbstractParam
             if (false === $this->nullable) {
                 $constraints[] = new NotNull();
             }
+            if (false === $this->allowBlank) {
+                $constraints[] = new NotBlank();
+            }
         }
 
         return $constraints;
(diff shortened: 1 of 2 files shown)


roukmoute:
Up?

Controller/Annotations/AbstractScalarParam.php
```diff
@@ -63,6 +64,12 @@ public function getConstraints()
             $constraints = array(
                 new All(array('constraints' => $constraints)),
             );
+            if (false === $this->nullable) {
+                $constraints[] = new NotNull();
+            }
+            if (false === $this->allowBlank) {
```
xabbuh: I think we should move the `if` from line 57 after the `if ($this->map)` block instead.

Controller/Annotations/AbstractScalarParam.php
```diff
@@ -63,6 +64,12 @@ public function getConstraints()
             $constraints = array(
                 new All(array('constraints' => $constraints)),
             );
+            if (false === $this->nullable) {
```
xabbuh: Isn't this something that should be added when `$this->map` is `false` too?

GuilhemN: If `$this->map` is `false` then it's added [here](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/da8d1ff7d91e5e178022b82a206eacf9b604f3b1/Controller/Annotations/AbstractParam.php#L69).

GuilhemN:
Thank you @roukmoute!

Controller/Annotations/AbstractScalarParam.php
GuilhemN: see https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1970

PR #1771, opened 2017-10-12
Have REST route generator ignore request body ParamConverters

### LINKED ISSUE #1198: Make a better marriage between body and param converter
From the [docs](http://symfony.com/doc/current/bundles/FOSRestBundle/request_body_converter_listener.html):

> You will probably want to disable the automatic route generation (@NoRoute) for routes using the body converter, and instead define the routes manually to avoid having the deserialized, type hinted objects ($post in this example) appear in the route as a parameter.

And that's unfortunately true, I have to redefine entirely the route:

``` php
/**
 * @param Server     $server
 * @param ServerStat $stat
 *
 * @Route("/servers/{server}/stats")
 * @Security("is_granted('ROLE_ADMIN')")
 * @ParamConverter("stat", converter="fos_rest.request_body")
 *
 * @return View
 */
public function postAction(Server $server, ServerStat $stat)
{
}
```

That works, but it is very painful IMHO.

There is any way to improve this management in order to avoid this?

If you can give me some clue, I could maybe work on it. :+1:

Parent5446:
When fetching a list of function parameters from a controller action
for use as URL parameters, check for the @ParamConverter
annotation. If a parameter has the annotation with the
"fos_rest.request_body" converter, then ignore it and do not include
it in the list of parameters (thus removing it from the URL).

This allows automatic route generation and ParamConverter to co-exist,
as otherwise a parameter that is expected to be derived from the body
would also be included as a URL parameter, requiring the developer to
manually specify the route.

Fixes #1198

### DIFF
--- a/Routing/Loader/Reader/RestActionReader.php
+++ b/Routing/Loader/Reader/RestActionReader.php
@@ -475,6 +475,17 @@ class RestActionReader
         // ignore all query params
         $params = $this->paramReader->getParamsFromMethod($method);
 
+        // check if a parameter is coming from the request body
+        $ignoreParameters = [];
+        if (class_exists('Sensio\\Bundle\\FrameworkExtraBundle\\Configuration\\ParamConverter')) {
+            $ignoreParameters = array_map(function ($annotation) {
+                return
+                    $annotation instanceof \Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter &&
+                    $annotation->getConverter() === 'fos_rest.request_body'
+                    ? $annotation->getName() : null;
+            }, $this->annotationReader->getMethodAnnotations($method));
+        }
+
         // ignore several type hinted arguments
         $ignoreClasses = [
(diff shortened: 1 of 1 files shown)


tonivdv:
@GuilhemN It would be great to fix this issue.  Are the failing tests the main reason this PR was never accepted? If yes, @Parent5446 I could help you make this PR pass the tests?

Let me know

GuilhemN:
Well this PR misses a test case, except that it is all good :+1:

GuilhemN:
They were failing 😅 but it seems it was a network issue because I just relaunched them to fix them

Parent5446:
:sweat_smile: I was wondering if this PR would ever see the light of day. Let me know if there's anything I need to do to get it merge-worthy (although it seems like it might be already).

GuilhemN:
Fixed in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1934, thanks!

Issue #1768, opened 2017-10-02
FOS Rest Bundle is incompatible with Sensio Framework Extra Bundle v4.0


Amunak:
Specifically when you use Sensio bundle v4 and set this configuration for FOS Rest:
```yaml
fos_rest:
    view:
        view_response_listener: true
```

an exception is thrown from `ConfigurationCheckPass:34`: `You must enable the SensioFrameworkExtraBundle view annotations to use the ViewResponseListener.`

This is probably due to the fact that you are checking for the `sensio_framework_extra.view.listener` service which was either made private or renamed (not sure but it's probably the first case as per [their changelog](https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/CHANGELOG.md#40)).

MichaelHindley:
I've ran into this today:

```Fatal error: Uncaught RuntimeException: You need to enable the parameter converter listeners in SensioFrameworkExtraBundle when using the FOSRestBundle RequestBodyParamConverter in /var/www/symfony/vendor/friendsofsymfony/rest-bundle/DependencyInjection/Compiler/ConfigurationCheckPass.php on line 29```

It seems indeed that it depends on stuff that is not available anymore!

**Comment by MonsieurBon** [comment]
Same problem with Sensio Framework Extra Bundle v5.0. I have set view_response_listener to 'force'.

**Comment by connorhu** [comment]
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/4a1b70199943460273c4ab8f8ff841c8c79e0d28/DependencyInjection/Compiler/ConfigurationCheckPass.php#L33

The new id is:
Sensio\Bundle\FrameworkExtraBundle\EventListener\TemplateListener

**Comment by pculka** [comment]
I'm in same boat... using extra bundle v 5.0.1 with view_response_listener: force or true.

**Comment by aequasi** [comment]
Simple fix for anyone that needs it:

Throw this in your services:

```
    sensio_framework_extra.view.listener:
        alias: Sensio\Bundle\FrameworkExtraBundle\EventListener\TemplateListener
```

Daredzik:
@aequasi give a quick fix but the real problem is in `bundles.php`, u need notice it the `Sensio\Bundle\FrameworkExtraBundle\SensioFrameworkExtraBundle` must be an after `FOS\RestBundle\FOSRestBundle` to work properly !

PR #1773, opened 2017-10-16
Sensio framework extra4.x

pculka:
fixes https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1768

### DIFF
--- a/DependencyInjection/Compiler/ConfigurationCheckPass.php
+++ b/DependencyInjection/Compiler/ConfigurationCheckPass.php
@@ -30,7 +30,8 @@ final class ConfigurationCheckPass implements CompilerPassInterface
         }
 
         if ($container->has('fos_rest.view_response_listener') && isset($container->getParameter('kernel.bundles')['SensioFrameworkExtraBundle'])) {
-            if (!$container->has('sensio_framework_extra.view.listener')) {
+            $config = $container->getExtensionConfig('sensio_framework_extra');
+            if (!(true === $config[0]['view']['annotations'] || $container->has('sensio_framework_extra.view.listener')))  {
                 throw new \RuntimeException('You must enable the SensioFrameworkExtraBundle view annotations to use the ViewResponseListener.');
             }
         }
--- a/DependencyInjection/Compiler/ConfigurationCheckPass.php
+++ b/DependencyInjection/Compiler/ConfigurationCheckPass.php
@@ -30,8 +30,7 @@ final class ConfigurationCheckPass implements CompilerPassInterface
         }
 
         if ($container->has('fos_rest.view_response_listener') && isset($container->getParameter('kernel.bundles')['SensioFrameworkExtraBundle'])) {
-            $config = $container->getExtensionConfig('sensio_framework_extra');
-            if (!(true === $config[0]['view']['annotations'] || $container->has('sensio_framework_extra.view.listener')))  {
+            if (!($container->has('sensio_framework_extra.view.listener') ^ $container->hasDefinition('Sensio\Bundle\FrameworkExtraBundle\EventListener\TemplateListener'))) {
                 throw new \RuntimeException('You must enable the SensioFrameworkExtraBundle view annotations to use the ViewResponseListener.');
             }
         }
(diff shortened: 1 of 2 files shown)


DependencyInjection/Compiler/ConfigurationCheckPass.php
```diff
@@ -30,7 +30,8 @@ public function process(ContainerBuilder $container)
         }
 
         if ($container->has('fos_rest.view_response_listener') && isset($container->getParameter('kernel.bundles')['SensioFrameworkExtraBundle'])) {
-            if (!$container->has('sensio_framework_extra.view.listener')) {
+            $config = $container->getExtensionConfig('sensio_framework_extra');
+            if (!(true === $config[0]['view']['annotations'] || $container->has('sensio_framework_extra.view.listener')))  {
```
stof: this is wrong for multiple reasons:

- you check only the first config of the bundle, which is not the final config at all (subsequent configs could enable it, or could re-disable it)
- there is no guarantee at all that `$config[0]['view']['annotations']` exists (the user might be configuring other parts of SensioFrameworkExtraBundle
- this leaves out any processing that the bundle could do on the config (which includes the fact of defining a default value)

pculka: Well in that case the only possible check is to loop through all the configuration and merge it. Default configuration is viewListener enabled in FrameworkExtraBundle...
Meaning the whole view.listener check useless and the sensioframeworkextra bundle dep should be properly written in the documentation?

pculka: Updated the PR to use a xor check against the new TemplateListener which replaces ViewReponseListener in SensioFrameworkExtraBundle  4+ (i.e. it either has new config or old config)

DependencyInjection/Compiler/ConfigurationCheckPass.php
```diff
@@ -30,7 +30,7 @@ public function process(ContainerBuilder $container)
         }
 
         if ($container->has('fos_rest.view_response_listener') && isset($container->getParameter('kernel.bundles')['SensioFrameworkExtraBundle'])) {
-            if (!$container->has('sensio_framework_extra.view.listener')) {
+            if (!($container->has('sensio_framework_extra.view.listener') ^ $container->hasDefinition('Sensio\Bundle\FrameworkExtraBundle\EventListener\TemplateListener'))) {
```
derrabus: Using a bitwise xor operator (`^`) here seems odd. Also, I don't see why the case that both definitions exist should be excluded explicitly. I'd prefer `||` instead.

Also, since we have php 5.5 here, you can safely use `TemplateListener::class` instead of the FQCN string.

derrabus:
The tests fail because there's obviously a similar issue with `RequestBodyParamConverter`.

pculka:
The xor is there as the configuration is mutually exclusive. Either you use the old verion or a new one. There’s nothing between.
I will check the ParamConverter too

pculka:
And also will swap out FQCN

derrabus:
> The xor is there as the configuration is mutually exclusive. Either you use the old verion or a new one. There’s nothing between.

True, but there's no reason to force this exclusiveness. If an application (or the FrameworkExtraBundle devs) decide to add an alias to the old service ID as a compatibility layer, there's no reason for FOS REST to fail on this.

DependencyInjection/Compiler/ConfigurationCheckPass.php
merk: If SensioFrameworkExtraBundle is not installed (and isnt required by FOSRestBundle), you cannot use TemplateListener::class because you will end up with the string 'TemplateListener' which is not the same thing.

derrabus: @merk This is wrong. `::class` will always resolve to the FQCN in your `use` statement, no matter if that class exists or not. Also, neither `use` nor `::class` will trigger autoloading. So it is indeed perfectly save to use it here.

https://3v4l.org/RkQNN

pculka:
Well, I've updated the code and fixed most of the build except Symfony 3.4 and 4.0, which fail due different tests, not those relevant to this PR.
Can somebody please merge this, so we can start using also newer versions of SymfonyFrameworkExtraBundle?

luispabon:
I'm very keen on this as well, just coding something up for a 2018 release and naturally we're working against the very latest

xabbuh:
Thank you @pculka.

PR #1774, opened 2017-10-19
Added translation domain to FormErrorNormalizer and FormErrorHandler

prosalov:
Added a translation domain parameter (validator.translation_domain) to the FormErrorNormalizer to be able to use actual translation domain from config.yml instead of 'validators' one. Also, I've added translation domain parameter to FormErrorHandler in order to be compatible with a new constructor signature of JMS Serializer's FormErrorHandler. Please take a look.

```
framework:
    validation:      { translation_domain: messages }
```

```
parameters:
    validator.translation_domain: messages
```

### DIFF
--- a/Resources/config/serializer.xml
+++ b/Resources/config/serializer.xml
@@ -17,6 +17,7 @@
         <!-- Normalizes FormInterface when using the symfony serializer -->
         <service id="fos_rest.serializer.form_error_normalizer" class="FOS\RestBundle\Serializer\Normalizer\FormErrorNormalizer" public="false">
             <argument type="service" id="translator" />
+            <argument>%validator.translation_domain%</argument>
             <tag name="serializer.normalizer" priority="-10" />
         </service>
     </services>
--- a/Serializer/Normalizer/FormErrorNormalizer.php
+++ b/Serializer/Normalizer/FormErrorNormalizer.php
@@ -25,9 +25,12 @@ class FormErrorNormalizer implements NormalizerInterface
 {
     private $translator;
 
-    public function __construct(TranslatorInterface $translator)
+    private $translationDomain;
+
+    public function __construct(TranslatorInterface $translator, $translationDomain = 'validators')
     {
         $this->translator = $translator;
+        $this->translationDomain = $translationDomain;
     }
 
     /**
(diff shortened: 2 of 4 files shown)


Resources/config/serializer.xml
```diff
@@ -17,6 +17,7 @@
         <!-- Normalizes FormInterface when using the symfony serializer -->
         <service id="fos_rest.serializer.form_error_normalizer" class="FOS\RestBundle\Serializer\Normalizer\FormErrorNormalizer" public="false">
             <argument type="service" id="translator" />
+            <argument>%validator.translation_domain%</argument>
```
Tobion: using DI paramters as extension point is not a good solution. we do not need a paramter for this. having a constructor argument on the FormErrorHandler is enough as you can then simply decorate the service and change the translation domain.

Serializer/Normalizer/FormErrorNormalizer.php
```diff
@@ -25,9 +25,12 @@ class FormErrorNormalizer implements NormalizerInterface
 {
     private $translator;
 
-    public function __construct(TranslatorInterface $translator)
+    private $translationDomain;
+
+    public function __construct(TranslatorInterface $translator, $translationDomain = 'validators')
```
Tobion: please add string parameter type

Tests/Serializer/Normalizer/FormErrorNormalizerTest.php
```diff
@@ -0,0 +1,120 @@
...
+
+use FOS\RestBundle\Serializer\Normalizer\FormErrorNormalizer;
+use Symfony\Component\Translation\Translator;
+
+class FormErrorNormalizerTest extends \PHPUnit_Framework_TestCase
+{
+    public function testDefaultTranslationDomain()
+    {
+        /** @var Translator|\PHPUnit_Framework_MockObject_MockObject $translator */
+        $translator = $this->getMockBuilder('Symfony\Component\Translation\TranslatorInterface')->getMock();
```
Tobion: please update the code to use `::class` constant

Tests/Serializer/Normalizer/FormErrorNormalizerTest.php
```diff
@@ -0,0 +1,120 @@
...
+
+        $translator->expects($this->once())
+            ->method('trans')
+            ->with(
+                $this->equalTo('error!'),
+                $this->equalTo([]),
+                $this->equalTo('validators')
+            );
+
+        $formError = $this->getMockBuilder('Symfony\Component\Form\FormError')->disableOriginalConstructor()->getMock();
```
Tobion: please use `$this->createMock` instead

xabbuh:

PR #1780, opened 2017-11-03
compatibility with Sensio Framework v4.0

connorhu:
The ID of the listener has changed, so we need to check both.
bug #1768


xabbuh:
Thank you @connorhu. I am closing here nonetheless as the same issue will already be fixed by #1773.

ruudk:
@xabbuh Could this be included in a new tagged release?

PR #1799, opened 2017-11-26
do not silence error logs

xabbuh:
The change done in #1797 had the disadvantage that it silenced all
log entries. Using the `SHELL_VERBOSITY` env var allows to define a
level for which log entries should still be shown.


xabbuh:
requires symfony/symfony#25168 to work

xabbuh:
[This information](https://travis-ci.org/FriendsOfSymfony/FOSRestBundle/jobs/307877898#L577-L603) was lost previously.

Issue #1790, opened 2017-11-20
Symfony4 | Expose Services


juliusstoerrle:
When using SF4 we get following ServiceNotFoundException:

> The "fos_rest.view_handler" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection instead.

This is because in SF4 all Services are private by default.

I'm happy to submit a pull request, where I would make all services public which do not have a visibility set yet, or should i go through and only enable those ones that give me an ServiceNotFound?

As I'm a first time contributor, is there anything i should pay attention to? I did not find a contrib readme.

(Tagging @xabbuh as he just suggested me to do a PR for this)

xabbuh:
I think we should make a case by case decision. Is each service actually meant to be used by third party code? Or is it only there for internal purposes? Services that are actually meant to be used could be turned to public to still allow users to make use of service location with the Symfony container. For the latter, we IMO don't need to do anything as users will see a deprecation with Symfony 3.4.

juliusstoerrle:
Are there any services a consumer needs? I do not access the container directly in my code but still get the ServiceNotFound for "fos_rest.view_handler"

xabbuh:
Does that happen when using Symfony 3.4 or 4.0? Would you be able to provide a small reproducing project?

xabbuh:
> Are there any services a consumer needs?

I don't have any in mind. Could also be that none of them really needs to be exposed (at least there shouldn't be many that should turned into public ones).

juliusstoerrle:
SF4, will try a demo but might take me a day or two

xabbuh:
Could you downgrade to Symfony 3.4 and check for the exact deprecation message (there should be one)?

juliusstoerrle:
Making a service is also needed, when the bundle itself accesses the container.
I just had a look at the [FosRestController](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Controller/FOSRestController.php) and that might be the culprit. Look at line 34

` $this->viewhandler = $this->container->get('fos_rest.view_handler');`

xabbuh:
I had a look at the code and I wonder how this can happen. As far as I can see, the service seems to be set to public: https://github.com/FriendsOfSymfony/FOSRestBundle/blob/3e65ef5f5cef2ecc6f6a18d204dc5193f0feb272/DependencyInjection/FOSRestExtension.php#L246

[...]

PR #1805, opened 2017-11-28
tests for definitions before they might be removed

xabbuh:
If the `JMSHandlersPass` is registered as a `TYPE_REMOVE` pass, the
definitions it depends on may already be removed before by the built-in
`RemoveUnusedDefinitionsPass`. That compiler pass removes private
services that are not referenced. This becomes critical, for example,
when running a Symfony 4 based application where services will be
private by default.

This is the first PR of a series of changes we need to do to fully solve
#1790.

### DIFF
--- a/FOSRestBundle.php
+++ b/FOSRestBundle.php
@@ -39,6 +39,6 @@ class FOSRestBundle extends Bundle
         $container->addCompilerPass(new FormatListenerRulesPass());
         $container->addCompilerPass(new TwigExceptionPass());
         $container->addCompilerPass(new JMSFormErrorHandlerPass());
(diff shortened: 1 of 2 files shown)


juliusstoerrle:
@xabbuh Thanks for looking into this. I just pulled your changes and my exceptions get serialized without any problem. I currently can not see any other issues, but I guess I'm using only a small subset off the features of the bundle.

xabbuh:
@juliusstoerrle Thanks for confirming. I might have to sync your example project with the other pending changes in JMS Serializer again, but IIRC I still got some errors this weekend.

PR #1818, opened 2017-12-12
Refactoring tests

carusogabriel:
I've refactored some tests using:
- `assertCount` instead of `count` function;
- `assertInstanceOf` instead of `instanceof` operator;
- `assertArrayHasKey` instead of `isset` function;
- `assertNull` instead of comparison with `null` keyword;
- `assertFalse` instead of comparison with `false` keyword;
- `assertTrue` instead of comparison with `true` keyword;
- `assertInternalType` instead of `is_*` functions;
- `assertSame` instead of strict comparisons.

### DIFF
--- a/Tests/Context/ContextTest.php
+++ b/Tests/Context/ContextTest.php
@@ -30,7 +30,7 @@ class ContextTest extends TestCase
     public function testDefaultValues()
     {
         $this->assertEquals([], $this->context->getAttributes());
-        $this->assertEquals(null, $this->context->getGroups());
+        $this->assertNull($this->context->getGroups());
     }
 
     public function testAttributes()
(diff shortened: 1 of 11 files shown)


Tests/View/ViewTest.php
```diff
@@ -46,7 +46,7 @@ public function testSetLocation()
 
         $view = View::createRedirect($url, $code);
         $this->assertAttributeEquals($url, 'location', $view);
-        $this->assertAttributeEquals(null, 'route', $view);
+        $this->assertAttributeEmpty('route', $view);
```
xabbuh: I am not sure about this. Empty is not necessarily the same as `null`.

carusogabriel: Gonna rollback this changes so :+1:

xabbuh:
Thank you @carusogabriel.

PR #1822, opened 2017-12-18
Fixed use statement for symfony 4 compatibility

MonsieurBon:
The @Route annotations is in a different namespace for symfony 4. I guess this will break compatibility with symfony < 4, so I'm not sure what is the best way to handle this. But currently, this fixes

`In Route.php line 21:
Attempted to load class "Route" from namespace "Sensio\Bundle\FrameworkExtraBundle\Configuration".                                                      
Did you forget a "use" statement for e.g. "Symfony\Component\Routing\Annotation\Route" or "Symfony\Component\Routing\Route"?`

for me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+2.4.0
+-----
+
+* [BC BREAK] The `@Route` annotation and all its children no longer extend SensioFrameworkExtraBundle's annotation.
+  The main effect is that `@Route::$service` is no longer available.
+
 2.3.1
 -----
 
--- a/Controller/Annotations/Route.php
+++ b/Controller/Annotations/Route.php
@@ -11,7 +11,7 @@
 
 namespace FOS\RestBundle\Controller\Annotations;
 
-use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route as BaseRoute;
+use Symfony\Component\Routing\Annotation\Route as BaseRoute;
 
 /**
  * Route annotation class.
--- a/Resources/doc/annotations-reference.rst
+++ b/Resources/doc/annotations-reference.rst
@@ -128,5 +128,5 @@ Example syntax:
     * @Route("", condition="context.getMethod() in ['GET', 'HEAD'] and request.headers.get('User-Agent') matches '/firefox/i'")
     */
 
-.. _`@Route Symfony annotation`: http://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/annotations/routing.html
+.. _`@Route Symfony annotation`: https://symfony.com/doc/current/routing.html
 .. _`Routing Conditions`: http://symfony.com/doc/current/book/routing.html#book-routing-conditions


GuilhemN:
I think we should define the class conditionnally :
```php
if (class_exists(LegacyBaseRoute::class)) {
    class Route extends LegacyBaseRoute {
        // ...
    }
} else {
    class Route extends BaseRoute {
        // ...
    }
}

**Comment by codereviewvideos** [comment]
```
<?php

/*
 * This file is part of the FOSRestBundle package.
 *
 * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
 *
 * For the full copyright and license information, please view the LICENSE
 */

use Sensio\Bundle\FrameworkExtraBundle\Configuration\Route as LegacyBaseRoute;
use Symfony\Component\Routing\Annotation\Route as Symfony4BaseRoute;

if (class_exists(LegacyBaseRoute::class)) {
    class BaseRoute extends LegacyBaseRoute {}
} else {
    class BaseRoute extends Symfony4BaseRoute {}
}

/**
 * Route annotation class.
 *
 * @Annotation
 */
class Route extends BaseRoute
{
    {
        parent::__construct($data);

        if (!$this->getMethods()) {
            $this->setMethods((array) $this->getMethod());
        }
    }

    /**
     * @return string|null
     */
    public function getMethod()
    {
        return;
    }
}
```

Not the most elegant looking thing I've ever seen, but it works (I think). Thoughts on this?

**Comment by GuilhemN** [comment]
That's fine for me, this is a temporary fix until we drop sf < 4 support :+1:

**Comment by sarcher** [comment]
This is now causing a deprecation warning with SensioFrameworkExtraBundle 5.2 (see sensiolabs/SensioFrameworkExtraBundle#562). The conditional fix above works; maybe it could be made into its own PR or somebody updates this one so it can be merged?

**Comment by codereviewvideos** [comment]
Sorry for the delay. I have submitted a PR with the suggested interim fix.

**Comment by xabbuh** [comment]
@GuilhemN Is that really necessary? I don't see anything that requires us to rely on this switch.

**Comment by GuilhemN** [comment]
@xabbuh you mean we should simply switch to symfony annotation ? That's a bc break I'd rather avoid.

Or do you mean we don't need to change the base annotation? Thing is our routing annotations can be used without sensioframeworkbundle but they depend on it right now which causes the mentionned error

**Comment by Tobion** [comment]
The only potential BC break by changing the parent I see is for people who used the Rest Route but loaded it with the FrameworkExtraBundle instead of the Rest Route loader: https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/6966aee1bfc020d6683156784d0c78c365d0f83b/Routing/AnnotatedRouteControllerLoader.php#L45
This "service" thing won't work anymore. But it seems like an edge case and a note in the changelog about the changed parent class should be enough.

**Comment by xabbuh** [comment]
I checked the history of the annotation and this was indeed changed on intent in the past to allow controllers to be services (see #604 and #646). I think if we were to make this change, we should bump requirements to at least 3.4 at the same time. An alternative solution would mean to ship our own annotation route loader and deprecate the `service` option there. That's probably a bit too much here.

**Comment by Tobion** [comment]
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Resources/doc/annotations-reference.rst#route also needs an update as it links to FrameworkExtraBundle

File: CHANGELOG.md [hunk 1]
```diff
@@ -1,6 +1,12 @@
 CHANGELOG
 =========
 
+2.4.0
+-----
+
+* [BC BREAK] The `@Route` annotation and all its children no longer extend SensioFrameworkExtraBundle's annotation.
+  The main effect is that `@Route::$service` is no longer available.
```
xabbuh: We should mention that you should use FCQN as service ids instead. And we should also update the branch alias in the `composer.json` file.

karousn:
So we should install SensioFrameworkExtraBundle until this PR not merged ?

GuilhemN:
@karousn if you want to use fosrest routing, yes. It has always been the case so far.

CHANGELOG.md
GuilhemN: Thanks, sorry I forgot to take care of this :)

umpirsky:
When can we expect to see this land in 2.4.0?

PR #1828, opened 2018-01-03
fix tests depending on the TwigBundle

xabbuh:


xabbuh:
see sensiolabs/SensioFrameworkExtraBundle#441

PR #1831, opened 2018-01-03
Fix tests

XWB:


.travis.yml
```diff
@@ -29,6 +29,7 @@ matrix:
       env: DEPENDENCIES='dev'
 
 before_install:
+  - echo "memory_limit=4G" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
```
xabbuh: looks like we can *now* use the `COMPOSER_MEMORY_LIMIT` var for this (see https://github.com/composer/composer/issues/6931#issuecomment-355240659)

Tests/Functional/app/Configuration/config.yml
```diff
@@ -32,6 +32,7 @@ security:
     firewalls:
         default:
             anonymous: ~
+            logout_on_user_change: true
```
xabbuh: this will likely fail with pre 3.4 Symfony versions

XWB: I guess there is no easy fix for this one.

.travis.yml
XWB: That's new to me, I'll have a look.

Tests/Functional/app/Configuration/config.yml
xabbuh: I'm gonna take a look into it.

.travis.yml
xabbuh: Looks like we sadly cannot use it now as the current release has some autoloading issues (see #1833). So let's stick with your approach for the moment.

.travis.yml
```diff
@@ -23,12 +23,12 @@ matrix:
       env: SYMFONY_LTS='^2'
     # Symfony 3 LTS
     - php: 7.0
-      env: SYMFONY_LTS='^3' DEPENDENCIES='dev'
+      env: SYMFONY_LTS='^3'
     # development dependencies
     - php: 7.1
-      env: DEPENDENCIES='dev'
```
xabbuh: We should keep this job, but add a dedicated job that uses PHP 7.1 with stable dependencies (it will automatically use Symfony 4 then right now).

XWB:
@xabbuh Any clue why the PHP 5.5 test fails?

> 1) FOS\RestBundle\Tests\Functional\SerializerErrorTest::testSerializeExceptionJson with data set #1 ('JMSSerializer')
Doctrine\Common\Annotations\AnnotationException: [Semantical Error] The annotation "@template" in class PhpCollection\AbstractMap was never imported. Did you maybe forget to add a "use" statement for this annotation?

> 2) FOS\RestBundle\Tests\Functional\SerializerErrorTest::testSerializeExceptionXml with data set #1 ('JMSSerializer', '<?xml version="1.0" encoding=...sult>\n')
Doctrine\Common\Annotations\AnnotationException: [Semantical Error] The annotation "@template" in class PhpCollection\AbstractMap was never imported. Did you maybe forget to add a "use" statement for this annotation?

As far as I can see, `sensio/framework-extra-bundle` is installed.

xabbuh:
@XWB Not yet, I started to investigate locally, but didn't come to a solution right now.

XWB:
@xabbuh Current status:

- [ ] PHP 5.5 test fails due to error above. No clue what's going on.
- [ ]  PHP 5.6 fixed by increasing memory limit.
- [ ] Symfony LTS 2 tests hangs, reason unknown.
- [ ] Symfony LTS 3 fixed by increasing memory limit.
- [ ] Symfony 4 test works.
- [ ] Symfony dev test fails. I moved it to `allow_failes` until somebody else can fix it.

xabbuh:
Thank you @XWB.

xabbuh:
@XWB FYI, the `deps=low` tests broke in #1805. Changing the type of the compiler pass again would make the test green, but that would of course bring back the issue that was fixed there.

XWB:
@xabbuh Symfony 2 test is broke due to the following error:

> 1) FOS\RestBundle\Tests\Functional\SerializerErrorTest::testSerializeExceptionJson with data set #1 ('JMSSerializer')
Doctrine\Common\Annotations\AnnotationException: [Semantical Error] The annotation "@dataProvider" in method FOS\RestBundle\Tests\Functional\SerializerErrorTest::testSerializeExceptionJson() was never imported. Did you maybe forget to add a "use" statement for this annotation?

Any clue?

PR #1838, opened 2018-01-08
Fix dependency to sensio/framework-extra-bundle

rokkie:
\FOS\RestBundle\Request\ParameterBag has a dependency to
\Doctrine\Common\Util\ClassUtils which originates from doctrine/common
which in turn a dependency of sensio/framework-extra-bundle but this is
declared as a dev-dependency. This causes an fatal error with the
message 'Class 'Doctrine\Common\Util\ClassUtils' not found'.

Fixes#259

### DIFF
--- a/Request/ParameterBag.php
+++ b/Request/ParameterBag.php
@@ -11,6 +11,7 @@
 
 namespace FOS\RestBundle\Request;
 
+use Doctrine\Common\Util\ClassUtils;
 use FOS\RestBundle\Controller\Annotations\ParamInterface;
 use Symfony\Component\HttpFoundation\Request;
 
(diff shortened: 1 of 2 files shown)


xabbuh:
I would rather add `doctrine/common` as a dependency then because that is what we need here.

rokkie:
Changed the dependency to `doctrine/common`.

xabbuh:
Thinking about this again, I wonder if we just modify the code instead. I mean we do not need to use the `ClassUtils::getClass()` method if Doctrine is not installed. In these cases, just using `get_class()` should be sufficient.

Request/ParameterBag.php
```diff
@@ -11,7 +11,6 @@
 
 namespace FOS\RestBundle\Request;
 
-use Doctrine\Common\Util\ClassUtils;
```
xabbuh: this should be reverted

Request/ParameterBag.php
```diff
@@ -76,8 +77,12 @@ private function initParams($requestId)
             );
         }
 
+        $cls = class_exists('\Doctrine\Common\Util\ClassUtils', true)
```
xabbuh: you can use `class_exists(ClassUtils::class)` here then

GuilhemN:
Thank you @rokkie!

PR #1840, opened 2018-01-11
make the view handler alias public

xabbuh:
The alias will be fetched from the container when using the built-in
FOSRestController class. Thus, we must ensure that it can still be
located when using Symfony 4.0 or higher.

### DIFF
--- a/DependencyInjection/FOSRestExtension.php
+++ b/DependencyInjection/FOSRestExtension.php
@@ -12,6 +12,7 @@
 namespace FOS\RestBundle\DependencyInjection;
 
 use Symfony\Component\Config\FileLocator;
+use Symfony\Component\DependencyInjection\Alias;
 use Symfony\Component\DependencyInjection\ChildDefinition;
 use Symfony\Component\DependencyInjection\Compiler\ServiceLocatorTagPass;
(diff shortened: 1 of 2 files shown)


XWB:
@xabbuh Any timeline for a new release?

smalot:
This issue is blocking my current Symfony upgrade to 3.4.4

HeyIAmBob:
Thanks @xabbuh for the fix in DependencyInjection/FOSRestExtension.php . 
Do you know when the next release is planned ?

The Rest Bundle does not work with Symfony4 without this fix.
With your fix (applied manually), it works like a charm

XWB:
@HeyIAmBob Next release is almost ready, @xabbuh just needs to add a tag to https://github.com/FriendsOfSymfony/FOSRestBundle/releases :)

HeyIAmBob:
That’s a good news !
Thanks for all

xabbuh:
It's probably going to be published by the end of this week.

smalot:
Any news ?

xabbuh:
2.3.1 was published a few days ago.

smalot:
Excellent !
Many thanks

PR #1842, opened 2018-01-11
don't set logout_on_user_change on Symfony 4.1

xabbuh:


xabbuh:
The remaining failures will be fixed by symfony/symfony#25471.

PR #1743, opened 2017-07-25
Improve the docs

GuilhemN:
I'd like to make the docs less boring to read and describing less edge cases.
So here a first step :)

PR #1852, opened 2018-01-26
[docs] Symfony 4.0 bundle registration

Idziak:
In Symfony 4.0 there are a new bundle registration procedure

### DIFF
--- a/Resources/doc/1-setting_up_the_bundle.rst
+++ b/Resources/doc/1-setting_up_the_bundle.rst
@@ -16,8 +16,11 @@ in the `installation chapter`_ of the Composer documentation.
 
 B) Enable the Bundle
 --------------------
+Then, enable the bundle.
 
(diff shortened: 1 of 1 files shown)


xabbuh:
If you are using Symfony Flex, the `config/bundles.php` file does not have to be changed manually. I would omit that part and just reword the existing paragraph to explain that the step is only necessary if you do not use Flex.

GuilhemN:
I agree with @xabbuh :+1:

GuilhemN:
I added a ``caution`` for users not using Flex. Does it look good to you?

GuilhemN:
We may also merge https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1743 directly to avoid merge conflicts.

GuilhemN:
Thank you @Idziak!

PR #1870, opened 2018-02-21
Fix exception controller

gdbonino:


Controller/TwigExceptionController.php
```diff
@@ -42,27 +41,27 @@ protected function findTemplate(Request $request, $statusCode, $showException)
         $format = $request->getRequestFormat();
 
         $name = $showException ? 'exception' : 'error';
-        if ($showException && 'html' === $format) {
+        if ($showException && 'html' == $format) {
```
xabbuh: What is the reason for this change?

xabbuh:
I am not sure I understand why this change is necessary. Can we create a test case that proves that these changes are indeed needed?

gdbonino:
As the comment on  [TwigExceptionController.php](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/64ef743cb90a9c239ad1c45172678ca39b942ee1/Controller/TwigExceptionController.php#L40) explicitly says, this method should be synchronized on a regular basis against TwigBundle's version.

Since TwigBundle's method has been significatively modified FOSRest cannot longer find the local override template for custom error messages. If you disable FOSRest, the template works as expected.

The only thing I did on this PR is to replicate existing TwigBundle's code so FOSRest can find and display the template correctly.

If you need a test case, let me know and I'll try to build and share an example so you can see the problem in action.

GuilhemN:
Thank you @gdbonino.

Issue #1851, opened 2018-01-25
Recover JMS Serializer handlers


ushatel:
Hello, Friends Of Symfony, 

Please, check this thing when this is possible.  

I am using JMS Serializer in my project with the extension jms_serializer.subscribing_handler registered 
 that allows me to make specific transformation for the type. 

After making composer updates of the vendors it won't work. 

Making some analyzes have found that the mistake rised from the \vendor\friendsofsymfony\rest-bundle\FOSRestBundle.php 

There were changes in that file  

>>> Was 

`
$container->addCompilerPass(new JMSHandlersPass(), PassConfig::TYPE_REMOVE); 
`

>>> 

>>> After upgrade 

`
$container->addCompilerPass(new JMSHandlersPass(), PassConfig::TYPE_BEFORE_REMOVING, -10); 
`

>>> 

It looks the sources of the changes are here 
https://github.com/FriendsOfSymfony/FOSRestBundle/commit/0f85a17149028f38708aca926e7eb8bbd639ca1d 
https://github.com/FriendsOfSymfony/FOSRestBundle/commit/f144dbbea507c2bca3753b96f8cf28ed82ddf550 

If I revert changes to the \vendor\friendsofsymfony\rest-bundle\FOSRestBundle.php all things goes well. 

Please, help to solve this problem. 

Thank you for the attention and wish you good time!

xabbuh:
Can you explain what exactly does not work?

ushatel:
The handler that installed with the  jms_serializer.subscribing_handler could not handle the type after downloading the latest FOS. When I revert changes to the previous version handle works well on the same code. 

Thank you for the attention.

xabbuh:
Sorry, but with this little information I am afraid I will not be able to help you. Maybe it's easier to explain your problem by pushing a small example project that shows what is not working for you.

ushatel:
Thank you for the answer. I will prepare the project as I could rise the time quote.

naitsirch:
Same problem here. It caused me a lot of headache today. After an update of FosRestBundle, the compiled service container builds the `jms_serializer.handler_registry` this way:

```php
protected function getJmsSerializer_HandlerRegistryService()
{
    return $this->services['jms_serializer.handler_registry'] = new \FOS\RestBundle\Serializer\JMSHandlerRegistry(
        new \JMS\Serializer\Handler\LazyHandlerRegistry($this),
[...]

PR #1873, opened 2018-02-27
manually decorate the core JMS handler registry

xabbuh:
This works around the fact that custom handlers are processed in
JMSSerializerBundle after Symfony's DecoratorServicePass has been
executed (which means that we cannot use proper service decoration).

The old code still used to work on Symfony 2.x applications where
FOSRestBundle was registered after JMSSerializerBundle. In those cases,
custom handlers were processed before the handler registry was replaced
by the compiler pass coming from FOSRestBundle.

More recent Symfony applications have not been affected by this bug as
the compiler pass priority would have ensured the correct order in which
the passes would have been executed.

### DIFF
--- a/DependencyInjection/Compiler/HandlerRegistryDecorationPass.php
+++ b/DependencyInjection/Compiler/HandlerRegistryDecorationPass.php
@@ -1,0 +1,52 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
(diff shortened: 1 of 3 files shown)


XWB:
Nice find @xabbuh 

Just one more test needs to be resolved:

> FOS\RestBundle\Tests\Functional\DependencyInjectionTest::testSerializerRelatedServicesAreNotRemovedWhenJmsSerializerBundleIsEnabled
Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException: The "test.jms_serializer.handler_registry" service or alias has been removed or inlined when the container was compiled. You should either make it public, or stop using the container directly and use dependency injection instead.

xabbuh:
@XWB here we go 🎉

XWB:
All green :tada:

Tobion:
Does this fix #1851?

xabbuh:
@Tobion yes, updating to 2.3.1 should help

PR #1877, opened 2018-03-07
JMSFormErrorHandlerPass checks before setting parameter

### LINKED ISSUE #1806: JMSFormErrorHandlerPass disables overwriting class parameter jms_serializer.form_error_handler.class
In #1742 the JMSFormErrorHandlerPass forces the `jms_serializer.form_error_handler.class` to the one from this bundle. I previously have overwritten this parameters with my own hander:
``` yaml
parameters:
```
This is now ignored/overruled by the CompilerPass.

You can fix it by completely overwriting the service definition:
``` yaml
services:
    jms_serializer.form_error_handler:
        class: 'MyBundle\Serializer\Handler\FormErrorHandler'
        arguments: ['@translator']
        tags: [{ name: 'jms_serializer.subscribing_handler' }]
```
Or write your own CompilerPass, but that seems a bit too much. 

Is this considered to be a BC break? And should we change the implementation of the compiler pass to first check if the parameter is set to the default value of JMS, and only then overwrite it?

I can create a PR for this solution and Im also open to other suggestions.

GijsL:
Only overwrite parameter if value is set to default to allow custom implementations.

Fixes #1806 

Please review and provide feedback if needed.


.php_cs
```diff
@@ -1,25 +0,0 @@
-<?php
-
-use Symfony\CS\Fixer\Contrib\HeaderCommentFixer;
```
xabbuh: please keep this file

Tests/DependencyInjection/Compiler/JMSFormErrorHandlerPassTest.php
```diff
@@ -0,0 +1,84 @@
...
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+class JMSFormErrorHandlerPassTest extends TestCase
+{
+    private $param = 'jms_serializer.form_error_handler.class';
+
+    public function testParameterNotSetWhenParameterNotFound()
+    {
+        $container = $this->getMockBuilder(ContainerBuilder::class)->getMock();
```
xabbuh: please do not mocker the DI related classes

xabbuh:
I wonder if we shouldn't stop changing the parameter at all and decorate the JMS service instead.

xabbuh:
see #1879

.php_cs
GijsL: Was deleted accidentally.

Tests/DependencyInjection/Compiler/JMSFormErrorHandlerPassTest.php
GijsL: I used the other CompilerPass tests as template. Shoud I use a real instance of ContainerBuilder?

GijsL:
I will update/fix the phpunit tests if this is still a merge-candidate, instead of #1879

@xabbuh @GuilhemN Please advise on how to proceed. I can close this PR if it will not be merged.

GuilhemN:
I just merged @xabbuh's PR, thanks for your PR still!

PR #1879, opened 2018-03-08
prefer decoration over parameter replacements

### LINKED ISSUE #1806: JMSFormErrorHandlerPass disables overwriting class parameter jms_serializer.form_error_handler.class

xabbuh:
Fixes #1806 and replaces #1876 and #1877.

ping @GuilhemN @tarjei @GijsL

### DIFF
--- a/Serializer/Normalizer/FormErrorHandler.php
+++ b/Serializer/Normalizer/FormErrorHandler.php
@@ -13,23 +13,27 @@ namespace FOS\RestBundle\Serializer\Normalizer;
 
 use JMS\Serializer\Context;
 use JMS\Serializer\Handler\FormErrorHandler as JMSFormErrorHandler;
+use JMS\Serializer\Handler\SubscribingHandlerInterface;
 use JMS\Serializer\JsonSerializationVisitor;
 use JMS\Serializer\XmlSerializationVisitor;
 use Symfony\Component\Form\Form;
 use JMS\Serializer\YamlSerializationVisitor;
-use Symfony\Component\Translation\TranslatorInterface;
 
 /**
  * Extend the JMS FormErrorHandler to include more informations when using the ViewHandler.
  */
-class FormErrorHandler extends JMSFormErrorHandler
+class FormErrorHandler implements SubscribingHandlerInterface
 {
-    private $translator;
+    private $formErrorHandler;
 
-    public function __construct(TranslatorInterface $translator)
+    public function __construct(JMSFormErrorHandler $formErrorHandler)
     {
-        $this->translator = $translator;
-        parent::__construct($translator);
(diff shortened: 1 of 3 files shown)


tarjei:
Looks good. One question: How should implementers use this. I.e. how should existing implementations using the class param be changed?

GijsL:
Not sure how this will work with my implementation: I have a completely customized `serializeFormToJson` function, which will be overwritten by the decorator, correct?

Serializer/Normalizer/FormErrorHandler.php
```diff
@@ -86,6 +90,11 @@ public function serializeFormToYml(YamlSerializationVisitor $visitor, Form $form
         return $result;
     }
 
+    public function __call($name, $arguments)
+    {
+        return call_user_func_array([$this->formErrorHandler, $name], $arguments);
```
GuilhemN: Is this for bc? If so I think we should deprecate these calls.

GuilhemN:
@tarjei @GijsL if you want to completely overwrite fos rest' handler, you may use jms class parameter and remove the service from fos rest. If you only want to do small changes, with @xabbuh's change, jms class parameter will work again and will overwrite jms handler which will then still be decorated by fos rest handler.

Serializer/Normalizer/FormErrorHandler.php
xabbuh: I decided to do this since the `FormErrorHandler` from JMSSerializerBundle does not implement an interface and I don't want to us to have to update our class every time a change in JMSSerializerBundle is made (i.e. new methods are added).

GijsL:
Ok, I understand. This was considered a BC break, and this will not revert that BC break. 

I prefer #1877 but I will not update/fix the unit tests if the plan is to merge this PR instead.

GuilhemN:
Thank you @xabbuh (and sorry for the delay).

PR #1891, opened 2018-05-10
Update controller prefix if it is defined as service with class name as id

gpekz:
When a controller is defined as service, for example when using the `controller.service_arguments` tag, `$request->attributes->get('_controller')` looks like this `App\MyController:myAction`.

The goal of the PR is to have `App\MyController::myAction` instead. IMHO I doesn't make sense that the `_controller` parameter is different depending on whether it is defined as a service or not. Unless there is a reason but I do not see which one.

### DIFF
--- a/Routing/Loader/RestRouteLoader.php
+++ b/Routing/Loader/RestRouteLoader.php
@@ -116,7 +116,8 @@ class RestRouteLoader extends Loader
 
         if ($this->container->has($controller)) {
             // service_id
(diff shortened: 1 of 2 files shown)


gpekz:
@GuilhemN

GuilhemN:
We can't do that because of BC, this possibility was introduced in 3.3 (or 3.4 maybe) but we still support 2.7. Thanks for the proposal still!

PR #1894, opened 2018-05-21
Add the possibility to resolve the controller name

gpekz:
Controller name looks like `App\MyController:myAction` when a controller is defined as a service.
The goal of this PR is to add the possibility to resolve the controller name in order to have `App\MyController::myAction`

#1891


GuilhemN:
Sorry I closed #1891 so quickly... It was on the right path, I just think instead of always using `::` class named services it should only be done when using sf>3.4 (this can be detected from a new class or new method introduced in the minor).

Tobion:
This is fixed with #1904. When using sf 4.1 it uses `::`.

PR #1895, opened 2018-05-22
Fix the tests

GuilhemN:


composer.json
```diff
@@ -57,7 +57,7 @@
         "symfony/web-profiler-bundle": "^2.7|^3.0|^4.0",
         "symfony/twig-bundle": "^2.7|^3.0|^4.0",
         "symfony/browser-kit": "^2.7|^3.0|^4.0",
-        "symfony/dependency-injection": "^2.7|^3.0|^4.0",
+        "symfony/dependency-injection": "^2.7.20|^3.0|^4.0",
```
GuilhemN: Tags were inherited before it thus making the ``FormErrorHandler`` not working.

Issue #1897, opened 2018-06-08
Referencing controllers with a single colon is deprecated since Symfony 4.1 when executing PHPUnit


jguyenot:
Hi,

since we are updated to Symfony 4.1 we have this error when launching 
`./vendor/bin/simple-phpunit`

Error message
```
  1x: Referencing controllers with a single colon is deprecated since Symfony 4.1. Use App\Core\User\Controller\UserGroupController::deleteAction instead.
    1x in ContainerTypeTest::testCRUD from App\Tests\Core\Container\Controller
```

I've check into the 
`vendor/friendsofsymfony/rest-bundle/Routing/Loader/RestRouteLoader.php`
and i apply a correction.

`Line 119 :: $prefix = $controller.':'; >> $prefix = $controller.'::';`
`Line 133 :: } elseif (false !== strpos($controller, ':')) { >> } elseif (false !== strpos($controller, '::')) {`

Could you check this correction to apply it into the next release?

pdugas:
I think we also need a fix for the same issue in `Resources/config/exception_listener.xml` where it's referencing `showAction`.  Add the colon there too.

pdugas:
I can confirm the changes to RestRouteLoader.php that @jguyenot proposed resolved the issue for me.

xabbuh:
Did you check if that still works with Symfony 3.4?

jguyenot:
Hi @xabbuh , it works fine with Symfony 4.1 but they deprecated the annotation with only one ":".

GuilhemN:
Your fix won't work, using 3.x an exception will be thrown: https://github.com/symfony/symfony/blob/3.3/src/Symfony/Component/HttpKernel/Controller/ControllerResolver.php#L180.

We need to change the syntax conditionally.

PR #1898, opened 2018-06-11
Handle SF4.1 Deprecation

pdugas:
Resolves this issue

`Referencing controllers with a single colon is deprecated since Symfony 4.1. Use fos_rest.exception.twig_controller::showAction instead.`

### DIFF
--- a/Resources/config/exception_listener.xml
+++ b/Resources/config/exception_listener.xml
@@ -8,7 +8,7 @@
         <service id="fos_rest.exception_listener" class="FOS\RestBundle\EventListener\ExceptionListener">
             <tag name="kernel.event_subscriber" />
             <tag name="monolog.logger" channel="request" />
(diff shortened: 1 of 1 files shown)


pdugas:
Should address second half of #1897

xabbuh:
Did any of you test this change with older Symfony versions? Does it work there?

pdugas:
I've only tested the change on my current 4.1 project. Don't have an older one readily available. Sorry.

XWB:
This won't work on older Symfony versions.

GuilhemN:
Should be done conditionally as in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1904.

xabbuh:
I am closing here as the issue was fixed with #1911. Thank you for the initial work @pdugas.

PR #1903, opened 2018-06-29
kebab-case key normalizer

oleg-andreyev:
[JSON API Specification](http://jsonapi.org/format/) is recommending to use hyphen minus (U+002D HYPHEN-MINUS, “-“) as separator between multiple words

Adding `KebabKeysNormalizer` which will normalize kebab-case into camelCase

### DIFF
--- a/Normalizer/KebabKeysNormalizer.php
+++ b/Normalizer/KebabKeysNormalizer.php
@@ -11,7 +11,7 @@
 namespace FOS\RestBundle\Normalizer;
 
 /**
- * Normalizes the array by changing its keys from kebab-case to camel case
+ * Normalizes the array by changing its keys from kebab-case to camel case.
  *
  * @author Oleg Andreyev <oleg@andreyev.lv>
  */
--- a/Normalizer/KebabKeysNormalizer.php
+++ b/Normalizer/KebabKeysNormalizer.php
@@ -10,60 +10,13 @@
 
 namespace FOS\RestBundle\Normalizer;
 
-use FOS\RestBundle\Normalizer\Exception\NormalizationException;
-
 /**
- * Normalizes the array by changing its keys from kebab case to camel case.
+ * Normalizes the array by changing its keys from kebab-case to camel case
  *
- * @author Oleg Andreyev <oleg.andreyev@live.com>
+ * @author Oleg Andreyev <oleg@andreyev.lv>
  */
-class KebabKeysNormalizer implements ArrayNormalizerInterface
+class KebabKeysNormalizer extends AbstractKeysNormalizer
 {
-    /**
-     * {@inheritdoc}
-     */
-    public function normalize(array $data)
-    {
-        $this->normalizeArray($data);
-
-        return $data;
-    }
-
-    /**
-     * Normalizes an array.
-     *
-     * @param array &$data
-     *
-     * @throws Exception\NormalizationException
-     */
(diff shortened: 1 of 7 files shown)


Tests/Normalizer/KebabKeysNormalizerTest.php
```diff
@@ -0,0 +1,67 @@
...
+        $array = $this->normalizeProviderCommon();
+        $array[] = array(
+            array('_-username' => 'foo', '-password' => 'bar', '-foo-bar' => 'foobar'),
+            array('_Username' => 'foo', 'Password' => 'bar', 'FooBar' => 'foobar'),
+        );
+
+        return $array;
+    }
+
+    private function normalizeProviderCommon()
```
Tobion: what's the point of this method it's only used once?

Normalizer/KebabKeysNormalizer.php
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * Normalizes an array.
+     *
+     * @param array &$data
+     *
+     * @throws Exception\NormalizationException
+     */
+    private function normalizeArray(array &$data)
```
Tobion: Why do you pass data by reference if you don't change data by reference at all but use temporary value `$normalizedData`. No need to pass data by reference then but just return it here.

Normalizer/KebabKeysNormalizer.php
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /**
+     * Normalizes a string.
+     *
+     * @param string $string
+     *
+     * @return string
+     */
+    protected function normalizeString($string)
```
Tobion: should be private. and please add param and return types as we require php 7.1 now.

oleg-andreyev: I actually have taken CamelKeysNormalizer as class-skeleton.

Tobion: I see. But it would still need to be changed now.

oleg-andreyev: Since it was quite awhile ago, I've updated this PR and addressed your request.

Tests/Normalizer/KebabKeysNormalizerTest.php
oleg-andreyev: removed.

Normalizer/KebabKeysNormalizer.php
oleg-andreyev: it's actually is changed by reference at the end of the method `$data = $normalizedData;`

Tests/Normalizer/KebabKeysNormalizerTest.php
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * @dataProvider normalizeProvider
+     */
+    public function testNormalize(array $array, array $expected)
+    {
+        $normalizer = new KebabKeysNormalizer();
+        $this->assertEquals($expected, $normalizer->normalize($array));
+    }
+
+    public function normalizeProvider()
```
Tobion: please add iterable return type and you can use `yield`

xabbuh:

oleg-andreyev:
@xabbuh sorry, I somehow missed recent comments to this PR. I'll re-open this PR after I'll rebase it on 2.x

PR #1904, opened 2018-07-01
Fix deprecations

GuilhemN:
Fixes https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1897

PR #1907, opened 2018-07-05
Simplify version check

GuilhemN:
Fixes https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1904#discussion_r199572760

### DIFF
--- a/Tests/Routing/Loader/RestYamlCollectionLoaderTest.php
+++ b/Tests/Routing/Loader/RestYamlCollectionLoaderTest.php
@@ -219,7 +219,7 @@ class RestYamlCollectionLoaderTest extends LoaderTest
 
         // We check that it's "controller:method" if sf < 4.1 (controller as service) and not "controller::method"
         $this->assertEquals(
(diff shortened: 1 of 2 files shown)


Tests/Routing/Loader/RestYamlCollectionLoaderTest.php
```diff
@@ -219,7 +219,7 @@ public function testControllerAsServiceWithClassName()
 
         // We check that it's "controller:method" if sf < 4.1 (controller as service) and not "controller::method"
         $this->assertEquals(
-            UsersController::class.(!method_exists(Kernel::class, 'getAnnotatedClassesToCompile') ? ':' : '::').'getUsersAction',
+            UsersController::class.(Kernel::VERSION_ID >= 40100 ? '::' : ':').'getUsersAction',
```
xabbuh: Is this actually depending on the version of the HttpKernel component? It depends on the Routing component being 4.1 or higher, doesn't it?

GuilhemN: It depends on the httpkernel component because the `ContainerControllerResolver` belongs to the component :slightly_smiling_face:

PR #1910, opened 2018-07-12
Fixes use statement of Route annotation for symfony 4 compatibility

codereviewvideos:
PR for https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1822


GuilhemN:
Indeed the deprecation is still triggered but there is a workaround and this solution doesn't introduce a BC break.
But you're right, #1822 is probably better.

xabbuh:
Closing in favour of #1822. Thanks for working on this issue @codereviewvideos.

PR #1911, opened 2018-07-12
Fix exception controller deprecations for Symfony >= 4.1

sarcher:
This extends the work demonstrated in #1898. It's a little tricky because the old-style controller definitions appear in two places, so this accounts for both of them in one spot.

Note that there are a ton of other deprecation notices in the test suite itself, but this fixes only the ones related to the exception controller definition. Many of them could be resolved by loading a different TestBundle routing file for Symfony >= 4.1; I wasn't sure how you'd prefer to handle that.

### DIFF
--- a/DependencyInjection/FOSRestExtension.php
+++ b/DependencyInjection/FOSRestExtension.php
@@ -21,8 +21,9 @@ use Symfony\Component\DependencyInjection\Loader\XmlFileLoader;
 use Symfony\Component\DependencyInjection\Reference;
 use Symfony\Component\Form\AbstractType;
 use Symfony\Component\Form\Extension\Core\Type\FormType;
-use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\HttpKernel\DependencyInjection\Extension;
+use Symfony\Component\HttpKernel\Kernel;
 
 class FOSRestExtension extends Extension
 {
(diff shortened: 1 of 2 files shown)


DependencyInjection/FOSRestExtension.php
```diff
@@ -351,6 +352,11 @@ private function loadException(array $config, XmlFileLoader $loader, ContainerBu
                 $container->getDefinition('fos_rest.exception_listener')->replaceArgument(0, 'fos_rest.exception.twig_controller:showAction');
             }
 
+            // rewrite default exception controller definition for Symfony >= 4.1; can come from above or from the xml file
+            if (Kernel::VERSION_ID >= 40100 && preg_match('/^fos_rest\.exception\.(twig_)*controller:showAction$/', $container->getDefinition('fos_rest.exception_listener')->getArgument(0), $matches) > 0) {
```
xabbuh: I would merge block with the previous `if` instead to have something like this:

```php
if (Kernel::VERSION_ID >= 40100) {
    $controller = 'fos_rest.exception.controller::showAction';
} else {
    $controller = 'fos_rest.exception.controller:showAction';
}

if ($config['exception']['exception_controller']) {
    $controller = $config['exception']['exception_controller'];
} elseif (isset($container->getParameter('kernel.bundles')['TwigBundle'])) {
    if (Kernel::VERSION_ID >= 40100) {
        $controller = 'fos_rest.exception.twig_controller::showAction';
    } else {
        $controller = 'fos_rest.exception.twig_controller:showAction';
    }
}

$container->getDefinition('fos_rest.exception_listener')->replaceArgument(0, $controller);
```

Then we can set the argument to `null` in the `exception_listener.xml` file:

```xml
<service id="fos_rest.exception_listener" class="FOS\RestBundle\EventListener\ExceptionListener">
    <tag name="kernel.event_subscriber" />
    <tag name="monolog.logger" channel="request" />
    <argument />
    <argument type="service" id="logger" on-invalid="null" />
</service>
```

sarcher: Sure thing, I can do that. I wasn't clear on why that was defaulted in both places, I'd say moving it out of the XML would be a good idea anyway.

xabbuh: That's because the coffee in the extension class uses the `twig_` prefix. :)

xabbuh:
Thank you @sarcher.

PR #1912, opened 2018-07-13
Test the presence of OptionsResolver when the body_converter validation option is enabled

2binfree:


DependencyInjection/Configuration.php
```diff
@@ -110,7 +110,20 @@ public function getConfigTreeBuilder()
                 ->arrayNode('body_converter')
                     ->canBeEnabled()
                     ->children()
-                        ->scalarNode('validate')->defaultFalse()->end()
+                        ->scalarNode('validate')
+                            ->defaultFalse()
+                            ->beforeNormalization()
+                                ->ifTrue()
+                                ->then(function() {
+                                    if (!class_exists('Symfony\Component\OptionsResolver\OptionsResolver')) {
```
xabbuh: let's use the `class` constant here

DependencyInjection/Configuration.php
```diff
@@ -110,7 +110,20 @@ public function getConfigTreeBuilder()
...
+                                        throw new InvalidConfigurationException(
```
xabbuh: please just place this on a single line

2binfree: Ok but the line is going to be very long.
[There MUST NOT be a hard limit on line length; the soft limit MUST be 120
characters; lines SHOULD be 80 characters or less.](https://www.php-fig.org/psr/psr-2/)

GuilhemN: We consider exception messages as exceptions :)

DependencyInjection/Configuration.php
```diff
@@ -110,7 +111,17 @@ public function getConfigTreeBuilder()
...
                     ->canBeEnabled()
                     ->children()
-                        ->scalarNode('validate')->defaultFalse()->end()
+                        ->scalarNode('validate')
+                            ->defaultFalse()
+                            ->beforeNormalization()
+                                ->ifTrue()
+                                ->then(function () {
+                                    if (!class_exists(OptionsResolver::class)) {
+                                        throw new InvalidConfigurationException("'body_converter.validate: true' require OptionsResolver component installation ( composer require symfony/options-resolver )");
```
GuilhemN: `requires`

GuilhemN:
Thank you @2binfree!

DependencyInjection/Configuration.php
```diff
@@ -110,7 +111,17 @@ public function getConfigTreeBuilder()
                 ->arrayNode('body_converter')
                     ->canBeEnabled()
                     ->children()
-                        ->scalarNode('validate')->defaultFalse()->end()
+                        ->scalarNode('validate')
+                            ->defaultFalse()
+                            ->beforeNormalization()
+                                ->ifTrue()
```
xabbuh: should be `always()` (see #1918)

goetas:
@GuilhemN can I ask why this was merged if it was breaking the test suite?

GuilhemN:
@goetas oops, I thought all test failures were unrelated (I didn't check all cases 😶)... It proves we need to fix our test suite :)

PR #1915, opened 2018-07-14
use forward compatible controller reference syntax

xabbuh:


Tests/Functional/Bundle/TestBundle/Resources/config/routing.php
```diff
@@ -0,0 +1,73 @@
+<?php
+
+use Symfony\Component\HttpKernel\Kernel;
+use Symfony\Component\Routing\Route;
+use Symfony\Component\Routing\RouteCollection;
+
+$routes = new RouteCollection();
+
+$routes->add('request_body_param_converter', new Route('/body-converter', array(
+    '_controller' => Kernel::VERSION_ID >= 40100 ? 'FOS\RestBundle\Tests\Functional\Bundle\TestBundle\Controller\RequestBodyParamConverterController::putPostAction' : 'TestBundle:RequestBodyParamConverter:putPost',
```
Tobion: I don't think you need to use the old syntax `TestBundle:RequestBodyParamConverter:putPost` at all. The controllers are not services, so the FQCN::action should work with SF 4.1 and older versions as well.

PR #1920, opened 2018-07-25
Fix failing tests

goetas:
fix error introduced with https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1912

Because of it `body_converter.validate` is always `NULL`


GuilhemN:
Thank you :)

PR #1923, opened 2018-07-26
The "Sensio\Bundle\FrameworkExtraBundle\Configuration\Route" annotati…

umpirsky:
…on is deprecated since version 5.2. Use "Symfony\Component\Routing\Annotation\Route" instead

BTW, why is [sensio/framework-extra-bundle](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/composer.json#L49) dev dependency?

### DIFF
--- a/Controller/Annotations/Route.php
+++ b/Controller/Annotations/Route.php
@@ -11,7 +11,7 @@
 
 namespace FOS\RestBundle\Controller\Annotations;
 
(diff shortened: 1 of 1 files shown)


GuilhemN:
Closing since there is already https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1822. FrameworkExtraBundle is a dev dependency because it is not strictly required.

PR #1926, opened 2018-07-28
Reset internal state of ViewHandler at the end of request

dnna:
ViewHandler keeps an internal state of the variables

- exclusionStrategyGroups
- exclusionStrategyVersion
- serializeNullStrategy

which causes inconsistencies when used with [PHP-PM](https://github.com/php-pm/php-pm). This PR fixes this by utilizing the "kernel.reset" tag introduced for this purpose in https://github.com/symfony/symfony/issues/23984 to reset the state of these variables after each request.

### DIFF
--- a/View/ViewHandler.php
+++ b/View/ViewHandler.php
@@ -502,4 +502,14 @@ class ViewHandler implements ConfigurableViewHandlerInterface
 
         return $form;
     }
+
+    /**
+     * Resets internal object state at the end of the request
+     */
+    public function reset()
+    {
+        $this->exclusionStrategyGroups = [];
+        $this->exclusionStrategyVersion = null;
+        $this->serializeNullStrategy = null;
+    }
 }
--- a/View/ViewHandler.php
+++ b/View/ViewHandler.php
@@ -117,6 +117,7 @@ class ViewHandler implements ConfigurableViewHandlerInterface
      * @param bool                  $serializeNull        Whether or not to serialize null view data
      * @param array                 $forceRedirects       If to force a redirect for the given key format, with value being the status code to use
      * @param string                $defaultEngine        default engine (twig, php ..)
+     * @param array                 $options              config options
      */
     public function __construct(
         UrlGeneratorInterface $urlGenerator,
@@ -129,7 +130,7 @@ class ViewHandler implements ConfigurableViewHandlerInterface
         $serializeNull = false,
         array $forceRedirects = null,
         $defaultEngine = 'twig',
-        array $options
+        array $options = []
     ) {
         $this->urlGenerator = $urlGenerator;
         $this->serializer = $serializer;
@@ -141,7 +142,11 @@ class ViewHandler implements ConfigurableViewHandlerInterface
         $this->serializeNull = $serializeNull;
         $this->forceRedirects = (array) $forceRedirects;
         $this->defaultEngine = $defaultEngine;
-        $this->options = $options;
+        $this->options = $options + [
+            'exclusionStrategyGroups' => [],
+            'exclusionStrategyVersion' => null,
+            'serializeNullStrategy' => null
+            ];
         $this->reset();
     }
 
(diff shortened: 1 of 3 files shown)


View/ViewHandler.php
```diff
@@ -502,4 +502,14 @@ private function getDataFromView(View $view)
...
     }
+
+    /**
+     * Resets internal object state at the end of the request
+     */
+    public function reset()
+    {
+        $this->exclusionStrategyGroups = [];
+        $this->exclusionStrategyVersion = null;
+        $this->serializeNullStrategy = null;
```
GuilhemN: I'm not sure resetting these fields is legit, they're defined in config so not per request, right ?

GuilhemN: Ok but then we should reset to config values, don't you think ?

dnna: Well, we reset them to the default values of ViewHandler, like they would be at the start of a new request. If there is config to set them to something else it will be called and explicitly set them, otherwise they should be the same as what the class sets as default.

dnna: To be more clear, the main difference between PPM and the normal PHP execution model is that objects are not destroyed at the end of the request. Otherwise the normal symfony request-response flow is executed (so if there is config to explicitly set values in the ViewHandler it will be called normally).

The problem only arises when the initial state of the object differs from what you would expect at the start of a normal request, as is the case here. So we just need to reset the class to what it would be at the start of a normal PHP request, before it even reaches Symfony.

GuilhemN: Thing is the initial state of the object differs from its state at the beginning of requests because of https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/DependencyInjection/FOSRestExtension.php#L400.

dnna: OK, `setSerializeNullStrategy` was a bad example and seems to be explicitly set in all cases :)

But take for example `setExclusionStrategyGroups` a few lines above. If `$config['serializer']['groups'] is set in one request, but in the next request its empty, then ViewHandler will carry over the exclusionStrategyGroups attribute from the last request that set it, instead of being empty as it should.

GuilhemN: Sure but I think we should save differently the config settings in order to use them when resetting

dnna: The thing is that technically the user can also call the `set*` methods in a controller, a listener, or other code regardless of the config. It could probably also work if `FOSRestExtension` explicitly calls all the `set*` methods of the `ViewHandler` regardless if they are empty (I think this is what you suggest?), but in my opinion it makes more sense to reset the object to its original state and call only the setters that are needed in `FOSRestExtension`.

GuilhemN: To be clearer, config options should be passed to the constructor and stored to be later restored. The constructor has already many arguments so we could group most of them under an `$options` argument but I wouldn't mind if you did just add the resetted config parameters as arguments in this PR.

dnna: @GuilhemN its done, I have added the config options as a constructor argument instead of method calls and reset resets them to the original values at construction.

View/ViewHandler.php
```diff
@@ -138,6 +141,8 @@ public function __construct(
         $this->serializeNull = $serializeNull;
         $this->forceRedirects = (array) $forceRedirects;
         $this->defaultEngine = $defaultEngine;
+        $this->options = $options;
```
GuilhemN: `$this->options = $options + ['exclusionStrategyGroups' => [], 'exclusionStrategyVersion' => null, 'serializeNullStrategy' => null];` to remove isset checks in `reset()` and ensure fields are reset to their default value if none was explicitely provided :)

View/ViewHandler.php
```diff
@@ -126,7 +128,8 @@ public function __construct(
         $emptyContentCode = Response::HTTP_NO_CONTENT,
         $serializeNull = false,
         array $forceRedirects = null,
-        $defaultEngine = 'twig'
+        $defaultEngine = 'twig',
+        array $options
```
GuilhemN: Should be optional: `$options = []`

dnna:
Ping @GuilhemN :) Are there any other issues with this one?

GuilhemN:
That's all good for me :+1:
Thanks!

PR #1929, opened 2018-08-08
Remove `fos_rest.decoder.xml` when the symfony serializer is absent

### LINKED ISSUE #1928: XmlDecoder service is registered even when Symfony seriliazer is not required
Hi,

The `symfony/serializer` is only in the suggested package, but the `\FOS\RestBundle\Decoder\XmlDecoder` service is always registered. 

So when XML is sent to an api without the symfony serializer, the exception `Class 'Symfony\Component\Serializer\Encoder\XmlEncoder' not found` is thrown. 

Maybe the XmlDecoder should be registered only when the service is available, or it should be removed in a compiler pass?
I'm willing to do a PR if required :)

magnetik:
To fix #1928 

Not sure if it's the cleanest way to go but it works.


DependencyInjection/Compiler/SerializerConfigurationPass.php
```diff
@@ -26,6 +27,10 @@
 {
     public function process(ContainerBuilder $container)
     {
+        if (!class_exists(XmlEncoder::class)) {
+            $container->removeDefinition('fos_rest.decoder.xml');
```
Tobion: This does not solve the problem you described. The service name is used in https://github.com/FriendsOfSymfony/FOSRestBundle/blob/5e38509c14798579ba5655846cc9c4876f0b40e7/DependencyInjection/Configuration.php#L287 and the service is tried to be fetched from the container

So insead the `Class 'Symfony\Component\Serializer\Encoder\XmlEncoder' not found`, this causes a service does not exist error. So the better solution would be to remove the decoder registration in the config I linked above when the serializer is not installed.

magnetik: Humm right, nice catch. 

https://github.com/FriendsOfSymfony/FOSRestBundle/blob/5e38509c14798579ba5655846cc9c4876f0b40e7/DependencyInjection/FOSRestExtension.php#L164

So checking if the "id" is a service here would work and help preventing bad configuration, right?

magnetik: @Tobion  ?

Tobion: Service existence checks should not be done in the extension as it then depends on the order the extensions are registered. As I said, you could check if the class exists and adjust the config default accordingly.

DependencyInjection/Configuration.php
```diff
@@ -269,6 +269,10 @@ private function addViewSection(ArrayNodeDefinition $rootNode)
 
     private function addBodyListenerSection(ArrayNodeDefinition $rootNode)
     {
+        $decodersDefaultValue = ['json' => 'fos_rest.decoder.json'];
+        if (class_exists('FOS\RestBundle\Decoder\XmlDecoder')) {
```
xabbuh: please add a `use` statement for this class and make use of the `class` constant here

DependencyInjection/Configuration.php
```diff
@@ -269,6 +270,10 @@ private function addViewSection(ArrayNodeDefinition $rootNode)
 
     private function addBodyListenerSection(ArrayNodeDefinition $rootNode)
     {
+        $decodersDefaultValue = ['json' => 'fos_rest.decoder.json'];
+        if (class_exists(XmlDecoder::class)) {
```
Tobion: What's the point of this? This class always exists. You need to check against `Symfony\Component\Serializer\Encoder\XmlEncoder` ...

magnetik: wow I'm really sorry about that. I should focus more when doing a pull request.

DependencyInjection/Configuration.php
```diff
@@ -11,6 +11,7 @@
 
 namespace FOS\RestBundle\DependencyInjection;
 
+use Symfony\Component\Serializer\Encoder\XmlEncoder;
```
Tobion: please sort them alphabetically like it was before

PR #1934, opened 2018-08-21
fix route generator not ignoring body param converters

### LINKED ISSUE #1198: Make a better marriage between body and param converter

tonivdv:
### This is an update on PR #1771 with the missing test case. All credits should go to @Parent5446 

When fetching a list of function parameters from a controller action
for use as URL parameters, check for the @ParamConverter
annotation. If a parameter has the annotation with the
"fos_rest.request_body" converter, then ignore it and do not include
it in the list of parameters (thus removing it from the URL).

This allows automatic route generation and ParamConverter to co-exist,
as otherwise a parameter that is expected to be derived from the body
would also be included as a URL parameter, requiring the developer to
manually specify the route.

fixes #1198

### DIFF
--- a/Routing/Loader/Reader/RestActionReader.php
+++ b/Routing/Loader/Reader/RestActionReader.php
@@ -477,7 +477,7 @@ class RestActionReader
 
         // check if a parameter is coming from the request body
         $ignoreParameters = [];
-        if (class_exists('Sensio\\Bundle\\FrameworkExtraBundle\\Configuration\\ParamConverter')) {
+        if (class_exists(\Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter::class)) {
             $ignoreParameters = array_map(function ($annotation) {
                 return
                     $annotation instanceof \Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter &&
--- a/Routing/Loader/Reader/RestActionReader.php
+++ b/Routing/Loader/Reader/RestActionReader.php
@@ -240,7 +240,7 @@ class RestActionReader
      * Reads action route.
      *
      * @param RestRouteCollection $collection
-     * @param ReflectionMethod   $method
+     * @param ReflectionMethod    $method
      * @param string[]            $resource
      *
      * @throws \InvalidArgumentException
@@ -431,7 +431,7 @@ class RestActionReader
      * Returns HTTP method and resources list from method signature.
      *
      * @param ReflectionMethod $method
-     * @param string[]          $resource
+     * @param string[]         $resource
      *
      * @return bool|array
      */
@@ -565,9 +565,9 @@ class RestActionReader
     /**
      * Generates URL parts for route from resources list.
      *
-     * @param string[]               $resources
+     * @param string[]              $resources
      * @param ReflectionParameter[] $arguments
-     * @param string                 $httpMethod
+     * @param string                $httpMethod
      *
      * @return array
      */
@@ -609,8 +609,8 @@ class RestActionReader
     /**
      * Returns custom HTTP method for provided list of resources, arguments, method.
      *
-     * @param string                 $httpMethod current HTTP method
-     * @param string[]               $resources  resources list
+     * @param string                $httpMethod current HTTP method
+     * @param string[]              $resources  resources list
      * @param ReflectionParameter[] $arguments  list of method arguments
      *
      * @return string
@@ -671,7 +671,7 @@ class RestActionReader
      * Reads method annotations.
      *
      * @param ReflectionMethod $reflectionMethod
-     * @param string            $annotationName
+     * @param string           $annotationName
      *
      * @return RouteAnnotation|null
      */
@@ -688,7 +688,7 @@ class RestActionReader
      * Reads method annotations.
      *
      * @param ReflectionMethod $reflectionMethod
-     * @param string            $annotationName
+     * @param string           $annotationName
      *
      * @return RouteAnnotation[]
      */
--- a/Routing/Loader/Reader/RestActionReader.php
+++ b/Routing/Loader/Reader/RestActionReader.php
@@ -475,6 +475,17 @@ class RestActionReader
         // ignore all query params
         $params = $this->paramReader->getParamsFromMethod($method);
 
+        // check if a parameter is coming from the request body
+        $ignoreParameters = [];
+        if (class_exists('Sensio\\Bundle\\FrameworkExtraBundle\\Configuration\\ParamConverter')) {
+            $ignoreParameters = array_map(function ($annotation) {
+                return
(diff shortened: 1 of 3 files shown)


tonivdv:
I was not able to run the php-cs fixer. Somehow it fails on my machine :/ And there is no doc on how it should be executed. It would be cool to have a docker config for that to avoid this kind of situations ;)

tonivdv:
ping @GuilhemN

tonivdv:
Fixed the CS issue with the provided diff by styleci 👍

Routing/Loader/Reader/RestActionReader.php
```diff
@@ -475,6 +475,17 @@ private function getMethodArguments(\ReflectionMethod $method)
         // ignore all query params
         $params = $this->paramReader->getParamsFromMethod($method);
 
+        // check if a parameter is coming from the request body
+        $ignoreParameters = [];
+        if (class_exists('Sensio\\Bundle\\FrameworkExtraBundle\\Configuration\\ParamConverter')) {
```
xabbuh: let's use the `class` constant here

Tests/Routing/Loader/RestRouteLoaderTest.php
```diff
@@ -370,6 +370,19 @@ public function testRequestTypeHintsIgnoredCorrectly()
...
+     * @see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1198
+     */
+    public function testParamConverterIsIgnoredInRouteGenerationCorrectly()
+    {
+        $collection = $this->loadFromControllerFixture('ParamConverterController');
+
+        $this->assertNotNull($collection->get('post_something'), 'route for "post_something" does not exist');
+        $this->assertEquals('/somethings.{_format}', $collection->get('post_something')->getPath());
+
+        $this->assertTrue(true);
```
xabbuh: this should be removed

Tests/Routing/Loader/RestRouteLoaderTest.php
```diff
@@ -370,6 +370,19 @@ public function testRequestTypeHintsIgnoredCorrectly()
...
 
+    /**
+     * @see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1198
+     */
+    public function testParamConverterIsIgnoredInRouteGenerationCorrectly()
+    {
+        $collection = $this->loadFromControllerFixture('ParamConverterController');
+
+        $this->assertNotNull($collection->get('post_something'), 'route for "post_something" does not exist');
+        $this->assertEquals('/somethings.{_format}', $collection->get('post_something')->getPath());
```
xabbuh: should be `assertSame()`

Tests/Routing/Loader/RestRouteLoaderTest.php
tonivdv: oops

Tests/Routing/Loader/RestRouteLoaderTest.php
tonivdv: Ok. Just copied from previous tests.

Routing/Loader/Reader/RestActionReader.php
```diff
@@ -475,6 +475,17 @@ private function getMethodArguments(\ReflectionMethod $method)
         // ignore all query params
         $params = $this->paramReader->getParamsFromMethod($method);
 
+        // check if a parameter is coming from the request body
+        $ignoreParameters = [];
+        if (class_exists(\Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter::class)) {
```
xabbuh: but please also add use case for the classes :)

Routing/Loader/Reader/RestActionReader.php
```diff
@@ -500,6 +511,10 @@ private function getMethodArguments(\ReflectionMethod $method)
                 }
             }
 
+            if (in_array($argument->getName(), $ignoreParameters)) {
```
xabbuh: we should also pass `true` as the third argument

Routing/Loader/Reader/RestActionReader.php
tonivdv: Not sure I understand what you mean, can you elaborate.

xabbuh: Oh, I meant `use` statements. 🙈

tonivdv: Alright, I prefer that too, but since a lot in the code is with FQN I thought it was some standard or something. Will do that simplification.

tonivdv: I changed all FQN in that method so it's cleaned up at the same time.

Routing/Loader/Reader/RestActionReader.php
```diff
@@ -466,21 +472,32 @@ private function getHttpMethodAndResourcesFromMethod(\ReflectionMethod $method,
...
      * Returns readable arguments from method.
      *
-     * @param \ReflectionMethod $method
+     * @param ReflectionMethod $method
      *
-     * @return \ReflectionParameter[]
+     * @return ReflectionParameter[]
      */
-    private function getMethodArguments(\ReflectionMethod $method)
+    private function getMethodArguments(ReflectionMethod $method)
```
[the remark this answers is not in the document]
tonivdv: Ah lol. Alright no problem.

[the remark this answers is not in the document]
tonivdv: done

GuilhemN:
Thank you @tonivdv for taking care of this!

GuilhemN:
We usually release when there are a significant number of bug fixes/features awaiting (see https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1870). I guess we can release 2.4 though, important fixes are awaiting right now. Is it ok for you @xabbuh?

xabbuh:
👍 let's go for it

GuilhemN:
Done, see https://github.com/FriendsOfSymfony/FOSRestBundle/releases/tag/2.4.0

PR #1945, opened 2018-10-02
Fix twig unnecessary required

pedropapa:


Tobion:
Related to #2002 as we need to find a way to do all that without the templating component part

Tobion:
Closing in favor of #2012. Thanks for the PR anyway!

Issue #1961, opened 2018-11-26
Undefined class constant DIRECTION_SERIALIZATION


pdias:
The change in ExceptionHandler from JMS\Serializer\GraphNavigator to JMS\Serializer\GraphNavigatorInterface will originate this error:

> Fatal error: Uncaught Symfony\Component\Debug\Exception\FatalThrowableError
>   : Undefined class constant 'DIRECTION_SERIALIZATION' in D:\wamp\www\api\ven
>   dor\friendsofsymfony\rest-bundle\Serializer\Normalizer\ExceptionHandler.php
>   :29

Thanks,
Paulo

xabbuh:
fixed in #1962

PR #1899, opened 2018-06-14
jms/serializer 2.0 compatibility

goetas:
This introduces the changes for supporting the upcoming jms/serializer 2.0 release

Some changes in composer.json and travisci are will not be necessary when serializer v1.13 will be released

PR #1962, opened 2018-11-29
Drop jms/serializer <1.13

IonBazan:
~~`GraphNavigatorInterface` has been introduced in JMS Serializer v1.12.0. We should use `GraphNavigator` instead to maintain backward compatibility with v1.11.0 and lower.~~

#1899 breaks compatibility with `jms/serializer <1.13.0` by using:

- `JMS\Serializer\GraphNavigatorInterface` (>=1.12)
- `JMS\Serializer\Context::hasAttribute()` (>=1.13)
- `JMS\Serializer\Context::getAttribute()` (>=1.13)

However, it is still possible to install older version of `jms/serializer` together with FOSRestBundle. This PR fixes it by updating a `conflict` constraint.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -75,7 +75,7 @@
     "conflict": {
         "sensio/framework-extra-bundle": "<3.0.13",
         "jms/serializer-bundle": "<2.0.0",
-        "jms/serializer": "1.3.0"
+        "jms/serializer": "<1.13.0"
     },
     "extra": {
         "branch-alias": {


composer.json
```diff
@@ -62,7 +62,7 @@
         "symfony/css-selector": "^2.7|^3.0|^4.0",
         "phpoption/phpoption": "^1.1",
         "jms/serializer-bundle": "^2.0|^3.0",
-        "jms/serializer": "^1.13|^2.0",
+        "jms/serializer": "^1.5|^2.0",
```
xabbuh: this change needs to be reverted (see the failing test)

IonBazan: I was trying to fix #1961 because FOSRestBundle allows installing `jms/serializer <1.12` and triggers an error. Should we drop support for `<1.12` by adding it to `conflicts` section?

xabbuh: Yes, I think that would be better.

xabbuh:
Thank you @IonBazan.

PR #1964, opened 2018-12-07
Use AbstractController to fix deprecated since Symfony 4.2

phamuyentri:


IonBazan:
@phamuyentri `Symfony\Bundle\FrameworkBundle\Controller\AbstractController` has been introduced in Symfony 3.3. I believe FOSRestBundle is trying to keep compatibility with Symfony 2.7 so I would leave that deprecation instead of breaking BC.

xabbuh:
I am not so much concerned about dropping support for Symfony 2 especially given the fact that even Symfony 2.8 doesn't receive bugfixes anymore. But, what's more important to consider is that making this change will probably break a lot of applications out there as the `AbstractController` does not allow to access all public services if you do not explicitly subscribe to each and every service you need.

I think we should do the following instead: Deprecate the existing controller class, introduce a new one extending `AbstractController` that offers the same functionality (we can share common code between both controller classes with a trait).

Oviglo:
+1

phamuyentri:
@xabbuh I understand your concern, but I have two reasons to fix like that:

1. As I checked, the different between AbstractController  and Controller is Abstract has function `getSubscribedServices`. Abstract has everything in Controller.
2. Symfony team has reason to mark it deprecated. If you use AbstractController instead of Controller and causes a bug; Symfony team should know (and fix) it.

xabbuh:
`getSubscribedServices()` is the important differences here because `AbstractController` does not have access to the full DI container, but only gets injected a subset of the full container giving you only access to the services specified in `getSubscribedServices()`.

phamuyentri:
Got it. Thank you, @xabbuh.
I pushed another fix which introduces new Abstract Controller [here](https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1966), please have a look.
Again, many thanks for your explanation.

PR #1967, opened 2018-12-07
Mark deprecated since Symfony 4.2 for Controller

phamuyentri:
Create new `AbstractFOSRestController` and mark `FOSRestController` as deprecated because `Symfony\Bundle\FrameworkBundle\Controller\Controller` is deprecated in Symfony 4.2

### DIFF
--- a/Controller/AbstractFOSRestController.php
+++ b/Controller/AbstractFOSRestController.php
@@ -36,7 +36,7 @@ abstract class AbstractFOSRestController extends AbstractController
     }
 
     /**
-     *
+     * @inheritdoc
      *
      * @return array
      */
--- a/Controller/AbstractFOSRestController.php
+++ b/Controller/AbstractFOSRestController.php
@@ -12,16 +12,12 @@
 namespace FOS\RestBundle\Controller;
 
 use FOS\RestBundle\View\ViewHandlerInterface;
-use Symfony\Bundle\FrameworkBundle\Controller\Controller;
+use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
 
 /**
  * Controllers using the View functionality of FOSRestBundle.
- *
- * @deprecated when used with Symfony 4.2, use {@see AbstractFOSRestController} instead.
- *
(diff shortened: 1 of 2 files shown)


Controller/AbstractFOSRestController.php
```diff
@@ -0,0 +1,37 @@
...
+
+    /**
+     * Get the ViewHandler.
+     *
+     * @return ViewHandlerInterface
+     */
+    protected function getViewHandler()
+    {
+        if (!$this->viewhandler instanceof ViewHandlerInterface) {
+            $this->viewhandler = $this->container->get('fos_rest.view_handler');
```
xabbuh: We will also have to override the `getSubscribedServices()` method so that we can actually access the `fos_rest.view_handler` service here:

```php
public static function getSubscribedServices()
{
    return array(
            'fos_rest.view_handler' => ViewHandlerInterface::class,
    );
}
```

Controller/FOSRestController.php
```diff
@@ -17,6 +17,8 @@
 /**
  * Controllers using the View functionality of FOSRestBundle.
  *
+ * @deprecated when used with Symfony 4.2, use {@see AbstractFOSRestController} instead.
```
xabbuh: I think we should just deprecate the class:

```php
/**
 * @deprecated since FOSRestBundle 2.5, use {@see AbstractFOSRestController} instead
 */
```

Controller/AbstractFOSRestController.php
phamuyentri: Should we add `'fos_rest.view_handler' into parent list, like this?
```
public static function getSubscribedServices()
 {
     $subscribedServices = parent::getSubscribedServices();
     $subscribedServices['fos_rest.view_handler'] = ViewHandlerInterface::class;

     return $subscribedServices;
 }
```

xabbuh: yes 👍

xabbuh:
Thank you @phamuyentri.

PR #1969, opened 2018-12-08
Remove obsolete doc

phamuyentri:


xabbuh:
Good catch, thanks @phamuyentri.

PR #1970, opened 2018-12-09
Apply NotBlank only to the whole array

GuilhemN:
Fixes https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1769/files/da8d1ff7d91e5e178022b82a206eacf9b604f3b1#r231027007

### DIFF
--- a/Controller/Annotations/AbstractScalarParam.php
+++ b/Controller/Annotations/AbstractScalarParam.php
@@ -56,10 +56,6 @@ abstract class AbstractScalarParam extends AbstractParam
             ));
         }
 
-        if (false === $this->allowBlank) {
-            $constraints[] = new NotBlank();
-        }
-
         // If the user wants to map the value, apply all constraints to every
         // value of the map
         if ($this->map) {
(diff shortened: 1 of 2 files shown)


Controller/Annotations/AbstractScalarParam.php
```diff
@@ -69,9 +65,10 @@ public function getConstraints()
             if (false === $this->nullable) {
                 $constraints[] = new NotNull();
             }
-            if (false === $this->allowBlank) {
-                $constraints[] = new NotBlank();
-            }
+        }
+
+        if (false === $this->allowBlank) {
```
xabbuh: I think my suggestion in #1769 was not good either. Keeping the constraint on line 59 is probably necessary so that it will be applied to the nested data. Though applying it to the collection itself does not make much sense IMO (and we also have the `NotNull` constraint).

GuilhemN: So should we just keep the `NotNull` additionnal check? I think it would make sense, I'm updating it pending your answer :)

xabbuh: 👍

Issue #1843, opened 2018-01-11
[SF3.4] Attempted to load class "NotNull" from namespace "Symfony\Component\Validator\Constraints"


angrygreenfrogs:

angrygreenfrogs:
Controller code:
  
  ```
/**
     * Login to the system via email/password and receive a security token.
     * @Rest\View()
     * @Route("/user/login.{_format}",defaults={"_format": "json"})
     * @Method("POST")
     *
     * @RequestParam(name="apikey", requirements="[a-zA-Z0-9]{32}", allowBlank=false, strict=true, description="Application API Key")
     * @RequestParam(name="email", requirements=".+@.+\..+", allowBlank=false, strict=true, description="User's e-mail address")
     * @RequestParam(name="password", requirements=".{4,30}", allowBlank=false, strict=true, description="User's password")
     * @param ParamFetcher $paramFetcher
     * @param Request $request
     * @param LoggerInterface $logger
     * @return Response
     * @throws HttpException on application failure
     */
    public function postLoginAction(ParamFetcher $paramFetcher, Request $request, LoggerInterface $logger)
    {
```

Upon sending an API request, this results in an error like:

```
[2018-01-11 16:13:57] php.CRITICAL: Class 'Symfony\Component\Validator\Constraints\NotNull' not found {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0): Class 'Symfony\\Component\\Validator\\Constraints\\NotNull' not found at /home/eross/myp/vendor/friendsofsymfony/rest-bundle/Controller/Annotations/AbstractParam.php:75)"} []
[...]

PR #1974, opened 2018-12-26
Throw meaningful exception when using ParamFetcher without symfony/validator installed

l-vo:
Related to #1843 

It seems symfony/validator (currently in composer dev requirements) should be in requirements instead, since at least the `NotNull` constraint is used here: https://github.com/FriendsOfSymfony/FOSRestBundle/blob/ee844447be0fe6fbe99ed678f70db342a60a4ccc/Controller/Annotations/AbstractParam.php#L75

### DIFF
--- a/DependencyInjection/FOSRestExtension.php
+++ b/DependencyInjection/FOSRestExtension.php
@@ -24,6 +24,7 @@ use Symfony\Component\Form\Extension\Core\Type\FormType;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\HttpKernel\DependencyInjection\Extension;
 use Symfony\Component\HttpKernel\Kernel;
+use Symfony\Component\Validator\Constraint;
 
 class FOSRestExtension extends Extension
 {
@@ -230,6 +231,10 @@ class FOSRestExtension extends Extension
     private function loadParamFetcherListener(array $config, XmlFileLoader $loader, ContainerBuilder $container)
     {
         if ($config['param_fetcher_listener']['enabled']) {
+            if (!class_exists(Constraint::class)) {
+                @trigger_error('Enabling the fos_rest.param_fetcher_listener option when the Symfony Validator component is not installed is deprecated since FOSRestBundle 2.6 and will throw an exception in 3.0. Disable the feature or install the symfony/validator package.', E_USER_DEPRECATED);
+            }
+
             $loader->load('param_fetcher_listener.xml');
 
             if (!empty($config['param_fetcher_listener']['service'])) {
--- a/Request/ParamFetcher.php
+++ b/Request/ParamFetcher.php
@@ -133,15 +133,16 @@ class ParamFetcher implements ParamFetcherInterface
             return $paramValue;
         }
 
-        $constraints = $param->getConstraints();
-        $this->resolveConstraints($constraints);
-        if (empty($constraints)) {
-            return $paramValue;
-        }
         if (null === $this->validator) {
             throw new \RuntimeException(
                 'The ParamFetcher requirements feature requires the symfony/validator component.'
             );
+        }
+
+        $constraints = $param->getConstraints();
+        $this->resolveConstraints($constraints);
+        if (empty($constraints)) {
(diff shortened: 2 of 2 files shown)


xabbuh:
The Validator component is only needed when making use of the param fetcher feature, isn't it? If so, I think a better fix would be to throw a meaningful exception when the feature is enabled, but the component is missing.

l-vo:
@xabbuh Indeed, this kind of check was already done but was executed too late in the code. I moved it before constraints retrieval.

xabbuh:
I still see one issue with this solution: This exception will only be thrown by runtime and thus could be missed by the developer. Wouldn't it make more sense to error at compile time in `FOSRestExtension:: loadParamFetcherListener()` if the `fos_rest. param_fetcher_listener ` option is enabled?

l-vo:
@xabbuh updated, thanks :)

Request/ParamFetcher.php
```diff
@@ -136,7 +136,7 @@ protected function cleanParamWithRequirements(ParamInterface $param, $paramValue
         }
         if (null === $this->validator) {
             throw new \RuntimeException(
-                'The ParamFetcher requirements feature requires the symfony/validator component.'
+                'The ParamFetcher requirements feature requires the validator service.'
```
xabbuh: I think we can improve this a bit by triggering a deprecation in the constructor when no validator is passed. That would allow us to get rid of this runtime check in the next major version.

xabbuh: But if you prefer to, we can do that in a follow-up PR.

l-vo: Done in #1980 but you've already seen it :)

DependencyInjection/FOSRestExtension.php
```diff
@@ -230,6 +231,12 @@ private function loadVersioning(array $config, XmlFileLoader $loader, ContainerB
     private function loadParamFetcherListener(array $config, XmlFileLoader $loader, ContainerBuilder $container)
     {
         if ($config['param_fetcher_listener']['enabled']) {
+            if (!class_exists(Constraint::class)) {
+                throw new \RuntimeException(
```
xabbuh: I am afraid I was too eager by suggesting to throw an exception here. There may be applications where someone enabled the option without using the feature at all. Having these applications break on an update is not the best experience. I think we should instead just trigger a deprecation here.

l-vo: Indeed it might be more prudent. But the developper might miss it too because it's only a deprecation. The initial goal of the PR is too avoid a "class not found" message. What do you think about doing both things:
- Add a deprecation at compile time
- Fix the validator check position like I suggest it earlier ([69e1d7](https://github.com/FriendsOfSymfony/FOSRestBundle/commit/69e1d7f94e1ff30a7dcd1b680ea3cdbecd5745a3)) to have a more mainingfull message than "class not found" ?

l-vo: I pushed the suggested changes

DependencyInjection/FOSRestExtension.php
```diff
@@ -230,6 +231,10 @@ private function loadVersioning(array $config, XmlFileLoader $loader, ContainerB
     private function loadParamFetcherListener(array $config, XmlFileLoader $loader, ContainerBuilder $container)
     {
         if ($config['param_fetcher_listener']['enabled']) {
+            if (!class_exists(Constraint::class)) {
+                @trigger_error('The ParamFetcher requirements feature requires the symfony/validator component.', E_USER_DEPRECATED);
```
xabbuh: I think we should reword this a bit now:

> Enabling the `fos_rest.param_fetcher_listener` option when the Symfony Validator component is not installed is deprecated since FOSRestBundle 2.6 and will throw an exception in 3.0. Disable the feature or install the `symfony/validator` package.

l-vo: @xabbuh Reworded, thank you :)

GuilhemN:
Thank you @l-vo!

PR #1980, opened 2019-01-17
Depreciate null validator on ParamFetcher constructor

l-vo:
Depreciate null validator on ParamFetcher constructor to remove runtime check (see below) at the next major version.

https://github.com/FriendsOfSymfony/FOSRestBundle/blob/a5fc73b84bdb2f0fdf58a717b322ceb6997f7bf3/Request/ParamFetcher.php#L137

### DIFF
--- a/Request/ParamFetcher.php
+++ b/Request/ParamFetcher.php
@@ -52,6 +52,10 @@ class ParamFetcher implements ParamFetcherInterface
      */
     public function __construct(ContainerInterface $container, ParamReaderInterface $paramReader, RequestStack $requestStack, ValidatorInterface $validator = null)
     {
+        if (null === $validator) {
+            @trigger_error(sprintf('Using no validator is deprecated since FOSRestBundle 2.6. The `$validator` constructor argument of the `%s` will become mandatory in 3.0.', __CLASS__), E_USER_DEPRECATED);
+        }
+
         $this->container = $container;
         $this->requestStack = $requestStack;
         $this->validator = $validator;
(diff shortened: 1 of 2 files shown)


Request/ParamFetcher.php
```diff
@@ -52,6 +52,10 @@ class ParamFetcher implements ParamFetcherInterface
      */
     public function __construct(ContainerInterface $container, ParamReaderInterface $paramReader, RequestStack $requestStack, ValidatorInterface $validator = null)
     {
+        if (null === $validator) {
+            @trigger_error("Using no validator is deprecated since version 2.6. It won't be possible anymore in 3.0.");
```
xabbuh: We should reword this a bit like this:

```php
@trigger_error(sprintf('Using no validator is deprecated since FOSRestBundle 2.6. The `$validator` constructor argument of the `%s` will become mandatory in 3.0.', __CLASS__), E_USER_DEPRECATED);
```

Tests/Request/ParamFetcherTest.php
```diff
@@ -83,6 +83,21 @@ public function setup()
         $this->container = $this->getMockBuilder('Symfony\Component\DependencyInjection\ContainerInterface')->getMock();
     }
 
+    public function testConstructorCallWithNullValidatorShouldTriggerDeprecation()
```
xabbuh: We can make use of the `@expectedDeprecation` feature of the PHPUnit bridge:

```php
/**
 * @expectedDeprecation put the deprecation message here
 * @group legacy
 */
public function testConstructorCallWithNullValidatorShouldTriggerDeprecation()
{
    new ParamFetcher($this->container, $this->paramReader, $this->requestStack, null);
}
```

Request/ParamFetcher.php
l-vo: My bad, I forgot E_USER_DEPRECATED type, sorry for that :/

Tests/Request/ParamFetcherTest.php
l-vo: Great ! I didn't know `expectedDeprecation` annotation, thank you :))

GuilhemN:
Thank you @l-vo :)

Issue #1996, opened 2019-05-16
serialize_null configuration not taken into account


solcik:
`serialize_null` configuration not taken into account.

```json
"name": "friendsofsymfony/rest-bundle",
"version": "2.5.0",

"name": "symfony/serializer",
"version": "v4.2.8",
```

```yml
fos_rest:
  serializer:
    serialize_null: false
```

The problem is with converting context in this file.

https://github.com/FriendsOfSymfony/FOSRestBundle/blob/a5fc73b84bdb2f0fdf58a717b322ceb6997f7bf3/Serializer/SymfonySerializerAdapter.php#L37

The correct attribute is `skip_null_values` instead of `serializeNull` and its value has to be negated.

```php
 $newContext['skip_null_values'] = !$context->getSerializeNull(); 
```

---

A temporary way to go:

I have to create Context and add a correct attribute to get it to work.

```php
return View::create($response, Response::HTTP_OK)->setContext(
  (new FOS\RestBundle\Context\Context())->setAttribute(Symfony\Component\Serializer\Normalizer\AbstractObjectNormalizer::SKIP_NULL_VALUES, true)
)
```

PR #1997, opened 2019-05-16
fix: symfony serializer skip_null_values configuration

solcik:
Resolves issue https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1996


Tobion:
Good catch. Thanks

javijavo:
Hi,
Can this fix be included in the 2.5 stable release?
Is a quite basic fix and it would help a lot given that functionality does not work at all.

Thanks!

Tobion:
Hi, the fix will be part of the next 2.6 release. Until then you can also use dev-master

PR #1999, opened 2019-05-30
Added path version resolver, to resolve versions passed via URI

cyberemissary:
### What is the problem
When controller has `@Version` annotation, in some cases the routes for that controller are generated using *requirements* and in others using *conditions* (not sure if this is intended, or can be normalized in a better way, as I'm not expert in routing). 

I have setup the situation in this MR when routing condition DOES get generated in which case the request fails with `404 Not Found` when version is passed via URI path. This is because none of the existing resolvers resolve the path version, so the `Request->attributes['version']` is set to default. Then, during routing, the *condition* evaluation tries uses the `Request->attributes['version']` for matching, which fails because the default version does not match the values from `@Version` annotation. 

> \FOS\RestBundle\Routing\Loader\Reader\RestActionReader::getVersionCondition() sets up the conditions

You can reproduce this by executing `\FOS\RestBundle\Tests\Functional\RoutingConditionsTest::testVersionInPathWithPrefix` and disabling the `path` resolver in `Tests/Functional/app/RoutingConditions/config.yml`.

### What This PR Does
1. Adds `PathVersionResolver`, which is a regex based resolver intended to resolve version passed via URI path prior to `RouterListener` execution.
2. Adds configuration options to customize guessing order and the regex for this resolver
3. Limit *symfony/framework-bundle* in **require-dev** to `<4.3`
    * in 4.3 they introduced static `$client` property in WebTestCase which breaks all tests
4. Fixed Tests/Routing/Loader/RestXmlCollectionLoaderTest.php asserting `'true' === true`
5. Add `doctrine/common` dependency. It was probably sub-dependency of another package previously, but now seems to be missing so tests are failing.
### Other Thoughts
* If you move the **prefix** option from `Tests/Functional/app/RoutingConditions/routing.yml` to `Tests/Functional/Bundle/TestBundle/Resources/config/rest_routing.yml` instead of *condition*, the routes are generated with *requirement* and everything works as expected, because *requirement* does not rely on *version* request attribute (which in this case is still be set to the default value during routing).
* CI failing due to deprecation in recently updated bundles. Sorry, don't have time to fix them here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ CHANGELOG
 
 * deprecated using the `ParamFetcher` class without passing a validator as the third argument, this
   argument will become mandatory in 3.0
+* added `PathVersionResolver` for better version resolution when version is passed via placeholder in URI
 
 2.5.0
 -----
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -397,19 +397,28 @@ final class Configuration implements ConfigurationInterface
                                     ->scalarNode('regex')->defaultValue('/(v|version)=(?P<version>[0-9\.]+)/')->end()
                                 ->end()
                             ->end()
+                            ->arrayNode('path')
+                                ->canBeDisabled()
+                                ->children()
+                                    ->scalarNode('regex')
+                                        ->defaultValue('/\\/(?P<version>v?[0-9\.]+)\\//')
+                                        ->info('If your version is in path like so /api/{version}/action the following regex could be used: /^\\/api\/(?P<version>v?[0-9\\.]+)\\//')
+                                    ->end()
+                                ->end()
+                            ->end()
                         ->end()
                     ->end()
                     ->arrayNode('guessing_order')
-                        ->defaultValue(['query', 'custom_header', 'media_type'])
+                        ->defaultValue(['query', 'custom_header', 'media_type', 'path'])
                         ->validate()
                             ->ifTrue(function ($v) {
                                 foreach ($v as $resolver) {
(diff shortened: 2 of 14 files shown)


zajca:
Great feature, I was thinking about allow custom VersionResolver (now it's only option to create own VersionListener) but I was missing only this resolver.

GuilhemN:
If I'm correct, adding a `version` attribute to one of your routes path would take precedence over the version listener and do what you'd expect, isn't that enough ?

GuilhemN:
If I'm correct, adding a `version` attribute to one of your routes path would take precedence over the version listener and do what you'd expect, isn't that enough ?

GuilhemN:
If I'm correct, adding a `version` attribute to one of your routes path would take precedence over the version listener and do what you'd expect, isn't that enough ?

zajca:
> If I'm correct, adding a `version` attribute to one of your routes path would take precedence over the version listener and do what you'd expect, isn't that enough ?

You have to add version attribute to every route, even when version is already in path and can be resolved automatically.

cyberemissary:
> If I'm correct, adding a `version` attribute to one of your routes path would take precedence over the version listener and do what you'd expect, isn't that enough ?

I'm not sure if I understand this correctly, but in the test case **version** attribute is part of the path via `prefix` key in config in `Tests/Functional/app/RoutingConditions/routing.yml`

xabbuh:
Is this still relevant given that the automatic route generation feature will be deprecated in 2.8 and will be removed in 3.0 (and so will the `@Version` annotation too)?

cyberemissary:
If that's the case, i guess there is no point in merging this.

xabbuh:
Thank you for the confirmation.

Issue #1402, opened 2016-03-22
ExceptionListener ignores zone config


jovobe:
If the ExceptionListener is enabled (fos_rest.exception.enabled = true) it triggers on every route. Is this the expected behavior or did just someone forget to add something like:
![rest-bundle exceptionlistener php - core - gitrepos afw core 2016-03-22 13-35-21](https://cloud.githubusercontent.com/assets/1265894/13951687/fc776b02-f032-11e5-9ef2-ab3cec9fbf34.png)

Ener-Getick:
I guess this was forgotten, WDYT @lsmith77 ?

xabbuh:
Yeah, looks like it was forgotten.

@jovobe Would you like to open a pull request?

Ener-Getick:
Fixed in #1406

spolischook:
Why this issue is closed?
The #1406 PR add the same code as in issue and as a result Listener triggered on each exception.
Created PR https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2004
Don't know is ZONE_ATTRIBUTE will be setted for all rest queries, but hopefully tests will catch this case

spolischook:
I broke some tests but my hand tests on work application give me a desired result - json exception on rest actions and html standard symfony exception on other routes.

spolischook:
Found the doc https://symfony.com/doc/master/bundles/FOSRestBundle/3-listener-support.html#zone-listener

PR #2004, opened 2019-06-21
Remove default value for ZONE_ATTRIBUTE in Request

spolischook:


xabbuh:
Can you explain which issue this is going to solve and please also add a test case for it?

spolischook:
@xabbuh this one https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1402 . Is it make sense? or I made something wrong?
I have non rest routes where I throw exception and Fos Rest Exception listener run for all resources, so I have issue with template.

Should I set ZONE_ATTRIBUTE parameter to false in non rest routes? and how?

spolischook:
Should I configure ZoneMatcherListener somehow?

xabbuh:
Did you configure any REST zones?

spolischook:
I guess no, did we have doc about it?

spolischook:
Found it https://symfony.com/doc/master/bundles/FOSRestBundle/3-listener-support.html#zone-listener

xabbuh:
IIRC the attribute will only ever be set if any zones are configured. That's why we assume that the default value is `true` to cover the case where no zones have been configured and all listeners should be enabled all the time.

spolischook:
Yea it make sense, but it disappointing that twig cant find the template for the exception.
I guess that twig exception could be catcher and native page with exception should be shown.
But I'm already configured zones, so my issues is resolved.
Maybe someone have time and desire to fix this behavior.

xabbuh:
> but it disappointing that twig cant find the template for the exception.

Can you clarify that a bit? I am not sure I completely understand what behaviour you are talking about.

PR #2005, opened 2019-06-25
[DOC] Adds missing use statement in order to constraints

alterphp:
use statement is required to use Symfony constraints in the requirements annotations


GuilhemN:
Good catch thanks !

PR #2010, opened 2019-07-24
Update AbstractScalarParam.php

### LINKED ISSUE #1496: allowBlank and nullable no longer work together as they did in 1.0
Previously you could specify allowBlank=false and nullable=true, which would account for allowing the user to not specify an API parameter, but if they DO specify the parameter, that parameter cannot be blank.

Example:

```
* @RequestParam(name="token", requirements=".{1,200}", allowBlank=false, strict=true, nullable=true, description="User's token")
```

This worked correctly in 1.0 where `/api/call?token=` would throw an error, but `/api/call?` would be fine.

Now in 2.0, however, the above is throwing an exception when you do not specify the parameter at all.

"Parameter \"token\" of value \"\" violated a constraint \"This value should not be blank.\""

I think this is incorrect, because nullable was set to true, so the value should be allowed to be null. The parameter is incorrectly being flagged as blank, but it is in fact null since it was not included at all in the request.

Gonzalo1987:
Fixed #1496

I don't know if there is neccesary make any test... I'm not a usual contributor to this project, but I believe that this fix solve the compatibility problem between nullable=true and allowBlank=false 

Greetings!

### DIFF
--- a/Controller/Annotations/AbstractScalarParam.php
+++ b/Controller/Annotations/AbstractScalarParam.php
@@ -65,7 +65,9 @@ abstract class AbstractScalarParam extends AbstractParam
         }
 
         if (false === $this->allowBlank) {
-            $constraints[] = new NotBlank();
+            $notBlank = new NotBlank();
+            $notBlank->allowNull = $this->nullable;
+            $constraints[] = $notBlank;
         }
 
         // If the user wants to map the value, apply all constraints to every
--- a/Controller/Annotations/AbstractScalarParam.php
+++ b/Controller/Annotations/AbstractScalarParam.php
@@ -66,7 +66,7 @@ abstract class AbstractScalarParam extends AbstractParam
 
         if (false === $this->allowBlank) {
             $notBlank = new NotBlank();
-            if (isset($notBlank->allowNull)) {
+            if (property_exists(NotBlank::class, 'allowNull')) {
                 $notBlank->allowNull = $this->nullable;
             }
             $constraints[] = $notBlank;
--- a/Controller/Annotations/AbstractScalarParam.php
+++ b/Controller/Annotations/AbstractScalarParam.php
@@ -66,7 +66,9 @@ abstract class AbstractScalarParam extends AbstractParam
 
         if (false === $this->allowBlank) {
             $notBlank = new NotBlank();
-            $notBlank->allowNull = $this->nullable;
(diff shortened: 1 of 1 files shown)


Controller/Annotations/AbstractScalarParam.php
```diff
@@ -65,7 +65,9 @@ public function getConstraints()
         }
 
         if (false === $this->allowBlank) {
-            $constraints[] = new NotBlank();
+            $notBlank = new NotBlank();
+            $notBlank->allowNull = $this->nullable;
```
xabbuh: We need to do this conditionally as the `allowNull` option didn't exist before Symfony 4.3.

Gonzalo1987: I have added the isset($notBlank->allowNull), thanks!

Controller/Annotations/AbstractScalarParam.php
```diff
@@ -65,7 +65,11 @@ public function getConstraints()
         }
 
         if (false === $this->allowBlank) {
-            $constraints[] = new NotBlank();
+            $notBlank = new NotBlank();
+            if (isset($notBlank->allowNull)) {
```
xabbuh: Are you sure this works as expected? I would expect this to always evaluate to `false` (see also https://3v4l.org/sdhnT). Probably using `property_exists()` is what we need to use here.

Gonzalo1987: For test it, I have commented the $allowNull value of the Constraint class and I didn't get any new error, It worked like expected (still having the bug #1496, but not adding new exceptions). I can use the property_exists if it is cleaner so.

xabbuh: If we didn't use `property_exists()`, our code would depend on the default value of the `$allowNull` property of the constraint class.

Gonzalo1987: Thank you very much for your recomendation @xabbuh, I've modified the condition according your advise. I believe that now it's fine.

Gonzalo1987:
@xabbuh It's now ok?

Gonzalo1987:
Why this PR remains as 'Changes requested'?

GuilhemN:
Thank you @Gonzalo1987! :)

Issue #2002, opened 2019-06-07
Deprecate symfony/templating


phoenixgao:
After upgrading to Symfony 4.3 we're getting a lot of symfony/templating component related deprecation notice in FOSRestBundle.

`symfony/templating` is deprecated in Symfony 4.3 and will be removed in Symfony 5.0
https://symfony.com/blog/new-in-symfony-4-3-deprecated-the-templating-component-integration.

Tobion:
IMO we should try not to depend on twig or templating which should be possible with https://github.com/symfony/symfony/pull/31065 in sf 4.4 and which then also makes use of https://tools.ietf.org/html/rfc7807

Issue #1887, opened 2018-04-25
Twig is required unnecessary


Preclowski:
Hi, I think `twig` dependency should be moved from require-dev to require or made optional some way.

Even if we use custom exception controller, there is registered service `fos_rest.exception.twig_controller` with dependency to `templating.engine.twig`.

```xml
        <service id="fos_rest.exception.twig_controller" class="FOS\RestBundle\Controller\TwigExceptionController" parent="fos_rest.exception.controller">
            <argument type="service" id="templating.engine.twig" />
        </service>
```

It would be nice if twig will not be required to use fosrest 😄

stof:
but is this service *used* when Twig is not available ?

Preclowski:
I dont think so, `fos_rest.exception.controller` should be used by default. But it's declared and symfony throws exception 

> The service "fos_rest.exception.twig_controller" has a dependency on a non-
    existent service "templating.engine.twig".

xabbuh:
Which version of FOSRestBundle do you use?

xabbuh:
And what does your config look like?

magnetik:
Might be related : https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1772

LLCDBDBBSysUser:
I have the same problem. Fix is available in #1945. Would be nice if it would get merged.

PR #2012, opened 2019-07-30
Fix removal of fos_rest.exception.twig_controller when templating/twig is not enabled

### LINKED ISSUE #1772: Symfony Flex (3.3) - Dependency with Twig
When I want to enable exception handlers, I get the following exception :

The service "fos_rest.exception.twig_controller" has a dependency on a non-existent service "templating.engine.twig".

Tobion:
Fixes #1772, #1887, #1945, #2002

With that you can remove 

```
framework:
    templating:
        engines: twig
```

which fixes the deprecations and it will use `\FOS\RestBundle\Controller\ExceptionController` instead of `\FOS\RestBundle\Controller\TwigExceptionController` correctly. This in turn means it does not use the TwigBundle error rendering but the implementation using the Serializer (Symfony or JMS). But the error response stays the same because of `\FOS\RestBundle\Serializer\Normalizer\ExceptionHandler::convertToArray` (or the equivalent for Symfony serializer).
So people can disable templating and twig and it should still be BC. If people use TwigBundle but want to disable SF templating part to get rid of deprecations, they need to set

```
fos_rest:
    exception:
        exception_controller: 'fos_rest.exception.controller::showAction'
```

because otherwise the unwanted twig controller get's used again: https://github.com/FriendsOfSymfony/FOSRestBundle/blob/6013d5131f94cd7367c423e52bdfb83c4f0613f9/DependencyInjection/FOSRestExtension.php#L360-L367

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ CHANGELOG
 
 * deprecated using the `ParamFetcher` class without passing a validator as the third argument, this
   argument will become mandatory in 3.0
+* fix compatiblity without the deprecated templating in Symfony 4.3; see #2012 on how to configure the FOSRestBundle
 
 2.5.0
 -----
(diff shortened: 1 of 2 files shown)


Tobion:
Test failures unrelated and already failed before...

Tobion:
For Symfony 4.4 compatibility we likely can remove/deprecate parts of the exception controllers and listeners etc as it will be obsolete with the new ErrorRenderer component. But that is a future topic to handle.

xabbuh:
Thank you @Tobion.

PR #2015, opened 2019-07-31
raise php/sf requirement, make symfony/templating optional dependency, fix tests

Tobion:
Also fix some depreactions as `Symfony\Bundle\FrameworkBundle\Templating\EngineInterface` is deprecated but the extended `Symfony\Component\Templating\EngineInterface` is not. And we do not need any methods of the FrameworkBundle EngineInteraface.


Tobion:
I fixed all the 

> Other deprecation notices (172)

due to the templating by extracting tests using html templating rending into a legacy test and disable the templating in tests where it's not needed. Also fixed the default value for the exception_controller when twig is enabled but the templating sf integration is not.

Issue #2017, opened 2019-08-02
Compatibility Symfony 4.3


maxhelias:
The class [EventListener/ExceptionListener.php](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/EventListener/ExceptionListener.php) extends from [Symfony\Component\HttpKernel\EventListener\ExceptionListener](https://github.com/symfony/symfony/blob/4.3/src/Symfony/Component/HttpKernel/EventListener/ExceptionListener.php) and this last is marked final since Symfony 4.3.

```The "Symfony\Component\HttpKernel\EventListener\ExceptionListener" class is considered final since Symfony 4.3. It may change without further notice as of its next major version. You should not extend it from "FOS\RestBundle\EventListener\ExceptionListener".```

**Comment by pdugas** [comment]
+1 for a fix.  Just hit this after doing some updating.

**Comment by maxhelias** [comment]
I'm trying to fix myself but I do not know what the `ZONE_ATTRIBUTE` attribute does. Because the rest of the class is identical to the parent.
So I don't know whether to make a separate listener or rather decorate the parent class.

**Comment by killerwolf** [comment]
@maxhelias  zone narrows listeners to only API request.

exemple : 
when requesting api.mydomain.com/* all Listener are active and are inactive all other subdomain.

for this to work, you need to specify your zone with host, patterns.
```
  routing_loader:
    zone:
      - { host: 'api.mydomain.com' }
```

here is the complete documentation
https://symfony.com/doc/master/bundles/FOSRestBundle/3-listener-support.html#zone-listener

killerwolf:
The only difference between ExceptionListener from FOSRESTBundle and The one provided by symfony/http-kernel is the FlattenException in the duplicateRequest function.

is this usefull ?

sarcher:
@killerwolf That is correct, the difference is that `duplicateRequest()` is overloaded to provide a `FlattenException`. The problem is that the `ExceptionController` class itself takes an `\Exception`, and the `FlattenException` class does not extend the base PHP `\Exception` class.

Since Symfony's class is now final, it can't be extended, and since the `duplicateRequest()` method is protected, it can't just be decorated either.

[...]

PR #2018, opened 2019-08-08
[WIP] ExceptionListener: fix deprecations from SF 4.3

thePanz:


maxhelias:
I'm not sure but I think this is not the right approach because the `EventSubscriberInterface` interface registers a new action on `onKernelException` so `HttpKernelExceptionListener` is executed twice.
I would go on a decorator but I don't know the purpose of `ZONE_ATTRIBUTE`

maxhelias:
Related #2017

thePanz:
I agree, I thought it would be a straightforward fix, but it is not
Furthermore: the HttpKernelExceptionListener is passing a FlattenException instead of the \Exception object.

@maxhelias Feel free to close this PR or continue working on ti :)

maxhelias:
@thePanz, I do not know, I'm not the owner. Maybe that's the purpose to calling twice

XWB:
Should be fixed in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2025

Issue #2020, opened 2019-08-15
VIewHandler can't accept Twig in Symfony 4.3 and beyond


sarcher:
After upgrading to Symfony 4.3, it appears impossible to both avoid deprecation notices and use Twig as the default render engine for the `ViewHandler` class. This message is seen:

> An instance of Symfony\Component\Templating\EngineInterface must be injected in
> FOS\RestBundle\View\ViewHandler to render templates.

Since the constructor for `ViewHandler` type hints an `EngineInterface` , the following configuration will fail because Twig is now an instance of `Twig\Environment`:

```
fos_rest:
    service:
        templating: twig
```

One workaround is to extend the `ViewHandler` class and then replace the default implementation, but this seems like an ugly necessity to use things in their (more or less) default setup. 

So, either I am missing something, or there needs to be a fix here. My best guess would be to remove the type hint and add a check to see that the injected object is either `Twig\Environment` or an instance of `EngineInterface` or `null`. 

I feel like I am missing something, though, since this was not mentioned in the other deprecation fixes. Is there another solution here?

maxhelias:
Is your issue not related to this PR #2012 ? Or it's an other issue ?

sarcher:
Apologies, I should have mentioned this is occurring against the current master:
https://github.com/FriendsOfSymfony/FOSRestBundle/commit/e1f91266ca37e423dfae9ec4e641b247fdb2915d

So, this is after that PR has been applied.

isakrubin:
Same issue, using dev-master and ViewHandler still type-hints "EngineInterface $templating = null" :/

sarcher:
This was closed by #2022

PR #2022, opened 2019-09-12
Inject Twig by default instead of EngineInterface for Symfony >= 4.3

sarcher:
This addresses #2020 by:

* Changing the default `services.templating` value from `templating` to `twig` if Symfony is >= 4.3 (it can still be overridden in `symfony/templating` is present
* Removing the type hint from `TwigExceptionController` and `ViewHandler` to allow either `Twig\Environment` or `EngineInterface` to be passed
* Borrowing the bridge code from TwigBundle to handle the new Twig template lookup via try/catch

Seems to work in 4.2 and 4.3, I tried it in both; we'll see what the CI says. 

Not sure this is the best approach but I needed something working and so far so good. Please let me know if this needs to look different. I considered just having two separate classes for with/without the templating component, but this is far less code to maintain.

### DIFF
--- a/Controller/TemplatingExceptionController.php
+++ b/Controller/TemplatingExceptionController.php
@@ -16,6 +16,7 @@ use FOS\RestBundle\View\ViewHandlerInterface;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Templating\EngineInterface;
 use Symfony\Component\Templating\TemplateReferenceInterface;
+use Twig\Environment;
 
 abstract class TemplatingExceptionController extends ExceptionController
 {
@@ -25,8 +26,18 @@ abstract class TemplatingExceptionController extends ExceptionController
         ViewHandlerInterface $viewHandler,
         ExceptionValueMap $exceptionCodes,
         $showException,
-        EngineInterface $templating
+        $templating
     ) {
+        if (!$templating instanceof EngineInterface && !$templating instanceof Environment) {
+            throw new \TypeError(sprintf(
+                'The fourth argument of %s must be an instance of %s or %s, but %s was given.',
+                __METHOD__,
+                EngineInterface::class,
+                Environment::class,
+                is_object($templating) ? get_class($templating) : gettype($templating)
+            ));
+        }
+
         parent::__construct($viewHandler, $exceptionCodes, $showException);
 
         $this->templating = $templating;
--- a/Controller/TwigExceptionController.php
+++ b/Controller/TwigExceptionController.php
@@ -12,6 +12,10 @@
 namespace FOS\RestBundle\Controller;
 
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\Templating\EngineInterface;
+use Twig\Environment;
+use Twig\Error\LoaderError;
+use Twig\Loader\ExistsLoaderInterface;
 
 /**
  * Custom ExceptionController that uses the view layer and supports HTTP response status code mapping.
@@ -48,14 +52,20 @@ class TwigExceptionController extends TemplatingExceptionController
         // For error pages, try to find a template for the specific HTTP status code and format
         if (!$showException) {
             $template = sprintf('@Twig/Exception/%s%s.%s.twig', $name, $statusCode, $format);
-            if ($this->templating->exists($template)) {
+            if (
+                ($this->templating instanceof EngineInterface && $this->templating->exists($template)) ||
+                ($this->templating instanceof Environment && $this->templateExists($template))
+            ) {
                 return $template;
             }
         }
 
         // try to find a template for the given format
         $template = sprintf('@Twig/Exception/%s.%s.twig', $name, $format);
-        if ($this->templating->exists($template)) {
+        if (
+            ($this->templating instanceof EngineInterface && $this->templating->exists($template)) ||
+            ($this->templating instanceof Environment && $this->templateExists($template))
+        ) {
             return $template;
         }
 
@@ -63,5 +73,28 @@ class TwigExceptionController extends TemplatingExceptionController
         $request->setRequestFormat('html');
 
         return sprintf('@Twig/Exception/%s.html.twig', $showException ? 'exception_full' : $name);
+    }
+
+    /**
+     * See if a template exists using the modern Twig mechanism.
+     *
+     * This code is based on TwigBundle and should be removed when the minimum required
+     * version of Twig is >= 3.0. See src/Symfony/Bundle/TwigBundle/Controller/ExceptionController.php
+     */
+    private function templateExists(string $template): bool
+    {
+        $loader = $this->templating->getLoader();
+        if ($loader instanceof ExistsLoaderInterface || method_exists($loader, 'exists')) {
+            return $loader->exists($template);
+        }
+
+        try {
+            $loader->getSourceContext($template)->getCode();
+
+            return true;
+        } catch (LoaderError $e) {
+        }
+
+        return false;
     }
 }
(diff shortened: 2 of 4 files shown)


alexander-schranz:
@mickel1138 This change is needed as templating configuration is deprecated and will be removed in symfony 5.0.

fabpot:
For the record, I've tested this PR on several apps, and it's indeed needed and works well.

Controller/TemplatingExceptionController.php
```diff
@@ -25,7 +24,7 @@ public function __construct(
         ViewHandlerInterface $viewHandler,
         ExceptionValueMap $exceptionCodes,
         $showException,
-        EngineInterface $templating
+        $templating
```
xabbuh: I think we should add a check which ensure that `$templating` is either an implementation of the `EngineInterface` or an instance of `Twig\Environment` and throw a `TypeError` otherwise.

Controller/TwigExceptionController.php
```diff
@@ -64,4 +74,33 @@ protected function findTemplate(Request $request, $statusCode, $showException)
...
+     * See if a template exists using the modern Twig mechanism.
+     *
+     * This code is based on TwigBundle and should be removed when the minimum required
+     * version of Twig is >= 3.0. See src/Symfony/Bundle/TwigBundle/Controller/ExceptionController.php
+     *
+     * @param string $template
+     *
+     * @return bool
+     */
+    protected function templateExists($template)
```
xabbuh: Should be `private` and we can make use of type declarations like this:

```php
private function templateExists(string $template): bool
```

DependencyInjection/Compiler/TwigExceptionPass.php
```diff
@@ -39,7 +39,11 @@ public function process(ContainerBuilder $container)
         }
 
         if (!$container->has('templating.engine.twig')) {
-            $container->removeDefinition('fos_rest.exception.twig_controller');
+            if ($container->has('twig') && $container->has('fos_rest.exception.twig_controller')) {
+                $container->getDefinition('fos_rest.exception.twig_controller')->replaceArgument(3, $container->getDefinition('twig'));
```
xabbuh: If we use `has()` instead of `hasDefinition()` inside the `if` (which is totally fine), we should use `findDefinition()` instead of `hasDefinition()` here.

DependencyInjection/Configuration.php
```diff
@@ -137,7 +138,13 @@ public function getConfigTreeBuilder()
                     ->addDefaultsIfNotSet()
                     ->children()
                         ->scalarNode('router')->defaultValue('router')->end()
-                        ->scalarNode('templating')->defaultValue('templating')->end()
+                        ->scalarNode('templating')->defaultValue(function () {
+                            if (Kernel::VERSION_ID >= 40300) {
```
xabbuh: The version of the kernel may be different than the version of other installed Symfony components. What component are we actually interested in to make the decision here?

View/ViewHandler.php
```diff
@@ -132,6 +133,15 @@ public function __construct(
         $defaultEngine = 'twig',
         array $options = []
     ) {
+        if (!is_null($templating) && !$templating instanceof EngineInterface && !$templating instanceof Environment) {
```
xabbuh: we usually use `null !== $templating` instead of calling `is_null()`

View/ViewHandler.php
```diff
@@ -132,6 +133,15 @@ public function __construct(
...
+            throw new \InvalidArgumentException(sprintf(
```
xabbuh: Let's throw a `TypeError` here instead. That will be consistent with how it's done in the Symfony code.

DependencyInjection/Configuration.php
sarcher: Good question; the thought was that if somebody has a kernel >= 4.3 and have not specified `templating` explicitly, they probably--but may not always be--using Twig as this is the configured default. The only other relevant component, I think, is the FrameworkBundle, but is it actually possible to have a different version of the kernel and FrameworkBundle?

There is probably some outlier here where somebody has a strange combination of things in a project for some reason, but we probably can't fully guess at that. In those instances they would just need to manually specify this option.

DependencyInjection/Compiler/TwigExceptionPass.php
sarcher: Makes sense; looks like `has()` includes aliases and `getDefinition()` does not, so good to be consistent. All set.

Controller/TwigExceptionController.php
sarcher: Will do; I had just copied the code directly from `TwigBundle`, wasn't sure how much it should be doctored up.

Controller/TemplatingExceptionController.php
sarcher: I added this in 3a18517434573e422ebac63406aa408b73229690, I also wrote a test but saw there weren't any really for stuff like that; if you want it, let me know and I can add it (see it here: https://gist.github.com/sarcher/300c605623ab131d7b78c821a99b0e34)

Controller/TwigExceptionController.php
```diff
@@ -64,4 +74,33 @@ protected function findTemplate(Request $request, $statusCode, $showException)
...
+     * This code is based on TwigBundle and should be removed when the minimum required
+     * version of Twig is >= 3.0. See src/Symfony/Bundle/TwigBundle/Controller/ExceptionController.php
+     *
+     * @param string $template
+     *
+     * @return bool
+     */
+    private function templateExists(string $template): bool
+    {
+        $template = (string) $template;
```
xabbuh: looks like this is not needed anymore since we have the parameter type declaration now :)

sarcher: Oops, missed that!

DependencyInjection/Configuration.php
fabpot: I would not change the default in this PR and keep that for another one. In a followup PR, I would deprecate this option as it's not relevant anymore and should be removed in the next major version of this bundle, where Twig will always be used.

Controller/TwigExceptionController.php
```diff
@@ -64,4 +74,31 @@ protected function findTemplate(Request $request, $statusCode, $showException)
...
         return sprintf('@Twig/Exception/%s.html.twig', $showException ? 'exception_full' : $name);
     }
+
+    /**
+     * See if a template exists using the modern Twig mechanism.
+     *
+     * This code is based on TwigBundle and should be removed when the minimum required
+     * version of Twig is >= 3.0. See src/Symfony/Bundle/TwigBundle/Controller/ExceptionController.php
+     *
+     * @param string $template
```
fabpot: Could be removed

Controller/TwigExceptionController.php
```diff
@@ -64,4 +74,31 @@ protected function findTemplate(Request $request, $statusCode, $showException)
...
+     *
+     * @return bool
```
fabpot: Could be removed as well

DependencyInjection/Configuration.php
sarcher: Okay; I've reverted that one.

Before I make that next PR, my understanding then is that the default could change along with a deprecation notice in 2.6, and then in yet another PR , remove it entirely for 3.x. (ex: "Specifying a templating engine other than Twig is deprecated in 2.6 and will be removed in 3.0")

Along with that change for 3.x, I assume that we would remove all references to `EngineInterface` which would simplify a lot of the other code. I haven't played around with master for this bundle yet but could certainly do that once I have a Symfony 5 RC running locally.

sarcher:
While we're at it, if anybody could take a look at #2017 and provide some guidance, I'd appreciate it. I'm migrating a rather large project to Symfony 5 to see what I can find but this one is blocking me since that extension point is now final. I have no good idea of why we are bypassing the flatten exception so I don't want to change things blindly.

DependencyInjection/Configuration.php
fabpot: That's correct.

PR #2023, opened 2019-10-08
Fix compatibility to new Symfony FlattenException

alexander-schranz:
See https://github.com/symfony/symfony/blob/4.4/src/Symfony/Component/ErrorRenderer/Exception/FlattenException.php


alexander-schranz:
Not sure if this PR is longer needed as `create` was also added to the new FlattenException in https://github.com/symfony/symfony/commit/928363c408c036cd385e402f9a816da73db05f3e#diff-234309116eb79dcc9603b3a49669e499 by @yceruto not sure which function will stay in 5.0 or if there will still be both.

xabbuh:
Thank you @alexander-schranz.

Issue #1185, opened 2015-10-22
parent/child resource error: Every parent controller must have `get{SINGULAR}Action($id)` method


pstephenson02:
In [this gist](https://gist.github.com/pstephenson02/d254712a8968d37fb4d8), I have 3 controllers defined. I want to establish a parent/child relationship between the cah_annotation and the standard_proposal resource. I cannot find the reason for this error:

``` bash
$ ./app/console debug:router


  [Symfony\Component\Config\Exception\FileLoaderLoadException]                                                    
  Every parent controller must have `get{SINGULAR}Action($id)` method                                             
  where {SINGULAR} is a singular form of associated object in /var/www/html/cdpaccess/src/ICC/ProposalBundle/Res  
  ources/config/routing.yml (which is being imported from "/var/www/html/cdpaccess/app/config/routing.yml").      


  [InvalidArgumentException]                                           
  Every parent controller must have `get{SINGULAR}Action($id)` method  
  where {SINGULAR} is a singular form of associated object             
```

The weird thing is, if I modify routing.yml and set the parent resource on cah_annotation to proposal:

``` yml
cah_annotation:
    parent: proposal
```

It works fine. Even if I delete the proposal resource entry all together I get the same error.

What am I doing wrong here? These controllers are _very_ similar, how is it that it works with the proposal resource but not standard_proposal?

``` bash
$ php composer.phar show -i | grep rest
friendsofsymfony/rest-bundle               1.7.2                 This Bundle provides various tools to rapidly...
```

xabbuh:
Can you check #1173?

pstephenson02:
I did first check https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1173 and its associated [stackoverflow](http://stackoverflow.com/questions/33087657/every-parent-controller-must-have-getsingularactionid-method-when-i-have) question.

I made sure that was not the issue here. The method parameters are defined correctly (or rather: are not incorrect in the way they were shown in the stack question)

xabbuh:
@pstephenson02 Instead of `getStandardProposalAction()` try `getAction()` instead.

xabbuh:
Though I think the error message should indeed be improved.

Tobion:
Closing as automatic route generation is deprecated: #1954

Issue #1954, opened 2018-11-07
RFC: Deprecate route creation based on class/method names


Tobion:
IMO the automagic creation of routes based on the class and method names is the worst part of the rest bundle.

1. it's alot of magic and things easily break just by renaming a class or method
2. it's hard to understand, document and maintain
3. it does not work with the new best practice of symfony to not suffix your contoller actions with `*Action` because of https://github.com/FriendsOfSymfony/FOSRestBundle/blob/1665b98c6543cf52699f6a1558121ca54c7565b2/Routing/Loader/Reader/RestActionReader.php#L475-L477
But we also cannot really remove this limitation because it would automagically create routes for methods you might not want, like setters if you use setter injection in some places
4. it does not work with `controller.service_arguments`, see #1807
5. there are so many issues and edge cases with this
    - #1759 is a huge problem. but it cannot really be changed without breaking bc. if we change the route names, it would be a bc break.
    - #1953
    - #1837 and #1733
    - #1808
    - #1789
    - #1310 
    - #1657

So I'd like a way to use the RestBundle routing features like 
- `FOS\RestBundle\Controller\Annotations\Get` and all the other method shortcuts 
- `FOS\RestBundle\Controller\Annotations\Version` 
- adding the format to the routes path

But without the route magic of creating routes based on class/method names. So routes need to be created explicitly using the symfony core routing annotations or the fos rest bundle shortcuts. This also means no more
- `FOS\RestBundle\Controller\Annotations\RouteResource`
- `FOS\RestBundle\Routing\ClassResourceInterface`

We could either deprecate this magic or create a config option to disable it.

lsmith77:
I am not sure I understand .. you can use 
https://symfony.com/doc/master/bundles/FOSRestBundle/7-manual-route-definition.html and the annotations listed there without having to disable the magic way .. so I am not sure what you are asking here.

do you want to deprecate it out of principle or because its actually making something impossible for you?

lsmith77:
ok. so how about we add new class level annotations to configure this?

lsmith77:
BTW as for deprecating .. I am open to this. I think anyone that really is focused on RAD is by now anyway using API Platform and the main use for this Bundle are people that are doing a lot of custom things that likely do not really benefit from the magic.

Tobion:
> ok. so how about we add new class level annotations to configure this?

[...]

PR #2024, opened 2019-10-18
Ignore SessionInterface and UserInterface controller action arguments

net02:
Route creation relies on controller actions' method arguments for a couple of things.. as such, some commonly injected arguments [are already filtered](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/e1f91266ca37e423dfae9ec4e641b247fdb2915d/Routing/Loader/Reader/RestActionReader.php#L530-L536). 
This PR includes two more commonly injected interfaces to the exclusion list, as seen in the [list of Symfony's built-in arguments resolvers](https://symfony.com/doc/3.4/controller/argument_value_resolver.html#built-in-value-resolvers).
 
Route name and path can be easily overwritten so this PR is usually easily workaround-able, but there's still one feature which relies on action arguments: controller hierarchy.
It needs at least one method [having exactly one argument](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/e1f91266ca37e423dfae9ec4e641b247fdb2915d/Routing/Loader/Reader/RestActionReader.php#L273-L275) on the parent controller, else you'd face a cryptic exception (#1185) later on during route generation for the child controller.

Moreover, aside from this PR, I feel there still is an open point with "uncommon" injected services (#1733) as sometimes injecting everything in the constructor may not be an option. Using the `@RouteResource` annotation was proposed in the past (#1764) but maybe it lacked the *"this fixes relying on having at least a method with exactly one argument"*.

If @Tobion agrees I'd like to work on an additional PR, following one or both of these approaches:
- resurrecting #1764, adding some test and writing a more helpful error message for the hierarchy exception pointing to the annotation;
- exclude arguments which are injected services, not by config as in #1733 but mimic-ing the [service injector itself](https://github.com/symfony/symfony/blob/3.4/src/Symfony/Component/HttpKernel/Controller/ArgumentResolver/ServiceValueResolver.php#L66-L75) to identify those arguments - we should already have everything that's need (container, `_controller` and the argument name).

### DIFF
--- a/Routing/Loader/Reader/RestActionReader.php
+++ b/Routing/Loader/Reader/RestActionReader.php
@@ -20,7 +20,9 @@ use FOS\RestBundle\Routing\RestRouteCollection;
 use Psr\Http\Message\MessageInterface;
 use Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter;
 use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Session\SessionInterface;
 use Symfony\Component\Routing\Route;
+use Symfony\Component\Security\Core\User\UserInterface;
 use Symfony\Component\Validator\ConstraintViolationListInterface;
 
 /**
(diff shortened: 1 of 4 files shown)


Tobion:
I think the whole route creation based on class/method/arguments is a broken concept, see #1954
So if it was my decision, I would remove all of this instead of adding more workarounds.

net02:
Aye, I stumbled upon that issue and I too wish there were already an alternative way to make it work.
But as of now, and for backwards compatibility until route creation based on class/method/arguments is dropped, it would obviously be better working than broken :wink:. 

Our workaround at the moment is literally adding in the parent controller a dummy action for a verb we're not using :nauseated_face:.
Since arguments injection is a thing in Symfony, I don't really see this as a workaround to be honest. It's actually fixing the - somewhat questionable - implementation.

net02:
Any thought on this @xabbuh @GuilhemN?

PR #2026, opened 2019-10-29
Update EngineInterface error message

XWB:
I just got this error because I removed the `framework.templating` setting:

> An instance of Symfony\Component\Templating\EngineInterface must be injected in FOS\RestBundle\View\ViewHandler to render templates.


View/ViewHandler.php
```diff
@@ -363,7 +363,12 @@ public function createRedirectResponse(View $view, $location, $format)
     public function renderTemplate(View $view, $format)
     {
         if (null === $this->templating) {
-            throw new \LogicException(sprintf('An instance of %s must be injected in %s to render templates.', EngineInterface::class, __CLASS__));
+            throw new \LogicException(sprintf(
```
Tobion: please update the constructor annotation `@param EngineInterface|Environment $templating`

XWB: @Tobion Done.

PR #2027, opened 2019-10-30
Use Twig engine for unit tests

XWB:
Because the Templating Component has been deprecated.


Tobion:
The templating component has not been deprecated. The FrameworkBundle integration has.
So this PR does not fix any deprecation does it? We can do this change anyway but it's not really necessary as far as I can see at the moment.

XWB:
You are right. I must say the use of `EngineInterface` confused me.

PR #2028, opened 2019-10-30
Use twig as templating engine when available

alexander-schranz:
Currently if `symfony/templating` is not installed the following error could happen:

> An instance of Symfony\Component\Templating\EngineInterface must be injected in FOS\RestBundle\View\ViewHandler to render templates.

thats because templating is set as default value. If templating is not available but twig it should use twig instead.

### DIFF
--- a/DependencyInjection/Compiler/ConfigurationCheckPass.php
+++ b/DependencyInjection/Compiler/ConfigurationCheckPass.php
@@ -39,10 +39,10 @@ final class ConfigurationCheckPass implements CompilerPassInterface
 
         if (!$container->has((string) $container->getAlias('fos_rest.templating'))) {
             $container->removeAlias('fos_rest.templating');
-        }
 
-        if ($container->hasDefinition('twig')) {
-            $container->setAlias('fos_rest.templating', 'twig');
+            if ($container->hasDefinition('twig')) {
+                $container->setAlias('fos_rest.templating', 'twig');
+            }
         }
     }
 }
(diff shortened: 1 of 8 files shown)


Tobion:
#2022 does not fix this right?

alexander-schranz:
@Tobion Thx for the review, you are right,  this currently overwrites it.
And no #2022 does not fix it as it still defaults to `templating` currently at https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/DependencyInjection/Configuration.php#L140 and there is no fallback to twig implemented.

XWB:
@Tobion No https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2022 does not fix the issue. The problem is that FOSRestBundle always worked without any configuration because EngineInterface was always injected into `ViewHandler`. After https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2022 got merged I got the following error:

> An instance of Symfony\Component\Templating\EngineInterface must be injected in FOS\RestBundle\View\ViewHandler to render templates.

I fixed it by creating `fos_rest.yaml` and configuring the templating engine:

```yaml
fos_rest:
    service:
        templating: twig
```

In my opinion FOSRestBundle should work without configuration, like it did before.

alexander-schranz:
@XWB yes this should be fixed with this PR. Need to check how I can pack this into a test case.

Tobion:
Thanks @alexander-schranz

PR #2025, opened 2019-10-21
Remove deprecation warning about final ExceptionListener

yceruto:
Replaces https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2018

PR #2032, opened 2019-11-22
Remove deprecated notice about controller_name_converter

maxhelias:


maxhelias:
@Tobion I do not know how to fix the failure in PHP 7.3.. a little help would be a pleasure :)

sarcher:
The failures actually have to do with Symfony 4,4/5, not with PHP 7.3; it looks like the Travis configuration is a little wonky. 7.1 and 7.2 are testing with the 3.X LTS which is 3.4. PHP 7.3 is testing against the latest but seems to be pulling a mixture of 4.4 and 5.0 components.

I haven't fully looked at this PR but I would say that of the reported failures, the first has to do with Twig 3.0, and the remaining three are likely solved by #2025 when it gets merged.

maxhelias:
@sarcher Thanks for the information :)
I blocked the Twig 3.0 update and the error is no longer there. I don't think it's the right solution but I don't know what it's all about, so I'm going to create a new issue
And the other failure, should be fixed by #2025

XWB:
@maxhelias As https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2025 has been merged, can you rebase?

maxhelias:
@XWB Done ! 
Btw, do you want to keep the Twig 3 conflit ? I'm not sur about that and i don't create a issue for

maxhelias:
Fixed in #2036

PR #2035, opened 2019-11-26
allow Symfony 5 components

xabbuh:


.travis.yml
```diff
@@ -31,6 +34,7 @@ matrix:
 before_install:
   - echo "memory_limit=4G" >> ~/.phpenv/versions/$(phpenv version-name)/etc/conf.d/travis.ini
   - phpenv config-rm xdebug.ini || true
+  - composer global require --no-progress --no-scripts --no-plugins symfony/flex dev-master
```
maxhelias: I am not sure but if we use Flex, maybe we can also use the constraint `extra.symfony.version` in place `SYMFONY_REQUIRE` ? And edit composer.json on each stage with `jq`

XWB: Or you simply add `composer config extra.symfony.require $SYMFONY_REQUIRE` instead of introducing flex to travis.

Form/Extension/DisableCSRFExtension.php
```diff
@@ -68,7 +68,7 @@ public function getExtendedType()
             ;
     }
 
-    public static function getExtendedTypes()
+    public static function getExtendedTypes() : iterable
```
GuilhemN: I did this change, I believe it is an acceptable bc break as it is not a class meant to be extended imo (should it be made `@final` by the way ?).

xabbuh: LGTM

PR #2053, opened 2020-01-20
Probably a typo in example "getAction"

extrablind:


XWB:
:+1:

xabbuh:
This change looks wrong to me. See the explaining sentence above the code example:

> There is a special convention to call the methods getAction and cgetAction, where the c stands for collection.

PR #2056, opened 2020-01-24
Update 3-listener-support.rst

sela:
f.e. isn't correct. change to i.e.

### DIFF
--- a/Resources/doc/3-listener-support.rst
+++ b/Resources/doc/3-listener-support.rst
@@ -5,7 +5,7 @@ Step 3: Listener support
 various events from decoding the request content in the request (body listener),
 determining the correct response format (format listener), reading parameters
 from the request (parameter fetcher listener), to formatting the response either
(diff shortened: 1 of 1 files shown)


Resources/doc/3-listener-support.rst
```diff
@@ -5,7 +5,7 @@ Step 3: Listener support
 various events from decoding the request content in the request (body listener),
 determining the correct response format (format listener), reading parameters
 from the request (parameter fetcher listener), to formatting the response either
-with a template engine like twig or to f.e. xml or json using a serializer (view
+with a template engine like twig or to i.e. xml or json using a serializer (view
```
xabbuh: Didn't you mean `e.g.`?

fxmauricard: As _f.e._ seams to be for _for exemple_, and that _e.g._ is for __*exempli gratia*__ "for example", but _i.e._ is for __*id est*__ "that is to say", I think you're right. ;)

PR #2060, opened 2020-01-29
Fixed explicit disabling of serializer max depth checks

cezarystepkowski:
JMS Serializer is configured with max depth checking enabled when method `disableMaxDepth` is called in `FOS\RestBundle\Context\Context`.

This PR fixes the behaviour to actually configure serializer without max depth checking enabled in this scenario.

### DIFF
--- a/Serializer/JMSSerializerAdapter.php
+++ b/Serializer/JMSSerializerAdapter.php
@@ -104,7 +104,7 @@ class JMSSerializerAdapter implements Serializer
         if (null !== $context->getGroups()) {
             $jmsContext->setGroups($context->getGroups());
         }
-        if (null !== $context->getMaxDepth(false) || null !== $context->isMaxDepthEnabled()) {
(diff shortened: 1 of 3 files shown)


Serializer/JMSSerializerAdapter.php
```diff
@@ -104,7 +104,7 @@ private function convertContext(Context $context, $direction)
         if (null !== $context->getGroups()) {
             $jmsContext->setGroups($context->getGroups());
         }
-        if (null !== $context->getMaxDepth(false) || null !== $context->isMaxDepthEnabled()) {
+        if (null !== $context->getMaxDepth(false) || (null !== $context->isMaxDepthEnabled() && true === $context->isMaxDepthEnabled())) {
```
XWB: As you are doing a strict type check, I believe the null check can be removed?

cezarystepkowski: You're right, I've removed it.

xabbuh:
Thank you @cezarystepkowski.

PR #2084, opened 2020-02-07
add Travis CI job using Symfony 5 components

xabbuh:


Tobion:
I've seen you started working on SF 5 compatibility.
Do you have a plan already what is going to get removed in version 3?

For example, I wonder what the whole templating part has to do in a rest-bundle? Can't we just say if you want to render html etc with templating, then do not use the view handler system of the rest-bundle? Then there is no point in having any templating logic in a rest-bundle or am I missing something?

If I read the documentation https://symfony.com/doc/master/bundles/FOSRestBundle/2-the-view-layer.html. Why would I need a rest-bundle to render a twig template? I can do that with symfony core very conveniently.

Resources/doc/annotations-reference.rst
```diff
@@ -74,7 +74,6 @@ View
 
     /**
      * @View(
-     *  templateVar="",
      *  statusCode=null,
      *  serializerGroups={},
      *  populateDefaultVars=true,
```
Tobion: this can be removed

Tobion: it can also be removed from the view annotation

EventListener/ViewResponseListener.php
```diff
@@ -72,9 +70,6 @@ public function onKernelView($event)
         }
 
         if ($configuration instanceof ViewAnnotation) {
```
Tobion: `$configuration = $request->attributes->get('_template');` where does the _template attribute get set? It seems the Sensio FrameworkExtraBundle can set this attribute. But how can it contain the FosRestBudnle view annotation?

xabbuh: IIRC that's happening because or `@View` annotation extends the `@Template` annotation from the SensioFrameworkExtraBundle.

Tobion: I see. I think we should also remove this inheritance.

Resources/doc/annotations-reference.rst
xabbuh: done in #2089

PR #2085, opened 2020-02-07
Fixed used alias for "FOS\RestBundle\View\ViewHandlerInterface"

### LINKED ISSUE #2029: AbstractFOSRestController break view handler registration
Hi, i hope that is not duplicate issue.
i had API based on deprecated FOSRestController which is related to Symfony 4.2 "micro-container" controller and commands.
So i decide to get rid of that deprecations and prepare for Symfony 5.x as soon as possible.
i switched to AbstractFOSRestController and simply ported single service i need in controller (that is part of my BaseApiController, so fos_rest.view_handler is registered from AbstractFOSRestController):

```php
/**
     * @return array
     */
    public static function getSubscribedServices()
    {
        return array_merge(
            parent::getSubscribedServices(),
            [
                AccountContextInterface::class => '?'.AccountContextInterface::class,
             ]
        );
    }
```

What break ?

When using AbstractFOSRestController my custom view handler which sign preformated S3 / Cloudfront urls at response level is not even registered in container.
I am using xml syntax and that is working with FOSRestController:

```xml
<service id="App\Infrastructure\Symfony\Rest\JsonViewHandler" class="App\Infrastructure\Symfony\Rest\JsonViewHandler" autoconfigure="true" autowire="true">
        </service>
        <service id="app_fos_rest.view_handler" parent="fos_rest.view_handler.default">
            <call method="registerHandler">
                <argument>json</argument>
                <argument type="collection">
                    <argument type="service" id="App\Infrastructure\Symfony\Rest\JsonViewHandler"></argument>
                    <argument>createResponse</argument>
                </argument>
            </call>
        </service>
```
```yaml
fos_rest:
    service:
        view_handler: app_fos_rest.view_handler
```

with AbstractFOSRestBundle .. registerHandler is not even called. however if i provide wrong service id then container complain so it is using app_fos_rest.view_handler but not calling "call"
that is 1:1 example based on official RSSFeed view handler => https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Resources/doc/examples/RssHandler.php

enricog84:
Changed `FOS\RestBundle\View\ViewHandlerInterface` to be an alias of "fos_rest.view_handler" instead of "fos_rest.view_handler.default". So in case if a custom ViewHandler has been configured under "fos_rest.service.view_handler" it will be used. Before this change, the used ViewHandler of Controller extending the `AbstractFOSRestController` would always have been pointing to `fos_rest.view_handler.default` because the service subscriber definition is `$subscribedServices['fos_rest.view_handler'] = ViewHandlerInterface::class;`.

This should fix #2029

### DIFF
--- a/Resources/config/view.xml
+++ b/Resources/config/view.xml
@@ -20,7 +20,7 @@
             <tag name="kernel.reset" method="reset" />
         </service>
 
-        <service id="FOS\RestBundle\View\ViewHandlerInterface" alias="fos_rest.view_handler.default" />
+        <service id="FOS\RestBundle\View\ViewHandlerInterface" alias="fos_rest.view_handler" />
 
         <service id="fos_rest.view_handler.jsonp" class="FOS\RestBundle\View\JsonpHandler" public="false">
             <argument /> <!-- JSONP callback parameter -->


XWB:
@enricog84 Can you apply this fix to the 2.x branch?

xabbuh:
Thank you @enricog84.

PR #2088, opened 2020-02-08
deprecate the templating integration

xabbuh:


Tests/Functional/app/ViewResponseListener/config.yml
```diff
@@ -13,8 +13,6 @@ fos_rest:
         formats:
             xml:  true
             json: true
-        templating_formats:
-            html: true
     format_listener:
         rules:
             - { path: ^/, priorities: [ html, json, xml ], fallback_format: ~, prefer_extension: true }
```
yceruto: can the `html` format be removed from here too?

lsmith77:
I think there might still be a few features lurking around in the ViewHandler that are only useful for HTML rendering. I briefly looked over the code and one feature that I noticed was `force_redirects`. This was basically about if when creating a resource if one should return the URL to the resource or redirect to the resource. The later bit I would say doesn't really make sense for a pure REST API

Tests/Functional/app/ViewResponseListener/config.yml
xabbuh: I tried to do that initially but that changed the status code of the response. However, deprecating the force redirects option as suggested by @lsmith77 solve this. We now have to dedicated test cases for this feature.

xabbuh:
@lsmith77 Good point, I also deprecated setting the `force_redirects` option.

PR #2089, opened 2020-02-10
remove templating support

xabbuh:


patrickbussmann:
One question, @xabbuh:
How to use normal templates at the moment?

With my configuration:
```json
# Read the documentation: https://symfony.com/doc/master/bundles/FOSRestBundle/index.html
fos_rest:
    exception:
        enabled: true
    routing_loader: true
    disable_csrf_role: ROLE_USER
    view:
        view_response_listener: true
        formats:
            json: true
    #    param_fetcher_listener:  true
    #    allowed_methods_listener:  true
    #    exception:
    #        codes:
    #            App\Exception\MyException: 403
    #        messages:
    #            App\Exception\MyException: Forbidden area.
    format_listener:
        rules:
            - { path: ^/api, prefer_extension: true, fallback_format: json, priorities: [ json, html ] }
```

I always get this error on login page.

http://localhost/public/login

    Format 'html' not supported, handler must be implemented

And even when I set all things to false.. I still get this error message.

Thanks in Advance!

xabbuh:
I am not sure if I understand what you mean. Can you give some more details about the problem that you have?

patrickbussmann:
Its a bit complicated. I try to find out whats exactly the problem.
But at the moment I think its this "Symfony5 is new and nothing is really compatible" thing.

Yesterday I integrated `HWIOAuthBundle` in my project which using `FOSRestBundle`.
Before I opened a page like:

    http://localhost/public/users

And when I'm not logged in, I will see an error of `FOSRestBundle`. Very pretty as JSON.
After the implementation there is a redirect.

First - yesterday when I created the comment above - I tought its a problem of this bundle.
Because the exception listener seems not to be called anymore.
But now I think its a problem of the other bundle.


Sorry for disturbing.
I think i must search for the problem and fix it by myself via Fork. Now maybe FOSOAuthServerBundle is wrong.

xabbuh:
Thank you for the feedback. If you think you found some other bug, feel free to open a new issue.

PR #2091, opened 2020-02-11
get and set ignored classes

### LINKED ISSUE #2090: RestActionReader ignoring user interface in 2.7.0
https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2024 introduced ignoring the UserInterface class as controller argument
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/ab328c73cadc1ee7a102bd4b6a90af5a3b213c10/Routing/Loader/Reader/RestActionReader.php#L532-L540

this breaks all configured children from a ``something_user_route`` parent because ignoring it will cause the following condition to be false
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/ab328c73cadc1ee7a102bd4b6a90af5a3b213c10/Routing/Loader/Reader/RestActionReader.php#L275
and without a singular name all children will trigger this exception
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/ab328c73cadc1ee7a102bd4b6a90af5a3b213c10/Routing/Loader/Reader/RestActionReader.php#L255

i do think it's fairy common to have a user entity to be part of a restful path?

anyway, any suggestions how to work around this issue without having to refactor a large amount of controllers?

wickedOne:
allow ignored classes to be set in the rest action reader

fixes #2090

### DIFF
--- a/Routing/Loader/Reader/RestActionReader.php
+++ b/Routing/Loader/Reader/RestActionReader.php
@@ -185,9 +185,12 @@ class RestActionReader
     /**
      * Set ignored classes.
      *
+     * These classes will be ignored for route construction when
+     * type hinted as method argument.
+     *
      * @param array $ignoredClasses
      */
-    public function setIgnoredClasses(array $ignoredClasses)
+    public function setIgnoredClasses(array $ignoredClasses): void
     {
         $this->ignoredClasses = $ignoredClasses;
     }
@@ -197,7 +200,7 @@ class RestActionReader
      *
      * @return array
      */
-    public function getIgnoredClasses()
+    public function getIgnoredClasses(): array
     {
         return $this->ignoredClasses;
     }
--- a/Routing/Loader/Reader/RestActionReader.php
+++ b/Routing/Loader/Reader/RestActionReader.php
@@ -67,6 +67,31 @@ class RestActionReader
     private $availableConventionalActions = ['new', 'edit', 'remove'];
     private $hasMethodPrefix;
 
+    /**
+     * ignore several type hinted arguments.
+     *
+     * @var array
+     */
+    private $ignoredClasses = [
+        ConstraintViolationListInterface::class,
+        MessageInterface::class,
+        ParamConverter::class,
+        ParamFetcherInterface::class,
+        Request::class,
+        SessionInterface::class,
+        UserInterface::class,
+    ];
+
+    /**
+     * Initializes controller reader.
+     *
+     * @param Reader               $annotationReader
+     * @param ParamReaderInterface $paramReader
+     * @param InflectorInterface   $inflector
+     * @param bool                 $includeFormat
+     * @param array                $formats
+     * @param bool                 $hasMethodPrefix
+     */
     public function __construct(Reader $annotationReader, ParamReaderInterface $paramReader, InflectorInterface $inflector, bool $includeFormat, array $formats = [], bool $hasMethodPrefix = true)
     {
         $this->annotationReader = $annotationReader;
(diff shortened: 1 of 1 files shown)


Routing/Loader/Reader/RestActionReader.php
```diff
@@ -236,6 +251,26 @@ public function getParents()
         return $this->parents;
     }
 
+    /**
+     * Set ignored classes.
```
Tobion: some detailed information about what this means is required

Routing/Loader/Reader/RestActionReader.php
```diff
@@ -236,6 +251,26 @@ public function getParents()
...
+    {
+        $this->ignoredClasses = $ignoredClasses;
+    }
+
+    /**
+     * Get ignored classes.
+     *
+     * @return array
+     */
+    public function getIgnoredClasses()
```
Tobion: please add return type

Routing/Loader/Reader/RestActionReader.php
```diff
@@ -116,6 +116,21 @@ class RestActionReader
      */
     private $hasMethodPrefix;
 
+    /**
+     * ignore several type hinted arguments.
+     *
+     * @var array
```
Tobion: please remove

Routing/Loader/Reader/RestActionReader.php
```diff
@@ -236,6 +251,29 @@ public function getParents()
         return $this->parents;
     }
 
+    /**
+     * Set ignored classes.
+     *
+     * These classes will be ignored for route construction when
+     * type hinted as method argument.
+     *
+     * @param array $ignoredClasses
```
Tobion: `@param string[] $ignoredClasses`

Routing/Loader/Reader/RestActionReader.php
```diff
@@ -236,6 +251,29 @@ public function getParents()
...
+     */
+    public function setIgnoredClasses(array $ignoredClasses): void
+    {
+        $this->ignoredClasses = $ignoredClasses;
+    }
+
+    /**
+     * Get ignored classes.
+     *
+     * @return array
```
Tobion: `@return string[]`

Routing/Loader/Reader/RestActionReader.php
```diff
@@ -236,6 +251,29 @@ public function getParents()
...
+     */
+    public function getIgnoredClasses(): array
```
Tobion: I don't think we need a getter for this. What's the use-case?

wickedOne: you might want to use it to exclude a specific class so when the default ignored classes are changed you don't have to update your code.

```php
$reader->setIgnoredClasses(arary_diff($reader->getIgnoredClasses(), [UserInterface::class]));
```

xabbuh: What about adding `addIgnoredClass()`/`addIgnoredClasses()` instead then? That would also make the intent more clear IMO when reading the code that calls any of these methods.

wickedOne: for this specific use case i actually need to remove an ignored class, but i agree it makes the intent a bit clearer.
do you want me to add a ``addIgnoredClass`` and ``removeIgnoredClass`` method instead of the getter and setter?

wickedOne: ping @xabbuh @Tobion do you want me to change the getter / setter to add / remove?

Tobion:
@wickedOne could you change the target to 2.x and rebase? Currently there is a conflict.

wickedOne:
rebased and changed the target.

Routing/Loader/Reader/RestActionReader.php
```diff
@@ -67,6 +67,29 @@ class RestActionReader
...
+        MessageInterface::class,
+        ParamConverter::class,
+        ParamFetcherInterface::class,
+        Request::class,
+        SessionInterface::class,
+        UserInterface::class,
+    ];
+
+    /**
+     * Initializes controller reader.
```
Tobion, line 84: all of this is irrelevant. maybe rebase problem.

wickedOne: yes, rebasee problem, sorry for that.
resolved...

PR #2093, opened 2020-02-19
Fix throwable error on exception controller

GuigZ-:
I've this problem on my projet

```Uncaught Error: Argument 1 passed to FOS\\RestBundle\\Controller\\ExceptionController::getStatusCode() must be an instance of Exception, instance of Error given, called in /var/www/vendor/friendsofsymfony/rest-bundle/Controller/ExceptionController.php on line 68```

To solve this i just change two files.

### DIFF
--- a/Controller/ExceptionController.php
+++ b/Controller/ExceptionController.php
@@ -65,7 +65,7 @@ class ExceptionController
     public function showAction(Request $request, $exception, DebugLoggerInterface $logger = null)
     {
         $currentContent = $this->getAndCleanOutputBuffering($request->headers->get('X-Php-Ob-Level', -1));
-        
+
         if ($exception instanceof \Exception) {
             $code = $this->getStatusCode($exception);
         } else {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by iisisrael** [comment]
Since the end result is that only a few lines of code have actually changed, it's probably a good idea to squash these commits.

**Comment by iisisrael** [comment]
Also, I'm looking for a way to add a functional test.  There hasn't been anything around this, so no current test validation for the change, except that manually applying these changes in an application that depends on this bundle fixes it.

**Comment by B-Galati** [comment]
Branch target should be 2.x no ?

File: Controller/ExceptionController.php [hunk 1]
```diff
@@ -89,11 +89,11 @@ protected function createView(\Exception $exception, $code, Request $request, $s
     /**
      * Determines the status code to use for the response.
      *
-     * @param \Exception $exception
+     * @param \Throwable $exception
      *
      * @return int
      */
-    protected function getStatusCode(\Exception $exception)
+    protected function getStatusCode(\Throwable $exception)
```
xabbuh: I am afraid that this is a BC break for child classes that we cannot make: https://3v4l.org/cLLKB

A possible solution could be to add a `private` `getStatusCodeFromThrowable()` method that will be called when the `$exception` argument in `showAction()` does not extend the `\Exception` class.

XWB:
@GuigZ- This should be in the 2.x branch. Can you do that?

GuigZ-:
@XWB it's done :)

XWB:
Thank you :)

PR #2097, opened 2020-02-28
fix SensioFrameworkExtraBundle compatibility

xabbuh:
fixes sensiolabs/SensioFrameworkExtraBundle#649

### DIFF
--- a/EventListener/ViewResponseListener.php
+++ b/EventListener/ViewResponseListener.php
@@ -96,7 +96,14 @@ class ViewResponseListener implements EventSubscriberInterface
                 $context->disableMaxDepth();
             }
 
-            list($controller, $action) = $configuration->getOwner();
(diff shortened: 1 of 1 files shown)


EventListener/ViewResponseListener.php
```diff
@@ -96,7 +96,13 @@ public function onKernelView($event)
...
             }
 
-            list($controller, $action) = $configuration->getOwner();
+            $owner = $configuration->getOwner();
+
+            if ([] === $owner) {
+                $owner = null;
+            }
+
+            [$controller, $action] = $owner;
```
Tobion: Maybe more understandable
```suggestion
            if ([] === $owner || null === $owner) {
                $controller = $action = null;
            } else {
                [$controller, $action] = $owner;
            }
```

PR #2102, opened 2020-03-02
bump PHP requirement to 7.2

xabbuh:


xabbuh:
This will allow us to add type declarations for method arguments without breaking BC (see https://3v4l.org/Vt3bJ).

Issue #2095, opened 2020-02-26
Not working Exception Handling in ExceptionController with PHP 7


MissbachMichael:
Please fix the bug:
php.CRITICAL: Uncaught Error: Argument 3 passed to FOS\RestBundle\Controller\ExceptionController::getTemplateData() must be an instance of Exception, instance of TypeError given, called in /var/www/html/vendor/friendsofsymfony/rest-bundle/Controller/ExceptionController.php on line 74
It´s already reported in https://github.com/FriendsOfSymfony/FOSRestBundle/issues/2092 after it was closed.

faresk93:
I'm getting the same error.
Previously this error was being launched in the getStatusCode() method. Now it's the getTemplateData().

**Line 69**: `vendor/friendsofsymfony/rest-bundle/Controller/ExceptionController.php`

`    if ($exception instanceof \Exception) {
            $code = $this->getStatusCode($exception);
        } else {
            $code = $this->getStatusCodeFromThrowable($exception);
        }`

The latest release **2.7.1** has made some improvements to check the exception instance in the showAction() method but apparently it has forgot to handle the getTemplateData() method requiring an \Exception Type as a 3rd parameter 
`(public function getTemplateData($currentContent, $code, \Exception $exception, DebugLoggerInterface $logger = null)`

For the moment, and as temporary workaround for this issue, I'm using a custom exception controller similar to the original fos_rest controller while avoiding this conflict !

MissbachMichael:
And please don´t forget to fix the method call on function "createView".

GuigZ-:
Pr #2096 is now ok, i wait feedback from fos team :)

xabbuh:
@MissbachMichael @faresk93 Can you both confirm that the current `2.7` branch after #2096 being merged fixes your issue? If yes, I will tag a new bugfix release later today.

faresk93:
I believe this hasn't solved the problem yet !
I got another exception related to **jms-serializer**

![image](https://user-images.githubusercontent.com/12989345/75664748-44979280-5c73-11ea-85ff-11b50e99ea0f.png)

xabbuh:
@faresk93 Thank you for the feedback. Please ping me if I can help you or when you need some feedback.

faresk93:
I created I pull request fixing this issue @xabbuh 
#2104

Issue #2092, opened 2020-02-12
Still getting "getStatusCode" Exception with Symfony 4.4 & FOSRestBundle 2.7


vincentjdc:
### Problem

In some cases (when I have a PHP syntax error), Symfony shows a 500 Exception with the message:

> Argument 1 passed to FOS\RestBundle\Controller\ExceptionController::getStatusCode() must be an instance of Exception, instance of ParseError given, called in /var/www/vendor/friendsofsymfony/rest-bundle/Controller/ExceptionController.php on line 68

The problem is present with Symfony 4.4 (freshly installed with `symfony new app --version=4.4) and FOSRestBundle 2.7

### To reproduce

The following example is pretty simple : 1 table `user` with 1 field `firstName` and a controller to list the users.

1. Pull the repository https://github.com/vincentjdc/fosrest-status-code-exception-example
2. Configure a database + update the .env file to set up the database
3. run composer install to install vendors
4. run bin/console `doctrine:migrations:migrate` to create tables
5. run bin/console `doctrine:fixtures:load` to insert a user in the user table
6. Go to the `/api` route, you should see a list with 1 user "Vincent"
7. Insert an error in `src/Entity/User.php` (by removing a semicolon for example)
8. Refresh the `/api` page, you shoud see the exception.

### Expected

An error message with the real problem (semicolon)

B-Galati:
I have the same problem FYI

GuigZ-:
I open pr to solve this : #2093

iisisrael:
@GuigZ- I can't get a functional test to handle a ParseError.  I tried both in the test controller and in a kernel response subscriber, trying to get the parse error to happen later in the process.  Patch is [here](https://gist.github.com/iisisrael/9764e8845ee4448c6580784daa3dd365).

GuigZ-:
> @GuigZ- I can't get a functional test to handle a ParseError. I tried both in the test controller and in a kernel response subscriber, trying to get the parse error to happen later in the process. Patch is [here](https://gist.github.com/iisisrael/9764e8845ee4448c6580784daa3dd365).

With which php version?

iisisrael:
> With which php version?

PHP 7.2.27-6+ubuntu18.04.1+deb.sury.org+1

> Edit: the case is not that, the error is triggered in controller not in subscriber

[...]

PR #2104, opened 2020-03-02
Fixed exception handling bug in showAction (ExceptionController)

faresk93:
Refactored showAction in ExceptionController to fix exception handling and a recent thrown exception bug:
`Undefined property : App\Kernel::$freshCache` exception related to jms-serializer

The current exception is being launched on handleView in case of none \Exception instance of an exception.
This  has been **fixed** by transforming all none \Exception instances (such as TypeError) to \Exception.

### DIFF
--- a/Controller/ExceptionController.php
+++ b/Controller/ExceptionController.php
@@ -70,15 +70,11 @@ class ExceptionController
             $code = $this->getStatusCode($exception);
         } else {
             $code = $this->getStatusCodeFromThrowable($exception);
+            $exception = new \Exception($exception->getMessage(), $code);
         }
         $templateData = $this->getTemplateData($currentContent, $code, $exception, $logger);
-
-        if ($exception instanceof \Exception) {
-            $view = $this->createView($exception, $code, $templateData, $request, $this->showException);
-        } else {
-            $view = $this->createViewFromThrowable($exception, $code, $templateData, $request, $this->showException);
-        }
-
+        $view = $this->createView($exception, $code, $templateData, $request, $this->showException);
+        
         $response = $this->viewHandler->handle($view);
 
         return $response;
--- a/Controller/ExceptionController.php
+++ b/Controller/ExceptionController.php
@@ -70,7 +70,7 @@ class ExceptionController
             $code = $this->getStatusCode($exception);
         } else {
             $code = $this->getStatusCodeFromThrowable($exception);
-            $exception = new \Exception($exception->getMessage(), $code);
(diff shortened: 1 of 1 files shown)


faresk93:
Pull request related to issue #2095

Controller/ExceptionController.php
```diff
@@ -70,15 +70,11 @@ public function showAction(Request $request, $exception, DebugLoggerInterface $l
             $code = $this->getStatusCode($exception);
         } else {
             $code = $this->getStatusCodeFromThrowable($exception);
+            $exception = new \Exception($exception->getMessage(), $code);
```
xabbuh: I am not convinced that this will work as expected since we will lose the full exception stack trace here.

faresk93: I'm currently still working on another fix. But testing this solution everything seems to work just fine.
The View class doesn't see to accept a TypeError Exception as a first argument. Converting the instance to a \Exception seemed was a first workaround !

xabbuh: Or to start with something else: Do you see the same error when using the Symfony serializer? If not, could we hook into the JMS serializer and fix how the exception is handled?

faresk93: Well after tracing the the problem down to its roots, I found out that the problem lies in the serialization handled by jms_serializer when handling the view object.
jms_serializer, somehow, can't manage to serialize the \TypeError Exception passed to the View Class and throws the previously mentioned exception !
I'll be looking deeply into this problem later. I'll explore another way to solve this !

GuigZ-: I think to this fix but i'm not convinced too ... for the meme reason

xabbuh: Maybe we can catch these cases in the `ExceptionHandler` class. I can try to look into it later today.

@faresk93 Do you have an application that allows to reproduce it or is the one from #2092 good enough?

faresk93: A test case should be to return a different type than the on expected from a method.

My case was an Entity Class getter returning a PersistentCollection with a return Type (php 7.3) being set to ArrayCollection (instead of just Collection) which throws a \TypeError exception !
(I'm not quite sure about the issue you mentioned)

xabbuh: Can you check if #2105 fixes this?

xabbuh:
Thank you for working on this @faresk93. This should be fixed by #2105.

faresk93:
Thank you as well

PR #2105, opened 2020-03-03
handle all Throwable instances during serialization

### LINKED ISSUE #2095: Not working Exception Handling in ExceptionController with PHP 7

xabbuh:
fixes #2095, replaces #2104

### DIFF
--- a/Serializer/Normalizer/AbstractExceptionNormalizer.php
+++ b/Serializer/Normalizer/AbstractExceptionNormalizer.php
@@ -44,17 +44,17 @@ class AbstractExceptionNormalizer
     /**
      * Extracts the exception message.
      *
-     * @param \Exception $exception
+     * @param \Throwable $throwable
      * @param int|null   $statusCode
      *
      * @return string
      */
-    protected function getExceptionMessage(\Exception $exception, $statusCode = null)
+    protected function getMessageFromThrowable(\Throwable $throwable, $statusCode = null)
     {
-        $showMessage = $this->messagesMap->resolveException($exception);
+        $showMessage = $this->messagesMap->resolveThrowable($throwable);
 
         if ($showMessage || $this->debug) {
-            return $exception->getMessage();
+            return $throwable->getMessage();
         }
 
         return array_key_exists($statusCode, Response::$statusTexts) ? Response::$statusTexts[$statusCode] : 'error';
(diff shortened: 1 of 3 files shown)


Serializer/Normalizer/AbstractExceptionNormalizer.php
```diff
@@ -44,14 +44,14 @@ public function __construct(ExceptionValueMap $messagesMap, $debug)
...
      * Extracts the exception message.
      *
-     * @param \Exception $exception
+     * @param \Throwable $exception
      * @param int|null   $statusCode
      *
      * @return string
      */
-    protected function getExceptionMessage(\Exception $exception, $statusCode = null)
+    protected function getExceptionMessage(\Throwable $exception, $statusCode = null)
```
GuigZ-: You change a definition of protected method, if someone overload this method, the code will be uncompatible.

xabbuh: Thank you for the heads up. In fact, here this change is safe as the `AbstractExceptionNormalizer` class is marked as `@internal` which means that we do not need to aim for BC here.

Serializer/Normalizer/AbstractExceptionNormalizer.php
```diff
@@ -44,17 +44,17 @@ public function __construct(ExceptionValueMap $messagesMap, $debug)
...
+     * @param \Throwable $throwable
      * @param int|null   $statusCode
      *
      * @return string
      */
-    protected function getExceptionMessage(\Exception $exception, $statusCode = null)
+    protected function getMessageFromThrowable(\Throwable $throwable, $statusCode = null)
     {
-        $showMessage = $this->messagesMap->resolveException($exception);
+        $showMessage = $this->messagesMap->resolveThrowable($throwable);
```
yceruto: I can't find the `resolveThrowable()` method here:
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/master/Util/ExceptionValueMap.php

How it should work?

xabbuh: It's only present in 2.x: https://github.com/FriendsOfSymfony/FOSRestBundle/blob/2.7/Util/ExceptionValueMap.php

The code for 3.0 was fixed in #2112.

PR #2108, opened 2020-03-03
decorate the exception_listener

lsmith77:
this seems like a cleaner and also more performant approach.
there is less risk of priorities causing issues etc.

I mostly came to this as part of looking into https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1692 and realizing this is another place where we should decorate. and I needed a simpler task to get into things again :)

### DIFF
--- a/DependencyInjection/Compiler/ErrorListenerPass.php
+++ b/DependencyInjection/Compiler/ErrorListenerPass.php
@@ -1,0 +1,38 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\RestBundle\DependencyInjection\Compiler;
+
+use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
+use Symfony\Component\DependencyInjection\ContainerBuilder;
+
+/**
+ * Handle the error error related listener.
+ *
+ * @author Lukas Kahwe Smith <smith@pooteeweet.org>
+ *
+ * @internal
+ */
+final class ErrorListenerPass implements CompilerPassInterface
+{
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->hasDefinition('fos_rest.exception_listener')) {
+            return;
+        }
+
+        if ($container->has('twig.exception_listener')) {
+            $container->getDefinition('fos_rest.exception_listener')->setDecoratedService('twig.exception_listener');
+        } elseif ($container->has('exception_listener')) {
+            $container->getDefinition('fos_rest.exception_listener')->setDecoratedService('exception_listener');
+        }
+    }
+}
--- a/EventListener/ExceptionListener.php
+++ b/EventListener/ExceptionListener.php
@@ -19,7 +19,7 @@ use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 use Symfony\Component\EventDispatcher\EventSubscriberInterface;
 use Symfony\Component\HttpKernel\Event\ControllerArgumentsEvent;
 use Symfony\Component\HttpKernel\Event\ExceptionEvent;
-use Symfony\Component\HttpKernel\EventListener\ExceptionListener as LegacyExceptionListener;
+use Symfony\Component\HttpKernel\Event\ResponseEvent;
 use Symfony\Component\HttpKernel\EventListener\ErrorListener;
 use Symfony\Component\HttpKernel\KernelEvents;
 
@@ -34,15 +34,20 @@ class ExceptionListener implements EventSubscriberInterface
 {
     private $exceptionListener;
     private $dispatcher;
+    private $innerExceptionListener;
 
-    public function __construct($controller, ?LoggerInterface $logger, EventDispatcherInterface $dispatcher)
+    public function __construct($controller, ?LoggerInterface $logger, $debug = false, EventDispatcherInterface $dispatcher, $innerExceptionListener = null)
     {
-        if (class_exists(ErrorListener::class)) {
-            $this->exceptionListener = new ErrorListener($controller, $logger);
-        } else {
-            $this->exceptionListener = new LegacyExceptionListener($controller, $logger);
-        }
+        $this->exceptionListener = new ErrorListener($controller, $logger, $debug);
         $this->dispatcher = $dispatcher;
+        $this->innerExceptionListener = $innerExceptionListener;
+    }
+
+    public function logKernelException(ExceptionEvent $event)
+    {
+        if ($this->innerExceptionListener) {
+            $this->innerExceptionListener->logKernelException($event);
+        }
     }
 
     public function onKernelException(ExceptionEvent $event): void
@@ -50,7 +55,11 @@ class ExceptionListener implements EventSubscriberInterface
         $request = $event->getRequest();
 
         if (!$request->attributes->get(FOSRestBundle::ZONE_ATTRIBUTE, true)) {
-            return;
+            if (null === $this->innerExceptionListener) {
+                return;
+            }
+
+            return $this->innerExceptionListener->onKernelException($event);
         }
 
         $exception = $event->getThrowable();
@@ -58,6 +67,7 @@ class ExceptionListener implements EventSubscriberInterface
         $controllerArgsListener = function ($event) use (&$controllerArgsListener, $exception) {
             /** @var ControllerArgumentsEvent $event */
             $arguments = $event->getArguments();
+
             foreach ($arguments as $k => $argument) {
                 if ($argument instanceof FlattenException || $argument instanceof LegacyFlattenException) {
                     $arguments[$k] = $exception;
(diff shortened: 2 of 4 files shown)


EventListener/ExceptionListener.php
```diff
@@ -34,15 +35,25 @@ class ExceptionListener implements EventSubscriberInterface
...
     private $exceptionListener;
     private $dispatcher;
+    private $innerExceptionListener;
 
-    public function __construct($controller, ?LoggerInterface $logger, EventDispatcherInterface $dispatcher)
+    public function __construct($controller, ?LoggerInterface $logger, $debug = false, EventDispatcherInterface $dispatcher, $innerExceptionListener = null)
     {
         if (class_exists(ErrorListener::class)) {
-            $this->exceptionListener = new ErrorListener($controller, $logger);
+            $this->exceptionListener = new ErrorListener($controller, $logger, $debug);
```
xabbuh: I am not sure that we really need the `$exceptionListener` property anymore now. If I don't miss anything we would use the decorated exception listener instead.

DependencyInjection/Compiler/ErrorListenerPass.php
```diff
@@ -0,0 +1,38 @@
...
+    public function process(ContainerBuilder $container)
+    {
+        if (!$container->hasDefinition('fos_rest.exception_listener')) {
+            return;
+        }
+
+        if ($container->has('twig.exception_listener')) {
+            $container->getDefinition('fos_rest.exception_listener')->setDecoratedService('twig.exception_listener');
+        } elseif ($container->has('exception_listener')) {
+            $container->getDefinition('fos_rest.exception_listener')->setDecoratedService('exception_listener');
```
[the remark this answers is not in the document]
lsmith77: right now our exception listener will work just fine even if there is no core exception listener for what ever reason.

EventListener/ExceptionListener.php
lsmith77: it is still needed. see the last line of `\FOS\RestBundle\EventListener\ExceptionListener::onKernelException()`, we still let the core listener handle the actual response creation.

DependencyInjection/Compiler/ErrorListenerPass.php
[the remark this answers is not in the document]
lsmith77: also our listener is only enabled if the user explicitly turned it on.

EventListener/ExceptionListener.php
xabbuh: Don't we have to delegate that to the decorated exception listener now? Something like this:

```php
public function __construct($controller, ?LoggerInterface $logger, EventDispatcherInterface $dispatcher, $decoratedExceptionListener = null)
{
    if (null !== $decoratedExceptionListener) {
        $this->exceptionListener = $decoratedExceptionListener;
    } elseif (class_exists(ErrorListener::class)) {
        $this->exceptionListener = new ErrorListener($controller, $logger);
    } else {
        $this->exceptionListener = new LegacyExceptionListener($controller, $logger);
    }

    $this->dispatcher = $dispatcher;
}
```

EventListener/ExceptionListener.php
```diff
@@ -34,15 +35,25 @@ class ExceptionListener implements EventSubscriberInterface
 {
     private $exceptionListener;
     private $dispatcher;
+    private $innerExceptionListener;
 
-    public function __construct($controller, ?LoggerInterface $logger, EventDispatcherInterface $dispatcher)
+    public function __construct($controller, ?LoggerInterface $logger, $debug = false, EventDispatcherInterface $dispatcher, $innerExceptionListener = null)
```
yceruto: I wonder if with this approach we no longer need the `LegacyExceptionListener`, we could use the `ErrorListener` always, this way the exception attribute would be a \Throwable instance (by default) and we can get rid of the controller argument replacement logic as well.

EventListener/ExceptionListener.php
lsmith77: basically the logic I implemented checks the zone config and if its within the zone .. it uses the `exceptionListener` and if its outside the zone it uses `innerExceptionListener`

EventListener/ExceptionListener.php
```diff
@@ -69,20 +85,38 @@ public function onKernelException($event)
...
+
     /**
      * {@inheritdoc}
      */
     public static function getSubscribedEvents()
     {
-        return array(
-            KernelEvents::EXCEPTION => array('onKernelException', -100),
-        );
+        if (class_exists(ErrorListener::class)) {
```
yceruto: related to my comment above, `master` branch supports 4.4+ only, this shouldn't be true always?

EventListener/ExceptionListener.php
lsmith77: good point! will adjust this

EventListener/ExceptionListener.php
lsmith77: which means I rather need to check for the LegacyExceptionListener

EventListener/ExceptionListener.php
yceruto: here is an example:  https://github.com/yceruto/FOSRestBundle/commit/3ae17f90bdbc01ef10f891a4a82e50ffaf3cc26a

EventListener/ExceptionListener.php
```diff
@@ -69,20 +80,38 @@ public function onKernelException($event)
...
+
     /**
      * {@inheritdoc}
      */
     public static function getSubscribedEvents()
     {
-        return array(
-            KernelEvents::EXCEPTION => array('onKernelException', -100),
-        );
+        if (class_exists(LegacyExceptionListener::class)) {
```
yceruto: It's not necessary imo, we can use `ErrorListener` always.

lsmith77: The events are not the same but I guess the additional event we would just skip.

EventListener/ExceptionListener.php
```diff
@@ -53,14 +58,19 @@ public function onKernelException($event)
...
+            if (null === $this->innerExceptionListener) {
+                return;
+            }
+
+            return $this->innerExceptionListener->onKernelException($event);
         }
 
         $exception = $event->getThrowable();
 
         $controllerArgsListener = function ($event) use (&$controllerArgsListener, $exception) {
```
yceruto: this code can be removed safely if the exception argument of the ExceptionController is typed as `\Throwable`.

EventListener/ExceptionListener.php
```diff
@@ -34,15 +34,20 @@ class ExceptionListener implements EventSubscriberInterface
...
-            $this->exceptionListener = new LegacyExceptionListener($controller, $logger);
-        }
+        $this->exceptionListener = new ErrorListener($controller, $logger, $debug);
         $this->dispatcher = $dispatcher;
+        $this->innerExceptionListener = $innerExceptionListener;
+    }
+
+    public function logKernelException(ExceptionEvent $event)
+    {
+        if ($this->innerExceptionListener) {
```
yceruto: we should be able to log the exception even if the inner listener is null. 
I'd suggest call to `$this->exceptionListener->logKernelException()`. Same for `onControllerArguments()` method.

EventListener/ExceptionListener.php
```diff
@@ -69,20 +79,34 @@ public function onKernelException($event)
...
             $this->dispatcher->removeListener(KernelEvents::CONTROLLER_ARGUMENTS, $controllerArgsListener);
         };
-        $this->dispatcher->addListener(KernelEvents::CONTROLLER_ARGUMENTS, $controllerArgsListener, -100);
+
+        $this->dispatcher->addListener(KernelEvents::CONTROLLER_ARGUMENTS, $controllerArgsListener);
 
         $this->exceptionListener->onKernelException($event);
     }
 
+    public function removeCspHeader(ResponseEvent $event): void
```
yceruto: I failed to see where this method is called. Can you verify if it's still necessary?

lsmith77: this is forwards compatibility, see symfony/symfony master

EventListener/ExceptionListener.php
lsmith77: the reason I didn't do this is because I wanted to keep this listener scoped with its current focus. if you want logging, then imho you should keep the core listener enabled. but I am open to reconsider.

EventListener/ExceptionListener.php
lsmith77: hmm removing this I get failures in the test suite .. didn't have time to investigate if these are just problems in the test suite. also not sure if we can do some tweaks to the ExceptionController to make it work without this code .. can you make your suggesting more concrete?

yceruto: Yes, I send you a concrete example in this comment https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2108#discussion_r387722224 :)

EventListener/ExceptionListener.php
yceruto: Oh! I see :) nvm.

EventListener/ExceptionListener.php
lsmith77: ah sorry .. can you send this as a PR into this PR?

yceruto: see https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2128

lsmith77:
@xabbuh / @yceruto with @yceruto's improvement merged into this PR, imho this PR is now quite solid.

xabbuh:
Sorry that it took me so long to look into this. The changes look great to me. I just wonder if we couldn't already implement them in 2.8. I tried to backport them in #2146. What do you think?

Tobion:
Closing in favor of #2161

PR #2109, opened 2020-03-04
remove previous exception, add tests for access denied listener

lsmith77:
This PR exprovides an alternative fix for https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1692

I tried to decorate the security component exception listener but it does not seem possible since the Firewall context has a typehint on the class (not an interface) on the exception listener (https://github.com/symfony/symfony/blob/master/src/Symfony/Bundle/SecurityBundle/Security/FirewallContext.php#L30) but the class is also marked as `@final` (https://github.com/symfony/symfony/blob/master/src/Symfony/Component/Security/Http/Firewall/ExceptionListener.php#L43). As such the fix is essentially a hack by not pass the previous exception to prevent the core listener from doing anything.

open questions
* [ ] instead of removing the previous exception we could also try to call `\Symfony\Component\Security\Http\Firewall\ExceptionListener::unregister()` on the given firewall instead but not sure how easy it is to find the relevant firewall
* [ ] yet another approach could be to somehow try and handle this in our ExceptionListener if https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2108 is merged. ie. use the AccessDeniedListener to set something on the request, so that we do not use the exception that was set in the core security exception listener but the one from the rest bundle AccessDeniedListener

### DIFF
--- a/EventListener/AccessDeniedListener.php
+++ b/EventListener/AccessDeniedListener.php
@@ -74,12 +74,12 @@ class AccessDeniedListener implements EventSubscriberInterface
         }
 
         if ($exception instanceof AccessDeniedException) {
-            $exception = new AccessDeniedHttpException('You do not have the necessary permissions', $exception);
+            $exception = new AccessDeniedHttpException('You do not have the necessary permissions');
         } elseif ($exception instanceof AuthenticationException) {
             if ($this->challenge) {
-                $exception = new UnauthorizedHttpException($this->challenge, 'You are not authenticated', $exception);
+                $exception = new UnauthorizedHttpException($this->challenge, 'You are not authenticated');
             } else {
-                $exception = new HttpException(401, 'You are not authenticated', $exception);
+                $exception = new HttpException(401, 'You are not authenticated');
             }
         }
 


lsmith77:
FYI .. I plan to have a look with @Tobion this week if maybe this entire listener could be dropped by documentation (and a small improvement to core).

lsmith77:
@xabbuh @yceruto @Tobion so I have added a bunch of tests to see how core behaves these days and it seems like as long as users rely on `http_basic` it works fine.

`json_login` and `form_login` both do not behave as one would expect in an API context. We can of course advise our users to not use `form_login` but core behaviour of `json_login` seems to just be JSON for `form_login` with the same redirect behaviour. But to fully analyse this we would need to implement a proper login endpoint for the `json_login`.

still I would say for 3.0 let us do the following:
1) replace this PR with a PR that removes the access denied listener
2) consider adding to core a `json_api_login` that returns a 403 rather than a 302 redirect

wdyt?

Tobion:
According to https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2161#issuecomment-618606871 this PR fixes the redirect. So I'd be ok to merge this. It just needs a rebase.

xabbuh:
Looks good to me to. Is there anything prevent us from merging this into `2.8`?

xabbuh:
I was stepping through the code to understand what is going on and I now think I understand why the patch fixes the problem. However, unfortunately the tests still pass when we revert the changes made to the `AccessDeniedListener`. I am going to look into how we need to modify them before merging to ensure that we cannot blindly introduce a regression in the future here.

xabbuh:
Thank you @lsmith77.

PR #2106, opened 2020-03-03
make classes final

xabbuh:

PR #2122, opened 2020-03-05
do not flag the serializer adapter services as lazy

xabbuh:
see https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2106#issuecomment-595304595


xabbuh:
@GuilhemN Do you remember why we flagged these services as lazy in #1252?

alterphp:
According to https://symfony.com/doc/current/service_container/lazy_services.html#why-lazy-services, I guess they are marked as **lazy** because only of them is really used at last. The alias `fos_rest.serializer` points to the selected serializer. No need to instanciate them all.

PR #2131, opened 2020-03-16
fix: Throwable configuration

GuigZ-:


DependencyInjection/Configuration.php
```diff
@@ -535,10 +535,10 @@ private function addExceptionSection(ArrayNodeDefinition $rootNode)
             ->end();
     }
 
-    private function testExceptionExists(string $exception)
+    private function testExceptionExists(string $throwable)
     {
-        if (!is_subclass_of($exception, \Exception::class) && !is_a($exception, \Exception::class, true)) {
-            throw new InvalidConfigurationException("FOSRestBundle exception mapper: Could not load class '$exception' or the class does not extend from '\\Exception'. Most probably this is a configuration problem.");
+        if (!is_subclass_of($throwable, \Throwable::class) && !is_a($throwable, \Throwable::class, true)) {
```
xabbuh: Looks like we can shorten this now to:

```php
if (!is_subclass_of($throwable, \Throwable::class)) {
    throw new InvalidConfigurationException("FOSRestBundle exception mapper: Could not load class '$throwable' or the class does not extend from '\\Throwable'. Most probably this is a configuration problem.");
}
```

xabbuh:
Thank you @GuigZ-.

PR #2146, opened 2020-03-28
decorate the exception_listener

xabbuh:


EventListener/ExceptionListener.php
```diff
@@ -77,16 +90,32 @@ public function onKernelException($event)
...
      * {@inheritdoc}
      */
     public static function getSubscribedEvents()
     {
-        return array(
-            KernelEvents::EXCEPTION => array('onKernelException', -100),
-        );
+        if (class_exists(ErrorListener::class)) {
+            return ErrorListener::getSubscribedEvents();
+        } else {
```
lsmith77: technically this else isn’t needed.

lsmith77:
not sure if its worth the risk .. I mean this will have side effects .. note I think the behavior will be more predictable but people might have worked around those issues and things might now suddenly behave differently.

xabbuh:
That's a valid argument. Do you think it's worth it to allow to already opt-in to this in 2.8? So that people can actually prepare for the change in 3.0?

lsmith77:
could we reasonably make this optional?

xabbuh:
I have pushed another commit. Do you have something like that in mind?

lsmith77:
yeah .. and in 3.x we default to `forward` (and remove the forward option).

xabbuh:
Yes, that what I have in mind. I will look into giving the tests and trigger a deprecation when the option is not enabled.

DependencyInjection/Configuration.php
```diff
@@ -479,6 +479,21 @@ private function addExceptionSection(ArrayNodeDefinition $rootNode)
                                 })
                             ->end()
                         ->end()
+                        ->booleanNode('forward')
```
Tobion: please add an info node with a description what this does

EventListener/ExceptionListener.php
```diff
@@ -31,17 +32,26 @@
  */
 class ExceptionListener implements EventSubscriberInterface
 {
-    private $exceptionListener;
+    private $errorListener;
```
yceruto: missing initialization?

DependencyInjection/Configuration.php
xabbuh: 👍 done

EventListener/ExceptionListener.php
```diff
@@ -33,15 +34,24 @@ class ExceptionListener implements EventSubscriberInterface
 {
     private $exceptionListener;
     private $dispatcher;
+    private $innerExceptionListener;
 
-    public function __construct($exceptionListener, EventDispatcherInterface $dispatcher)
+    public function __construct($exceptionListener, EventDispatcherInterface $dispatcher, $innerExceptionListener = null)
```
Tobion: Do we still need this `ExceptionListener` class in 3.0? I don't see what this custom one brings? I thought we will just rely on the core ErrorListener and don't need to decorate at all? But then the "forward" option wouldn't mean anything anymore.

xabbuh: The `ErrorListener` will use a different controller than what the core listener uses. So yes, I think we still need to make the distinction.

Tobion: I don't understand. I think we can just let people change the `framework.error_controller` config (or automate this with a recipes etc.) to the rest-bundle ExceptionController.

We can even think about not needing the rest-bundle ExceptionController at all. It seems to provide
1. mapping exceptions to status codes (this could IMO be moved to core)
2. Using the viewhandler for rendering. Does not seem that useful anymore since sf 4.4 provides better error rendering and support content-negotiation.

xabbuh: I have no detailed knowledge about the difference here. Though following your thought would lead to the consequence that we could/should drop the exception listener as well as the controller and its option in 3.0?

xabbuh: @yceruto @lsmith77 Do you have any thoughts here?

yceruto: I agree that this feature "mapping exceptions to status codes (this could IMO be moved to core)" is the only blocker to remove all these custom error mechanism. 

I need to look at this a bit more... perhaps we can hook this feature into the new error rendering system, but anyway that would be for 3.x (with support for 4.4+)

xabbuh: > but anyway that would be for 3.x (with support for 4.4+)

Sure, but a solution for 3.x would influence our deprecation layer in 2.8.

xabbuh: We could also decide not to deprecate our own listener for now and keep that for a later 4.x release whenever that will be.

Tobion: I've looked at it in more detail and I'm pretty sure we should not have this custom exception listener in 3.0.
So instead of enhancing the listener in 2.x to use decoration, I'd instead mark it as deprecated.
In 3.0 we can rely on the core `Symfony\Component\HttpKernel\EventListener\ErrorListener`. To use a custom controller, we can let people configure `framework.error_controller` or we decorate the default `\Symfony\Component\HttpKernel\Controller\ErrorController` to add the logic from `private function getStatusCodeFromThrowable(\Throwable $exception)`.

Tobion: This way we can also deprecate `fos_rest.exception.exception_controller` and `fos_rest.exception.service`

yceruto: If this is about changing the status code only, we could use an exception/response listeners to support this specific feature:

 * An exception listener to grap the original exception class only,
 * A response listener to set the status code if available (`FOS\RestBundle\Controller\ExceptionController::getStatusCodeFromThrowable`)

This way it's less intrusive imho, and all this can be achieved in one class. Deprecating everything else.

xabbuh: Do you think something like #2161 would solve it?

xabbuh:
Do you all agree that #2161 rather looks like the solution and that we should finish that one instead of spending time here?

xabbuh:
I am closing here in favour of #2161 as I think that's the way we should solve this. We can still reopen here if there is strong disagreement though.

PR #2150, opened 2020-03-30
[add] serializerAttributes property for View annotation

gam6itko:
It will be very useful for developers to be able to add a couple of attributes for use it in Listeners and Handlers


gam6itko:
Please rerun travis job

PR #2151, opened 2020-03-30
decorate the legacy layer

xabbuh:


alexander-schranz:
Any update to this? Think this is the last open part for 3.0 or?

PR #2152, opened 2020-04-05
deprecate the automatic route generation feature

### LINKED ISSUE #1954: RFC: Deprecate route creation based on class/method names

xabbuh:
fixes #1954


alexander-schranz:
Any update to this? Think this is the last open part for 3.0 or?

Linking the issue here: #1954

PR #2157, opened 2020-04-17
add missing deprecations for automatic route generation

Tobion:
Add some missing deprecations from #2152 for #1954


xabbuh:
Looks like a test needs to be updated.

xabbuh:
Thank you Tobias.

Issue #2019, opened 2019-08-12
Problem Details for machine-readable details of errors in a HTTP response


maxbeckers:
There is a RFC 7807 for a response structure for error cases. For more details see https://tools.ietf.org/html/rfc7807.
For a form validation error it could be like this:
```
   {
   "type": "https://example.net/validation-error",
   "title": "Your request parameters didn't validate.",
   "invalid-params": [ {
                         "name": "age",
                         "reason": "must be a positive integer"
                       },
                       {
                         "name": "color",
                         "reason": "must be 'green', 'red' or 'blue'"}
                     ]
   }
```

GuilhemN:
It does look more suited than our current format indeed but it would force our users to put a lot of energy in the upgrade to the next major release of this bundle for no much gain imo.

We could offer an option to use this rfc for new projects but I wouldn't make it the default because of backward compatibility.

Tobion:
In Symfony 4.4 there is a new error rendering that will is compatible with this RFC. If we want to make the FosRestBundle compatible with SF 4.4 and prevents deprecations, we likely need to switch to that new error rendering as well. So this change will come sooner or later. The question is only how to make the best transition.

maxbeckers:
I think we can have a new congif for this bundle, which is the current error response handling by default but can be set to the rfc response structure. With this solution it could be possible to add custom response structures for use cases with other error response requirements.

xabbuh:
FOSRestBundle 2.8/3.0 will allow to enable RFC 7807 compatible error responses leveraging the Symfony ErrorHandler component with the `flatten_exception_format` option:

```yaml
fos_rest:
       exception:
           flatten_exception_format: 'rfc7807'
```

PR #2161, opened 2020-04-17
add listener mapping exceptions to status codes

xabbuh:


EventListener/ResponseStatusCodeListener.php
```diff
@@ -0,0 +1,74 @@
...
+        if (method_exists($event, 'getThrowable')) {
+            $throwable = $event->getThrowable();
+        } else {
+            $throwable = $event->getException();
+        }
+
+        $statusCode = $this->exceptionValueMap->resolveThrowable($throwable);
+
+        if (is_int($statusCode)) {
+            $request->attributes->set(FOSRestBundle::STATUS_CODE_ATTRIBUTE, $statusCode);
```
Tobion, line 61: I think we could just set a private property on the ResponseStatusCodeListener instead of polluting the request class.

yceruto: agree with Tabion

xabbuh: Not sure that's a good idea as it could easily break with sub requests.

Tobion: I think we should skip sub-request here. I don't see the value for subrequest. Also the  old logic with the ExceptionController was also only called for the master request.

Serializer/Normalizer/ExceptionHandler.php
```diff
@@ -11,14 +11,16 @@
...
+
 use JMS\Serializer\Context;
 use JMS\Serializer\GraphNavigatorInterface;
 use JMS\Serializer\Handler\SubscribingHandlerInterface;
 use JMS\Serializer\JsonSerializationVisitor;
 use JMS\Serializer\XmlSerializationVisitor;
 
 /**
- * @internal since 2.8
+ * @deprecated since 2.8
```
Tobion, line 23: The error serialization in json/xml from symfony requires the symfony/serializer. If people use the jms/serializer instead, how will exceptions be serialized to json/xml?

EventListener/ResponseStatusCodeListener.php
xabbuh: Fair enough, I am still not a big fan of making the service stateful though, but if turns out not always work as desired we can switch to the initial implementation without any issues. I have also added a test to ensure that the listener only acts on master requests.

Serializer/Normalizer/ExceptionHandler.php
xabbuh: For people using the Symfony Serializer component, there is a new `ProblemNormalizer` that deals with `FlattenException` instances and seamlessly hooks into the `SerializerErrorRenderer`. I am currently looking into if and how we can ship a similar solution for people who prefer JMS Serializer.

xabbuh: I think we can add a custom handler in this bundle, but the renderer would probably best live in the JmsSerializerBundle.

Tobion: I think it's a good thing to provide a normalizer that supports the old `{ code: ..., message: ...}` error response for BC. But maybe it's best to provide an option to opt-in to the new problem details format according to RFC 7807 that Symfony ErrorHandler uses as well, see also #2019

xabbuh: We could add an option to enable/disable the `ProblemNormalizer`. If it is not registered, the `SerializerErrorRenderer` would already render the standard format.

xabbuh: I have added another option. One needs to enable it to register the new `ProblemNormalizer`. Otherwise the RFC 7807 compatible normalizer provided by Symfony will do its job.

Tobion: Not sure about the naming of ProblemNormalizer. The name suggests to use the Problem Details RFC. But it's the opposite. Maybe `ThrowableNormalizer`?
Also I see correctly, people need to enable `serialize_flatten_exception` (i.e. ProblemNormalizer), to support the `fos_rest.exception.messages` config or?
Ideally, the exception message config should work either way (new style with problem details rfc or old style, with jms or symfony serializer).

xabbuh: I have renamed the class to `FlattenExceptionNormalizer`. I have also changed the new config option to `flatten_exception_format` which allows to choose the format in which `FlattenException` instances are serialized (`legacy` or `rfc7807`).

xabbuh:
Okay, this is ready to be reviewed. This PR now also contains a `FlattenExceptionHandler` which one could use in a JMS Serializer based project (admittedly wiring the error renderer with JMS Serializer is a bit tricky right now, but that's out of the scope of FOSRestBundle).

DependencyInjection/Configuration.php
```diff
@@ -473,22 +473,52 @@ private function addExceptionSection(ArrayNodeDefinition $rootNode)
...
+                        ->end()
                         ->scalarNode('exception_controller')
-                            ->defaultValue(static function () {
-                                @trigger_error('Not setting the "fos_rest.exception.exception_controller" configuration option is deprecated since FOSRestBundle 2.8. Its default value will be set to "fos_rest.exception.controller::showAction" in 3.0.', E_USER_DEPRECATED);
+                            ->defaultNull()
+                            ->setDeprecated('The "%path%.%node%" option is deprecated since FOSRestBundle 2.8.')
+                        ->end()
+                        ->booleanNode('serialize_exceptions')
+                            ->defaultValue(function () {
+                                @trigger_error('Enabling the "fos_rest.exception.serialize_exceptions" option is deprecated since FOSRestBundle 2.8.', E_USER_DEPRECATED);
```
alexander-schranz, line 501: Why deprecating this? We using exception code in the api to handle errors in the frontend application. We are required to have the response like: `{"code": "7007", "message": "Exception Message"}` in our rest api.

xabbuh: The new `FlattenExceptionHandler`/`FlattenExceptionNormalizer` should handle that (see the dedicated tests). You just need to make sure that exceptions are handled by the ErrorHandler component.

alexander-schranz: Ok that would be great!

broncha:
@xabbuh This still does not handle the redirection to the entry point by the firewall, does it?
I updated to 2.x-dev and json request still gets redirected to the login route! Am I missing something?

xabbuh:
@broncha To be honest I do not have any idea what we are talking about here. Do you refer to an existing issue?

broncha:
@xabbuh sorry. I came across this PR from another thread where discussions were going on about increasing the priority of the exception listener for handling redirection to entry-point. With the exception listener deprecated, how should one go about returning a 401 response for JSON rather than redirecting to the entry point? `access_denied_listener` does not seem to work either

xabbuh:
@broncha Off hand I do not have an answer for that. Can you please open an issue so that your request doesn't get lost?

Tobion:
@broncha isn't https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2109 what you are talking about?

broncha:
@Tobion Ah yes. Also #1692 was how I found this PR. So we are still not there yet. No problem.

Tobion:
@broncha can you test if #1692 fixes your issue?

broncha:
Both resulted in a RuntimeException

```
Resources are not supported in serialized data. Path: 
Monolog\Handler\StreamHandler -> 
Symfony\Bridge\Monolog\Logger -> 
Symfony\Component\HttpKernel\Debug\TraceableEventDispatcher -> 
Symfony\Component\HttpKernel\HttpKernel -> 
Symfony\Component\HttpKernel\Event\ExceptionEvent -> 
Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException
```

broncha:
I might have missed `serialize_exceptions: true` 
With that, it worked with #2109 but still redirected to login with #1692

PR #2156, opened 2020-04-17
remove the deprecated automatic route feature

xabbuh:

PR #2162, opened 2020-04-17
remove obsolete routing fixtures

Tobion:
Continuation of #2156


xabbuh:
I think we can also remove the XML schema definition files now.

xabbuh:
I have also merged #2163. So we could remove the remaining deprecated annotations here too.

Tobion:
Done.

PR #2166, opened 2020-04-20
Update 2-the-view-layer.rst

yyaremenko:
hi there,
I did some simplifications to make this section a bit more readable (especially for dumbs like me) and fixed some formatting and URLs

also, I have a question - there are at least two pages currently available on the Symfony documentation site:
https://symfony.com/doc/master/bundles/FOSRestBundle/5-automatic-route-generation_single-restful-controller.html
https://symfony.com/doc/master/bundles/FOSRestBundle/6-automatic-route-generation_multiple-restful-controllers.html

but they seem to be unavailable in the repository; if they were deleted from the repo, I would strongly suggest removing them from the site, too, because they confuse really a lot

### DIFF
--- a/Resources/doc/2-the-view-layer.rst
+++ b/Resources/doc/2-the-view-layer.rst
@@ -1,24 +1,35 @@
-Step 2: The view layer
+Step 2: The View Layer
 ======================
 
-Introduction
-------------
+Why View Layer
+--------------
 
-The view layer makes it possible to write ``format`` (html, json, xml, etc)
-agnostic controllers, by placing a layer between the Controller and the
-generation of the final output via a serializer.
+The View Layer resides between the Controller and the Serializer:
 
-The bundle works both with the `Symfony Serializer Component`_ and the more
-sophisticated `serializer`_ created by Johannes Schmitt and integrated via the
-`JMSSerializerBundle`_.
+``Controller --> View Layer --> Serializer --> [final output]``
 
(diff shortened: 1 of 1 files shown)


xabbuh:
Thank you for the PR @yyaremenko. I will look into it later.

About the removed documentation: Can you confirm that you no longer see them on the website? We updated the documentation during the last days and it might be that it just took some time that those changes were reflected on the website.

yyaremenko:
hi @xabbuh,
thank you for the prompt reply; 

yes, I can confirm now - the links are no longer available on the website (though they were available by the moment I created a pull request)

it's none of my business, but maybe it would be nice to return HTTP 301 redirect, instead of 404, to the main documentation page of the FOSTRestBundle for such removed links - that'd be good both for the searching engine indexation and a person who googles for the documentation... just an idea.

yyaremenko:
also, @xabbuh, since I have a chance to speak to you directly, I have one more question:
https://symfony.com/doc/master/bundles/FOSRestBundle/7-manual-route-definition.html

says, in the very beginning
> If the automatic route generation does not fit your needs

but I found nothing about the automatic route generation made by the FOSRestBundle in the current documentation (except for the now deleted documentation we discussed above)

probably I looked in the wrong direction and I am missing a thing? or perhaps there should be some update to the documentation regarding how the routes are auto-generated by the bundle (with some practical examples)

PR #2168, opened 2020-04-22
Provide the formats as parameter to be reusable in other bundles

alexander-schranz:
For creating the FOSRestRoutingBundle I need a way to access the formats of the FOSRestBundle.

### DIFF
--- a/DependencyInjection/FOSRestExtension.php
+++ b/DependencyInjection/FOSRestExtension.php
@@ -330,6 +330,8 @@ class FOSRestExtension extends Extension
             }
         }
 
+        $container->getDefinition('fos_rest.view_formats_provider')->replaceArgument(0, $formats);
+
         if ($config['routing_loader']['enabled']) {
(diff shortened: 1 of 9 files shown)


xabbuh:
Thanks for the PR! I wonder if it could be worth to add a service providing access to the configured formats instead. Using parameters always comes with the drawback that it's nearly impossible to properly deprecate them.

alexander-schranz:
@xabbuh sure I can create a service for it and use expression to access inject the formats e.g.:

```xml
<argument type="expression">service('fos_rest.formats_provider').getFormat()</argument>
```

Where do you want me to place it? How you would name it `FormatsProvider`?

xabbuh:
👍 for `FormatsProvider`, and I think 2.8 is a good target to not introduce new features in 3.0 that are not present in 2.8.

alexander-schranz:
@xabbuh ready for review!

alexander-schranz:
Ok we did solve the issue using a compiler pass when the FOSRestBundle is available and read the formats from the ViewHandler service.

PR #2172, opened 2020-04-23
Remove accidentally added return type from ExceptionHandler

alexander-schranz:
In #2126 this was accidently added to a protected function which could be overwritten.


xabbuh:
Good catch, thanks @alexander-schranz.

PR #2173, opened 2020-04-24
remove expression-language suggest dependency

Tobion:
This is not needed anymore and not specific to rest-bundle but applies to symfony/routing in general when using a "condition" option.


xabbuh:
for some reason tests fail with this change

Tobion:
There was one controller annotation that uses the condition for version check which made all the tests fail. So we need to keep the dev dependency but I removed it from suggest.

xabbuh:
Thank you @Tobion.

Issue #2167, opened 2020-04-20
Mapped Exception Return status code instead of given code


alexander-schranz:
Testing current master branch (pre 3.0) with sulu.

We have an exception which look like this:

https://github.com/sulu/sulu/blob/00d86acd9dc7c1f91c7561497fe0f7262456ad5c/src/Sulu/Bundle/CategoryBundle/Exception/KeywordIsMultipleReferencedException.php#L31

It has a `$code` attribute which is set to `2002`. And this exception is mapped to the statuscode [409](https://github.com/sulu/sulu/blob/2cccc18a10d948feec020463b48274fb7c76477f/src/Sulu/Bundle/CategoryBundle/DependencyInjection/SuluCategoryExtension.php#L56).

Now instead of getting [2002](https://github.com/sulu/sulu/blob/00d86acd9dc7c1f91c7561497fe0f7262456ad5c/src/Sulu/Bundle/CategoryBundle/Tests/Functional/Controller/KeywordControllerTest.php#L295) in the tests the json returns `409`.

xabbuh:
Could you check if the changes from #2161 applied on `master` would solve the issue here?

xabbuh:
@alexander-schranz #2161 should be fully available on `master` now. Can you check if the mapping works as expected?

alexander-schranz:
@xabbuh seems like we did decorate the old exception listener and the `$code` was set by us to the `$exception->getCode()`.
Seems like I need to decorate the new `FlattenExceptionNormalizer` to set again  the exception code instead of the http status code.

The mapping works now correct.

alexander-schranz:
@xabbuh The Exception Status Code mapping works but when running `--prefer-lowest` which will use symfony 4.4 the FlattenExceptionNormalizer is never called and so `message` and `code` is not set returned.

xabbuh:
Any idea what is happening? Is the `ErrorRender` actually called? Will the `SerializerErrorRenderer` be used?

alexander-schranz:
@xabbuh no it seems there is only the deprecated TwigBundle:ExceptionController called in Symfony 4.4

alexander-schranz:
@xabbuh when disabling it with:

```php
        if ($container->hasExtension('twig') && class_exists(ExceptionController::class)) {
            // Disable the deprecated exception contorller
            $container->prependExtensionConfig('twig', [
                'exception_controller' => null,
            ]);
        }
```

[...]

PR #2178, opened 2020-04-27
Throw exception for false combination of twig exception controller and exception_listener configuration

alexander-schranz:
By default in symfony 4.4 the older exception controller is used so the error handler component is not used. If the exceptionListener is disabled the newer ErrorController need to be used else the mapping configuraiton and listeners don't work.

This would fix https://github.com/FriendsOfSymfony/FOSRestBundle/issues/2167#issuecomment-619902695 where on `prefer-lowest` and disabled `exception_listener` the error handler is not called and so the exception not correctly serialized

### DIFF
--- a/DependencyInjection/Compiler/TwigExceptionPass.php
+++ b/DependencyInjection/Compiler/TwigExceptionPass.php
@@ -24,6 +24,14 @@ final class TwigExceptionPass implements CompilerPassInterface
 {
     public function process(ContainerBuilder $container)
     {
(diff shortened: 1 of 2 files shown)


xabbuh:
I think we shouldn't silently change the user's configuration as it might have impact on their application which would not be easy to debug. Instead I think we should throw an error during container compilation when the exception listener is disabled, but the deprecated `ExceptionController` from TwigBundle is still being used.

alexander-schranz:
@xabbuh PR updated!

xabbuh:
Thank you @alexander-schranz.

PR #2181, opened 2020-05-05
fix triggering deprecations if routing loader is enabled

xabbuh:


DependencyInjection/Configuration.php
```diff
@@ -118,6 +117,13 @@ public function getConfigTreeBuilder(): TreeBuilder
                         })
                     ->end()
                     ->children()
+                        ->booleanNode('enabled')
+                            ->defaultValue(function () {
+                                @trigger_error('Enabling the route generation feature is deprecated since FOSRestBundle 2.8.', E_USER_DEPRECATED);
```
Tobion, line 122: doesn't this trigger two deprecations now? one for the default case and one in normalization.

xabbuh: I don't think so. IIRC normalization is only applied on user defined values and default values are only applied after the user config has been processed.

PR #2182, opened 2020-05-06
add a SerializerErrorRenderer

xabbuh:


ErrorRenderer/SerializerErrorRenderer.php
```diff
@@ -31,7 +32,8 @@ public function __construct(SerializerInterface $serializer, RequestStack $reque
     public function render(\Throwable $exception): FlattenException
     {
         $flattenException = FlattenException::createFromThrowable($exception);
+        $format = SymfonySerializerErrorRenderer::getPreferredFormat($this->requestStack)($flattenException);
 
-        return $flattenException->setAsString($this->serializer->serialize($flattenException, SerializerErrorRenderer::getPreferredFormat($this->requestStack)($flattenException)));
+        return $flattenException->setAsString($this->serializer->serialize($flattenException, $format, new Context()));
```
Tobion, line 37: this error renderer replaces the symfony one right. so i guess it would need the same functionality:
- fallback renderer to support html
- adding the headers
- handling debug flag (also for compatibility with the error preview functionality)

xabbuh: That was  indeed the plan. I have no integrated all of this here in this PR.

PR #2184, opened 2020-05-07
add access denied listener tests

xabbuh:


xabbuh:
Thank you @lsmith77.

lsmith77:
shall we close https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2109 then ?
I will also look into the topic of removing the previous exception in another PR .. but looks like I will only get to it over the weekend.

xabbuh:
I think I am going to merge #2109 as well as #2182 tonight so that we can tag the first beta releases. We can still add tests for it later in another PR.

PR #2187, opened 2020-05-07
Remove requirement to symfony/debug

alexander-schranz:
In the 3.0 there seems no longer a need to require `Symfony\Component\Debug`.


xabbuh:
Thank you @alexander-schranz.

PR #2188, opened 2020-05-08
Prepare body_listener default change and fix error renderer

Tobion:
if the symfony/serializer is not installed, the error_renderer.serializer alias is not used by symfony and so the custom one from rest-bundle wouldn't be active at all. we need to change the error_renderer alias directly

also jms throws a different error for unsupported formats which needs to be caught to trigger the fallback.

the error render also need to set headers when serializing errors to make sure symfony doesn't use the default html content type for json/xml (when the error happens before the FormatListener sets the request format)


xabbuh:
Thank you @Tobion.

PR #2193, opened 2020-05-10
make sure tests fail without access denied listener

lsmith77:
core already seems to do all the right things when using basic auth and/or json_login. so at least for 3.0, imho we should just remove the AccessDeniedListener and tell people to use `json_login` or adapt whatever custom solution they have.

however when using a custom guard authenticator it does not work properly out of the box unless one enables the the AccessDeniedListener and only with the removed `$exception` propagation done in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2109

that being said an alternative and likely cleaner approach in place of the AccessDeniedListener would be to offer a default implementation for a failure handler. so maybe for 3.0 this is what should be done? ie. drop AccessDeniedListener and add a failure handler? then we avoid even going into the exception handling process and directly return the appropriate response.

and unrelated issue I noticed is if I enable the `exception_listener` in the tests (ie. set `true` here https://github.com/FriendsOfSymfony/FOSRestBundle/blob/fix-access-denied-tests/Tests/Functional/app/AccessDeniedListener/config.yml#L14), then the explode with a circular dependency in the `AbstractNormalizer`:
```
Symfony\Component\Serializer\Exception\CircularReferenceException: A circular reference has been detected when serializing the object of class "Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException" (configured limit: 1).

/FOSRestBundle/vendor/symfony/serializer/Normalizer/AbstractNormalizer.php:330
...
```

### DIFF
--- a/Tests/Functional/AccessDeniedListenerTest.php
+++ b/Tests/Functional/AccessDeniedListenerTest.php
@@ -19,6 +19,10 @@ class AccessDeniedListenerTest extends WebTestCase
 
     public static function setUpBeforeClass()
     {
+        if (!interface_exists(ErrorRendererInterface::class)) {
+            self::markTestSkipped();
+        }
+
         parent::setUpBeforeClass();
         static::$client = static::createClient(['test_case' => 'AccessDeniedListener']);
     }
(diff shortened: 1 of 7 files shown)


xabbuh:
Thank you Lukas.

PR #2226, opened 2020-05-30
clean up the readme file

xabbuh:
I suggest to remove stuff that tends to outdate too often and no longer promote deprecated features.


README.md
```diff
@@ -5,10 +5,8 @@ This bundle provides various tools to rapidly develop RESTful API's &
 applications with Symfony. Features include:
```
Tobion, line 5: you could list the exception.messages and the rfc compatible error rendering that works with both the symfony/serializer as well as jms

PR #2227, opened 2020-05-30
Deprecate the behavior of the JMSHandlerRegistry to improve performance in v3

GuilhemN:
Sorry I am quite late at submitting changes for v3...

I think it wasn't a good idea to introduce such an handler in the first place (sorry for that) and I think it should be reworked to do just what we want it to: inherit the FlattenException handler.

As @goetas noted (a while ago now) in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1899#discussion_r233769559, it has terrible performance issues and I think v3 would be a great opportunity to fix it.

I hope it's still ok for you but if not, well I missed the boat :)

### DIFF
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -110,7 +110,7 @@ final class Configuration implements ConfigurationInterface
                             ->prototype('scalar')->end()
                         ->end()
                         ->booleanNode('serialize_null')->defaultFalse()->end()
-                        ->booleanNode('disable_custom_jms_registry')->defaultFalse()->end()
+                        ->booleanNode('enable_jms_registry')->defaultTrue()->end()
                     ->end()
                 ->end()
                 ->arrayNode('zone')
--- a/DependencyInjection/Configuration.php
+++ b/DependencyInjection/Configuration.php
@@ -110,6 +110,7 @@ final class Configuration implements ConfigurationInterface
                             ->prototype('scalar')->end()
                         ->end()
                         ->booleanNode('serialize_null')->defaultFalse()->end()
+                        ->booleanNode('disable_custom_jms_registry')->defaultFalse()->end()
                     ->end()
                 ->end()
                 ->arrayNode('zone')
--- a/DependencyInjection/FOSRestExtension.php
+++ b/DependencyInjection/FOSRestExtension.php
@@ -363,8 +363,8 @@ class FOSRestExtension extends Extension
         $options['serializeNullStrategy'] = $config['serializer']['serialize_null'];
         $viewHandler->addArgument($options);
 
-        if ($config['serializer']['disable_custom_jms_registry']) {
-            $container->setParameter('fos_rest.serializer.disable_custom_jms_registry', true);
+        if ($config['serializer']['enable_jms_registry']) {
+            $container->setParameter('fos_rest.serializer.enable_jms_registry', true);
         }
     }
 
--- a/DependencyInjection/FOSRestExtension.php
+++ b/DependencyInjection/FOSRestExtension.php
@@ -362,6 +362,10 @@ class FOSRestExtension extends Extension
 
         $options['serializeNullStrategy'] = $config['serializer']['serialize_null'];
         $viewHandler->addArgument($options);
+
+        if ($config['serializer']['disable_custom_jms_registry']) {
+            $container->setParameter('fos_rest.serializer.disable_custom_jms_registry', true);
+        }
     }
 
     private function loadZoneMatcherListener(array $config, XmlFileLoader $loader, ContainerBuilder $container): void
--- a/Serializer/JMSHandlerRegistry.php
+++ b/Serializer/JMSHandlerRegistry.php
@@ -59,7 +59,7 @@ class JMSHandlerRegistry implements HandlerRegistryInterface
                 return $handler;
             }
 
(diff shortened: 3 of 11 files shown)


goetas:
I have talked to @xabbuh over slack about this:
> imo should be done in a major, because with that handler you have introduced a new feature in jms :slightly_smiling_face:
the ability to define handlers for a class based on its parent

I'm :+1: for this change

xabbuh:
I think we should not block the 3.0 release with this PR and first discuss a bit what we want to achieve here. For example, would it make sense to add an option to completely disable our custom registries (in 3.1 then)? If I don't miss anything, they are not that useful anymore.

GuilhemN:
> would it make sense to add an option to completely disable our custom registries (in 3.1 then)

You're right they could be completely removed, in 3.x the bundle only uses instances of `FlattenException`.
The creation of an option would delay their removal to 4.0 but would still allow perf improvement. Its use could be encouraged through the recipe, and maybe through a deprecation notice, so I'm in favor of doing this :+1: It would also avoid unneeded pressure to finish and review this PR on time :)

I'll modify this PR accordingly then!

xabbuh:
👍 sounds like a good plan

DependencyInjection/FOSRestExtension.php
```diff
@@ -362,6 +362,10 @@ private function loadSerializer(array $config, ContainerBuilder $container): voi
 
         $options['serializeNullStrategy'] = $config['serializer']['serialize_null'];
         $viewHandler->addArgument($options);
+
+        if ($config['serializer']['disable_custom_jms_registry']) {
```
GuilhemN: I'm not sure whether we should trigger a deprecation warning in case this option is not activated as it will create noise for the user but shouldn't make the removal of the custom registry safer (the deprecation triggered [here](https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2227/files#diff-87bb7b48eab03df25a3b87a9dd7e44ccR61) ensures it has no effect).

Serializer/JMSHandlerRegistry.php
```diff
@@ -51,11 +53,18 @@ public function registerHandler($direction, $typeName, $format, $handler): void
      */
     public function getHandler($direction, $typeName, $format)
     {
+        $first = true;
         do {
             $handler = $this->registry->getHandler($direction, $typeName, $format);
             if (null !== $handler) {
+                if (!$first) {
+                    @trigger_error(sprintf('Relying on the custom registry %s to inherit the JMS handler of type `%s` is deprecated since FOSRestBundle 3.1. It will be removed in version 4.0. Use the option `fos_rest.serializer.disable_custom_jms_registry` to disable it.', __CLASS__, $typeName), E_USER_DEPRECATED);
```
xabbuh: Are we sure this deprecation will always be triggered? I could imagine that's not the case if `$first` never becomes `false`. I tend to think that triggering a deprecation when the option is set to `false` would be the better approach. Deprecating this class is not really necessary (at least not triggering a deprecation) as it is already marked as internal.

DependencyInjection/Configuration.php
```diff
@@ -110,6 +110,7 @@ public function getConfigTreeBuilder(): TreeBuilder
                             ->prototype('scalar')->end()
                         ->end()
                         ->booleanNode('serialize_null')->defaultFalse()->end()
+                        ->booleanNode('disable_custom_jms_registry')->defaultFalse()->end()
```
xabbuh: To be in line with the names of the other config options I think I would rename this one to `enable_jms_registry` and set it to `true` by default.

Serializer/JMSHandlerRegistry.php
GuilhemN: If the deprecation is not triggered it is either because
- The handler returned is the one expected (the one returned by the jms registry)
- No handler is returned and the behavior is thus the same as the jms registry

I wanted to deprecate is relying on the behavior of this class, so as long as the result is the same as what would jms return I believe it's ok to not trigger a deprecation.
But if we want to warn users they could improve performance by disabling this class, triggering a deprecation when `$first` would indeed be legit.

So in my opinion it depends whether we want to warn users about the performance improvement possibility or not. I think it is better to privilege DX but I would appreciate having your opinion on this point :)

DependencyInjection/Configuration.php
GuilhemN: I agree, it's updated. Thanks!

GuilhemN:
The travis failure is quite strange, I don't think it's related to this PR.

Serializer/JMSHandlerRegistry.php
xabbuh: I think we should always trigger the deprecation to make users aware of the performance penalty this may have.

GuilhemN: ok, it's updated :)

UPGRADING-3.0.md
```diff
@@ -156,3 +156,6 @@ Upgrading From 2.x To 3.0
    * `FOS\RestBundle\View\JsonpHandler`
    * `FOS\RestBundle\View\View`
    * `FOS\RestBundle\View\ViewHandler`
+
+ * The JMS Handlers `JMSHandlerRegistry` and `JMSHandlerRegistryV2` behaviors have changed.
```
W0rma: @GuilhemN I guess this should be moved to the CHANGELOG instead since 3.0 has already been released

GuilhemN: Indeed, updated

goetas:
closing in favor of https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2319 for jms/serializer-bundle v4

PR #2228, opened 2020-05-31
Improve deprecations DX

GuilhemN:
While trying to add support for FOSRestBundle 3.x to NelmioApiDocBundle, I found a few deprecations that could be improved:
- the deprecations from the routing services don't specify how to solve them, I added a reference to a config field
- deprecations of suboptions of `fos_rest.exception` even if `exception` is not enabled. I solved it by using `validate` instead of `defaultValue` + `beforeNormalization` at the level of `exception`.


xabbuh:
Thank you @GuilhemN.

PR #2230, opened 2020-06-02
Added support for query string parameter converter.

nuvolapl:
### Usage
`config/packages/fos_rest.yaml`:
```yaml
fos_rest:
    query_string_converter:
        enabled: true
# (...)
```

`src/Controller/ExampleController`:
```php
<?php

declare(strict_types=1);

namespace App\Controller;

use FOS\RestBundle\Pagination\LimitOffsetPagination;
use FOS\RestBundle\Filter\IdFilter;
use Sensio\Bundle\FrameworkExtraBundle\Configuration\ParamConverter;

class ExampleController
{
    /**
     * @ParamConverter("pagination", converter="fos_rest.query_string")
     * @ParamConverter("filter", converter="fos_rest.query_string")
     */
    public function cget(LimitOffsetPagination $pagination, IdFilter $filter): iterable
    {
        // (...)
    }
}
```

### Example request
```bash
curl 'http://localhost/api/examples?pagination[limit]=1&pagination[offset]=2&filter[id]=526e8fc9-9ba2-493b-ab96-4213309ec82f'
```


lopesrichard:
I've been searching for the same solution. But sounds better using the query params without the array sintax.

Ex: `curl 'http://localhost/api/examples?limit=1&offset=2'`

nuvolapl:
@xabbuh @GuilhemN 
image

@lopesrichard Are you sure? I have a different opinion. Grouped parameters are much more user-friendly in the code.

goetas:
I think that this implementation is too much opinionated.  Pagination has always been a tricky topic.

- `pagination[limit]=1&pagination[offset]=2`
- `limit=1&offset=2'`
- `page=1&size=100'`
- `page=1'`
- The `Range` header
- whatever...


Unless there is a general agreement on what should be done, I would prefer to now add this into the core of this bundle

maybe only the query string conversion could be extracted.

PR #2233, opened 2020-06-04
improve route loader upgrade instructions

### LINKED ISSUE #2165: Add upgrade instructions for deprecated routing loader
We need better upgrade instructions for switching from the automatic route generation loader to SF Core routing loader.

One part:
Using the method shortcuts also works when using the symfony core annotation loader as they extend the core Route annotation. So we can keep them in rest-bundle and don't need to deprecate them. The only thing that changes is that it changes the route name as rest-bundle used a different naming convention and it removes the `.{_format}` suffix from routes if you used `routing_loader.include_format` option (which was enabled by default).

Before:
```
Acme\Controller\TestController:
    type: rest
    resource: Acme\Controller\TestController
```

After:
```
Acme\Controller\TestController:
    type: annotation
    resource: Acme\Controller\TestController
```

Migrate to explicit routes with the help of `bin/console debug:router --show-controllers`

Tobion:
Fixes #2165

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,8 +30,34 @@ CHANGELOG
        routing_loader: false
    ```
 
-   You need to configure your routes explicitly or consider using the
-   [RestRoutingBundle](https://github.com/handcraftedinthealps/RestRoutingBundle).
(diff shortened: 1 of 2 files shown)


xabbuh:
Thank you Tobias.

PR #2263, opened 2020-08-26
Allow to specify Throwable as exception class

a-menshchikov:
Currently we can't use Throwable in exception.codes because check method uses `is_subclass_of`.
This PR fixes this issue.


xabbuh:
Can you please add a test case to ensure that we don't break this accidentally in the future? And can you then please rebase and retarget on `2.x` as I think this fix should be included in 2.8.x too?

a-menshchikov:
@xabbuh done.

xabbuh:
Thank you @a-menshchikov.

PR #2266, opened 2020-09-02
Update version match documentation

kjkooistra-youwe:
Matching version argument does not work, as it's null.
Use requirements instead.

Also see https://github.com/symfony/symfony/issues/38034

### DIFF
--- a/Resources/doc/versioning.rst
+++ b/Resources/doc/versioning.rst
@@ -152,10 +152,11 @@ That's it, it should work now.
 How to match a specific version in my routing ?
 -----------------------------------------------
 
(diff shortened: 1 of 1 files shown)


xabbuh:
I think we should have both examples. If the version is not derived from the path, you will still need to use the `condition` option. You can take a look at the `Version2Controller` class in the tests which contains examples for both.

kjkooistra-youwe:
Hmm, but won't that require a more structural update? How will that work with multiple options enabled, for instance? https://symfony.com/doc/3.x/bundles/FOSRestBundle/versioning.html#how-to-version-your-api

PR #2276, opened 2020-11-27
Bump willdurand/negotiation to ^3.0 to allow PHP 8.0

michaljusiega:
As explained in: https://github.com/willdurand/Negotiation/releases/tag/3.0.0 `FOSRESTBundle` blocks update and allow to use in PHP 8.

```
composer depends willdurand/negotiation
friendsofsymfony/rest-bundle  3.0.3  requires  willdurand/negotiation (^2.0)
```

There is no side effects (I think) because main change was in internal class from `Match` (which is a reserved keyword) to `AcceptMatch`

### DIFF
--- a/Negotiation/FormatNegotiator.php
+++ b/Negotiation/FormatNegotiator.php
@@ -40,7 +40,7 @@ class FormatNegotiator extends BaseNegotiator
         $this->map[] = [$requestMatcher, $options];
     }
 
-    public function getBest($header, array $priorities = [])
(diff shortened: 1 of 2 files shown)


michaljusiega:
CI is not related, I think.

@xabbuh friendly ping ;)

michaljusiega:
Closed in favor of https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2278

PR #2280, opened 2020-12-03
bump PHPUnit version and fix php 8 support

xabbuh:


GuilhemN:
:+1: I did not realize you were still maintaining v2. So yes it's better to do this in 2.x to simplify future merge conflicts.

GuilhemN:
I think we can safely merge this :)

kevinpapst:
I know at least one project that would be very happy to see this tagged (in v3) 😁 and I guess there are 100+ more who are just waiting silently.

Thanks for taking care of it! 👍

GuilhemN:
https://github.com/FriendsOfSymfony/FOSRestBundle/releases/tag/3.0.4 done :)

kevinpapst:
Thank you both!!!!

PR #2285, opened 2020-12-17
failsafe check for php 8

### LINKED ISSUE #2284: Problem with 2.x on PHP 7.x
Hi @GuilhemN ,

your fix for PHP 8 in the ClassUtils tokenizer here:
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/d60e0ba3cc1f3dce11cedf65d8535cb17c506bd3/Routing/Loader/ClassUtils.php#L27-L31
seems to work in a standalone project.

The problem with this check is, that some libraries define this constant in user land code, see [phpDocumentor](https://github.com/phpDocumentor/TypeResolver/blob/1.x/src/Types/ContextFactory.php#L46) and [swagger-php](https://github.com/zircote/swagger-php/blob/master/src/functions.php#L22) for example (these two are used in a project of mine). And with this constant defined in user land code, the tokenizer breaks and the first namespace token is always `true`, resulting in a broken namespace `1\\MyController` (see this [CI run for example](https://github.com/kevinpapst/kimai2/pull/2158/checks?check_run_id=1567511121)).

At least the Swagger bundle is likely be used in other projects that also use the Rest bundle.

Now we could argue that the 2.x branch is kind of deprecated and that defining this constant in user land code is the real problem, BUT other libraries fix the problem in another way (a quick google searched brought up [this Drupal code](https://api.drupal.org/api/drupal/vendor%21doctrine%21annotations%21lib%21Doctrine%21Common%21Annotations%21TokenParser.php/function/TokenParser%3A%3AparseNamespace/9.2.x)

Instead of using `if (defined('T_NAME_QUALIFIED')) {` we could simply use `if (PHP_VERSION_ID >= 80000) {`.

Would you accept a PR for this? I already added one for the [handcrafterinthealps routing bundle](https://github.com/handcraftedinthealps/RestRoutingBundle/pull/8) to support users of the 3.x branch.

kevinpapst:
fixes #2284

### DIFF
--- a/Routing/Loader/ClassUtils.php
+++ b/Routing/Loader/ClassUtils.php
@@ -24,7 +24,7 @@ class ClassUtils
         $namespace = false;
         $tokens = token_get_all(file_get_contents($file));
 
-        if (defined('T_NAME_QUALIFIED')) {
+        if (PHP_VERSION_ID >= 80000) {
             $namespaceToken = T_NAME_QUALIFIED;
         } else {
             $namespaceToken = T_STRING;


JanStorm:
We had the same issue, can confirm this fixes it!

GuilhemN:
Thank you @kevinpapst!
And released as 2.8.5.

kevinpapst:
Thanks for the quick merge and release @GuilhemN

PR #2297, opened 2021-03-10
POST method should not throw exception on empty payload without content-type

fliespl:
Content-Type header field is used to specify type of data in the body request.

When request is sent with an empty payload - it should be accepted the same way as DELETE request, cause in such cases Content-Type header is not applicable.


goetas:
hi, 
thanks for your PR. this makes sense to me.

could you add a test for this please?

PR #2298, opened 2021-03-16
Responses with HTTP status 204/ 304 must not set the Content-Type header

kryptychon:
Hi there,

IMHO the `Content-Type` header must not be set if the `ViewHandler` is dealing with an empty `Response`.

I'm aware this might break BC but nevertheless, a Content-Type header does not really make sense in an empty response.


goetas:
makes sense, could you please add a test for this change?

PR #2306, opened 2021-04-23
ensure compatibility with Doctrine ORM 2.8

### LINKED ISSUE #2302: EntityToIdObjectTransformer incompatible with doctrine/orm v2.8.X
## Issue

The first constructor parameter of class EntityToIdObjectTransformer uses the type declaration `Doctrine\Common\Persistence\ObjectManager`.

This causes a **TypeError** to be thrown when a ObjectManager implementation that uses doctrine/persistence v2.X is passed. 

## Example

*Error message:* 
`Argument 1 passed to FOS\RestBundle\Form\Transformer\EntityToIdObjectTransformer::__construct() must be an instance of Doctrine\Common\Persistence\ObjectManager, instance of Doctrine\ORM\EntityManager given ...`

Here Doctrine\ORM\EntityManager implements Doctrine\ORM\EntityManagerInterface which in turn extends Doctrine\Persistence\ObjectManager.

## Cause

The doctrine/common package was separated into multiple packages (see https://github.com/doctrine/common/issues/826). Use of the namespace `Doctrine\Common\Persistence` was deprecated and use of the **new package namespace** `Doctrine\Persistence` encouraged. The backwards compatibility layer for this deprecated namespace was eventually removed in v2.0.0 of doctrine/persistence (see https://github.com/doctrine/persistence/pull/106).

Both doctrine/orm v2.8.0 and doctrine/common v3.0.0, switched to version 2 of doctrine/persistence (see https://github.com/doctrine/orm/pull/8166 and https://github.com/doctrine/common/pull/890).

## Mitigation
Current mitigation for developers using doctrine/orm and friendsofsymfony/rest-bundle is to restrict the dependencies in their projects as follows:

*File composer.json:*
```json
"require": {
    ...
    "doctrine/common": "^2.13",
    "doctrine/orm": "2.7.*",
    "doctrine/persistence": "^1.3",
    ...
}
```

xabbuh:
fixes #2302

### DIFF
--- a/Form/Transformer/EntityToIdObjectTransformer.php
+++ b/Form/Transformer/EntityToIdObjectTransformer.php
@@ -11,7 +11,8 @@
 
 namespace FOS\RestBundle\Form\Transformer;
 
-use Doctrine\Common\Persistence\ObjectManager;
+use Doctrine\Common\Persistence\ObjectManager as LegacyObjectManager;
+use Doctrine\Persistence\ObjectManager;
 use Symfony\Component\Form\DataTransformerInterface;
 use Symfony\Component\Form\Exception\TransformationFailedException;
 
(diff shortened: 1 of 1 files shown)


Form/Transformer/EntityToIdObjectTransformer.php
```diff
@@ -27,8 +28,12 @@ class EntityToIdObjectTransformer implements DataTransformerInterface
     private $om;
     private $entityName;
 
-    public function __construct(ObjectManager $om, string $entityName)
+    public function __construct($om, string $entityName)
```
marmichalski, line 31: As the minimum required PHP version is 7.2, couldn't you use `object` typehint here? This would allow to remove `is_object` check. Unless that's on purpose, for some reason 🤔

GuilhemN:
Thank you @xabbuh!

PR #2307, opened 2021-05-06
Fixed bug for PHP 8

Victor-Truhanovich:
Attempted to call an undefined method named "isBuiltin" of class "ReflectionUnionType".

### DIFF
--- a/EventListener/ParamFetcherListener.php
+++ b/EventListener/ParamFetcherListener.php
@@ -88,12 +88,18 @@ class ParamFetcherListener
     private function isParamFetcherType(\ReflectionParameter $controllerParam): bool
     {
         $type = $controllerParam->getType();
-        if (null === $type || $type->isBuiltin() || !$type instanceof \ReflectionNamedType) {
-            return false;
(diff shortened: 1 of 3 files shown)


xabbuh:
Can you please add a test case that would fail without this change?

Victor-Truhanovich:
```
/**
 * @Rest\Route("accounts")
 */
class AccountController extends AbstractFOSRestController implements ApiControllerInterface
{
    /**
     * @Rest\Get("/{id}", requirements={"id"="\d+"})
     *
     * @param Account|EntityInterface $account
     *
     * @return View
     */
    public function read(Account|EntityInterface $account): View
    {
        return $this->accountResponseManager->create($account);
    }
}
```

Victor-Truhanovich:
> Can you please add a test case that would fail without this change?

Sorry, I added tests for my code, but I never got an answer from you

xabbuh (approved):

xabbuh:
Thank you @Viktor-Truhanovich.

PR #2315, opened 2021-07-15
Update psr/log to allow new versions and only require in dev

mbabker:
This PR does two things:

1) Moves the `psr/log` dependency to `require-dev`, the only places in the repo that use the `Psr\Log` namespace are a docs example and the test kernel, meaning the dependency isn't actually a production dependency.

2) Allows for the new versions of the `psr/log` package to be used when able (which, probably won't be the case for a while thanks to the dependency chain, but inside this bundle, there's nothing directly blocking it).

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -30,7 +30,6 @@
     },
     "require": {
         "php": "^7.2|^8.0",
-        "psr/log": "^1.0",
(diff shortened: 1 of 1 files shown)


xabbuh:
Thank you @mbabker.

Issue #1812, opened 2017-12-06
Using FOS\RestBundle\Controller\Annotations\View throws exception in Symfony 4


matigda:
I'm migrating from Sf 3 to Sf4, this is how my controller looks like

```

use FOS\RestBundle\Controller\Annotations\Post;
use FOS\RestBundle\Controller\Annotations as Rest;

... some code

    /**
     * @Post("/users")
     * @Rest\View(statusCode=201)
     */
    public function registerAction(Request $request)
    {
          ... some action
    }
```

And then I receive this exception:

> Warning: ReflectionObject::__construct() expects parameter 1 to be object, null given in vendor/friendsofsymfony/rest-bundle/EventListener/ViewResponseListener.php line 162

This same code works on Symfony 3.3.10. Also - when I remove this - `@Rest\View(statusCode=201)` then everything works.

wibimaster:
+1, same problem with Symfony 3.4 (installed with Symfony Flex).
Installation of Twig bundle resolve the problem.
But it's ugly to have to install Twig for JSON rendered ;)

matigda:
Well I can actually make a PR to solve it but the question is how authors would like it to be resolved. Whether add this definition again in some compiler pass in this bundle or just create another listener which would do what should be done or ... some other solution which I haven't thought of ;)

DavidGarciaCat:
Hi all,

No more activity here? I have the same problem, using SF 3.4

xabbuh:
I think we should investigate if we could move/copy some logic to FOSRestBundle.

xabbuh:
Can you confirm that the change proposed in #1834 will give a better error message?

DavidGarciaCat:
Hi @xabbuh 

Regarding PR https://github.com/FriendsOfSymfony/FOSRestBundle/pull/1834

<img width="773" alt="screen shot 2018-01-04 at 23 02 40" src="https://user-images.githubusercontent.com/7809429/34588162-74f23e40-f1a3-11e7-8fd1-2be15bb7b8ac.png">

Quick question: given the PR description is pointing the `TwigBundle` and `FOSRestBundle` is a Bundle to create RESTful APIs (that nowadays usually return XML or JSON) what's the point to force me to install and enable a Bundle that deals with templates?

Maybe I don't understand the logic behind this, but it seems useless to me to throw an error just because I don't have the `TwigBundle` when I have no intention to manage anything related with that Bundle or templates.

[...]

PR #2320, opened 2021-08-06
Enable usage of the bundle without Twig

Th3Mouk:
As I made a fork, and following https://github.com/FriendsOfSymfony/FOSRestBundle/issues/1812#issuecomment-893277145, I open this PR which will evolve according to the feedbacks

### DIFF
--- a/DependencyInjection/Compiler/ConfigurationCheckPass.php
+++ b/DependencyInjection/Compiler/ConfigurationCheckPass.php
@@ -13,7 +13,6 @@ namespace FOS\RestBundle\DependencyInjection\Compiler;
 
 use Symfony\Component\DependencyInjection\Compiler\CompilerPassInterface;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
-use Sensio\Bundle\FrameworkExtraBundle\EventListener\TemplateListener;
 use Sensio\Bundle\FrameworkExtraBundle\EventListener\ParamConverterListener;
 
 /**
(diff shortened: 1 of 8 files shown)


GuilhemN:
Indeed since https://github.com/FriendsOfSymfony/FOSRestBundle/commit/73f0dfd6dde7bac0d399dd536805c404eb2f744d#diff-2359bb4681ba7bf807bf836fd67da56e5983f70efdc2bcca29768026d4c2352c removed support of templating I think we can now safely remove this condition to have twig installed.
So 👍🏻 for me.

GuilhemN (approved):

W0rma:
@Th3Mouk Should the dev dependency `symfony/twig-bundle` and all its usages in the tests be removed to check if the tests will still pass (https://github.com/FriendsOfSymfony/FOSRestBundle/blob/3.x/composer.json#L56)?

Th3Mouk:
@W0rma I didn't find any occurence with `use (.*)?Twig(.*)?;` so this dependency seems no longer needed

composer.json
```diff
@@ -53,7 +53,6 @@
         "symfony/yaml": "^4.4|^5.0",
         "symfony/security-bundle": "^4.4|^5.0",
         "symfony/web-profiler-bundle": "^4.4|^5.0",
-        "symfony/twig-bundle": "^4.4|^5.0",
```
W0rma: @Th3Mouk TwigBundle is still enabled in some tests, f.e. https://github.com/Th3Mouk/FOSRestBundle/blob/fix/without-twig/Tests/Functional/app/Version/bundles.php#L14

Even after this change `symfony/twig-bundle` is still installed since it is a dependency of `symfony/web-profiler-bundle`. 
I guess that's the reason why the tests are still green.

Furthermore from the test name I assume that https://github.com/Th3Mouk/FOSRestBundle/blob/fix/without-twig/Tests/Functional/RequestBodyParamConverterTest.php#L35 is related to TwigBundle.

Therefore I guess we need to keep the dev dependency but should just not enable the bundle for all test cases **except** https://github.com/Th3Mouk/FOSRestBundle/blob/fix/without-twig/Tests/Functional/app/RequestBodyParamConverterTwigBundle/bundles.php#L14

WDYT?

W0rma (changes requested):

composer.json
Th3Mouk: @W0rma I upgraded the PR, i tried to understand the remaining test case and see if i could add a missing one but all is ok, all seems covered

goetas:
thank you

PR #2312, opened 2021-06-21
Fix compatibility of Route annotations with symfony >= 5.3

W0rma:
Constructing Route annotations using named arguments was added in symfony 5.2 (https://github.com/symfony/symfony/pull/40266)

Closes https://github.com/FriendsOfSymfony/FOSRestBundle/issues/2311

PR #2321, opened 2021-08-09
Fix CI

W0rma:
This PR fixes some deprecations introduced in symfony 5.3.

The fixes are only related to the configuration of the functional tests.

Deprecations regarding the Route annotations will be fixed in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2312


GuilhemN:
That's an impressive work! Thanks a lot @W0rma :)

PR #2325, opened 2021-08-17
Add support for native PHP8 Route attributes

### LINKED ISSUE #2293: Incorrect annotation type-hinging
Hi. I'm using FOSRest version 3.0.x with new php 8 attributes.
I noticed that you extends the Symfony `Route` annotation, but miss all constructor arguments (except `data`)

```php
<?php

/*
 * This file is part of the FOSRestBundle package.
 *
 * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
 *
 * For the full copyright and license information, please view the LICENSE
 */

use Symfony\Component\Routing\Annotation\Route as BaseRoute;

/**
 * Route annotation class.
 *
 * @Annotation
 */
class Route extends BaseRoute
{
    {
        parent::__construct($data);

        if (!$this->getMethods()) {
            $this->setMethods((array) $this->getMethod());
        }
    }

    /**
     * @return string|null
     */
    public function getMethod()
    {
        return;
    }
}
```

But the original consutructor looks like:

```php
    /**
     * @param array|string      $data         data array managed by the Doctrine Annotations library or the path
     * @param array|string|null $path
     * @param string[]          $requirements
     * @param string[]          $methods
     * @param string[]          $schemes
     *
     * @throws \BadMethodCallException
     */
    public function __construct(
        $data = [],
        $path = null,
        string $name = null,
        array $requirements = [],
        array $options = [],
        array $defaults = [],
        string $host = null,
        array $methods = [],
        array $schemes = [],
        string $condition = null,
        int $priority = null,
        string $locale = null,
        string $format = null,
        bool $utf8 = null,
        bool $stateless = null
    ) {
       ....
       ....
    }
```
Because of what it is impossible to use such a expressions like this:

image

And more one notice in PhpStorm

image

To avoid this notice the annotation class shout be marked by annotation `Attribute`, like original Symfony `Route` class:

```php
/**
 * Annotation class for @Route().
 *
 * @Annotation
 * @Target({"CLASS", "METHOD"})
 *
 * @author Fabien Potencier <fabien@symfony.com>
 * @author Alexander M. Turek <me@derrabus.de>
 */
#[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD)]
class Route
```

**Not only this annotation has some problem.**

W0rma:
Closes #2293

### DIFF
--- a/Tests/Functional/Bundle/TestBundle/Controller/RouteAttributesController.php
+++ b/Tests/Functional/Bundle/TestBundle/Controller/RouteAttributesController.php
@@ -1,0 +1,55 @@
+<?php
+
+/*
+ * This file is part of the FOSRestBundle package.
+ *
+ * (c) FriendsOfSymfony <http://friendsofsymfony.github.com/>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace FOS\RestBundle\Tests\Functional\Bundle\TestBundle\Controller;
+
+use FOS\RestBundle\Controller\AbstractFOSRestController;
+use FOS\RestBundle\Controller\Annotations as Rest;
+use FOS\RestBundle\View\View;
+use Symfony\Component\HttpFoundation\Request;
+
+/**
+ * Controller to test native PHP8 Route attributes.
+ */
+#[Rest\Route('/products')]
+class RouteAttributesController extends AbstractFOSRestController
+{
+    /**
+     * @return View view instance
+     *
+     * @Rest\View()
+     */
+    #[Rest\Get(path: '/{page}', name: 'product_list', requirements: ['page' => '\d+'], defaults: ['_format' => 'json'])]
+    public function listAction(int $page)
+    {
+        $view = $this->view([
+            ['name' => 'product1'],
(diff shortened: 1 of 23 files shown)


GuilhemN:
Awesome, thank you @W0rma!

Tests/Functional/Bundle/TestBundle/Controller/RouteAttributesController.php
```diff
@@ -0,0 +1,55 @@
...
+ */
+#[Rest\Route('/products')]
+class RouteAttributesController extends AbstractFOSRestController
+{
+    /**
+     * @return View view instance
+     *
+     * @Rest\View()
+     */
+    #[Rest\Get(path: '/{page}', name: 'product_list', requirements: ['page' => '\d+'], defaults: ['_format' => 'json'])]
```
diegozavaletatelanto, line 30: Thank you!! can you update de documentation please?

W0rma: Please see https://fosrestbundle.readthedocs.io/en/3.x/annotations-reference.html#route

Issue #2301, opened 2021-03-26
Missing Attribute Annotation 


robmro27:
```
namespace FOS\RestBundle\Controller\Annotations;

/**
 * View annotation class.
 *
 * @Annotation
 * @Target({"METHOD","CLASS"})
 */
class View extends Template
{
```

```
#[View]
```

Maybe I missed something but above class not support new PHP 8 attributes ? Checked all classes in Annotation folder. This is not supported yet ?

xabbuh:
That's right, FOSRestBundle doesn't support attributes yet. Would you like to give this a try?

robmro27:
Hmm...checked some things but this will not be easy as for some annotations Symfony blocks you. Check in example `@Route`.
Symfony misses `\ReflectionAttribute::IS_INSTANCEOF` parameter so all annotations extending symfony build it `@Route` will not work. 

Maybe I am wrong, but this is the code that is problem IMO: https://github.com/symfony/symfony/blob/5.x/src/Symfony/Component/Routing/Loader/AnnotationClassLoader.php#L375

I think they need to add this to make `@Delete @Get @Head @Link @Patch @Post @Put @Unlink @Lock @Unlock @PropFind @PropPatch @Move @Mkcol @Copy` work

mbabker:
The other part of this that's tricky though is the changed `Symfony\Component\Routing\Annotation\Route` constructor signature to support attributes (the same changes in the validator is why in one of my newer OSS bundles I gave up on Symfony 4.4 support).  Symfony 5.2 expanded the constructor to support named arguments, then 5.3 deprecates the `$data` array in favor of the new arguments.  So the `FOS\RestBundle\Controller\Annotations\Route` constructor would need its own B/C layer of some kind or there's the potential of having to drop support for 4.4 if that ends up being too difficult to pull off.

Funny enough, the `\ReflectionAttribute::IS_INSTANCEOF` issue doesn't exist for the `@View` annotation; the framework extra bundle does use that when parsing attributes.

robmro27:
Added PR to SF https://github.com/symfony/symfony/pull/41014, it shows it will be merged in 5.* probably, work on this could be done after that...

robmro27:
There is probably another problem with `Params` annotations. Currently they could be like this, nesting Constraints. I think it's not possible to pass this Constraint similar way to attribute unfortunately. Maybe there was something similar somewhere and was resolved ?
```
/**
 * @QueryParam(
[...]

PR #2326, opened 2021-08-19
Add support for native PHP8 View attributes

W0rma:
Related to #2301

### DIFF
--- a/Resources/doc/annotations-reference.rst
+++ b/Resources/doc/annotations-reference.rst
@@ -68,17 +68,32 @@ FileParam
 View
 ----
 
-.. code-block:: php
+.. tabs::
 
-    use FOS\RestBundle\Controller\Annotations\View;
+   .. tab:: Annotations
 
-    /**
-     * @View(
-     *  statusCode=null,
-     *  serializerGroups={},
-     *  serializerEnableMaxDepthChecks=false
-     * )
-     */
+    .. code-block:: php
+
+        use FOS\RestBundle\Controller\Annotations\View;
+
+        /**
+         * @View(
+         *  statusCode=null,
+         *  serializerGroups={},
+         *  serializerEnableMaxDepthChecks=false
+         * )
+         */
+   .. tab:: Attributes
+
+    .. code-block:: php
+
+        use FOS\RestBundle\Controller\Annotations\View;
+
+        #[View(
+            statusCode: null,
+            serializerGroups: [],
+            serializerEnableMaxDepthChecks: false
+        )]
 
 Routing
 -------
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -70,7 +70,7 @@
     },
     "conflict": {
         "doctrine/annotations": "<1.12",
-        "sensio/framework-extra-bundle": "<5.2.3",
+        "sensio/framework-extra-bundle": "<6.1",
```
W0rma, line 74: I had to drop support for older versions since attributes support for the `Template` annotation was added in https://github.com/sensiolabs/SensioFrameworkExtraBundle/pull/707

Resources/doc/annotations-reference.rst
```diff
@@ -68,17 +68,29 @@ FileParam
 View
 ----
 
-.. code-block:: php
+.. configuration-block::
```
W0rma: I'm not sure if the usage of this syntax is allowed. I just took it from symfony-docs

Tests/Functional/Bundle/TestBundle/Controller/AttributesController.php
```diff
@@ -13,43 +13,29 @@
...
-class RouteAttributesController extends AbstractFOSRestController
+class AttributesController extends AbstractFOSRestController
 {
-    /**
-     * @return View view instance
-     *
-     * @Rest\View()
-     */
     #[Rest\Get(path: '/{page}', name: 'product_list', requirements: ['page' => '\d+'], defaults: ['_format' => 'json'])]
     public function listAction(int $page)
```
mbabker, line 26: It seems like the `#[Rest\View()]` attribute is missing here?  The annotation is removed but the attribute isn't added like the other action below.

W0rma: Thank you for the hint. Should be fixed now

Resources/doc/annotations-reference.rst
W0rma: The documentation is currently broken anyway (see https://github.com/FriendsOfSymfony/FOSRestBundle/issues/2329). So maybe this doesn't really matter.

W0rma:
@goetas Please let me know how I can help to move this PR forward.

I will finish https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2327 as soon as this PR got merged

deluxetom (approved):
Awesome work @W0rma, thank you!

Resources/doc/annotations-reference.rst
W0rma: I started using the `::tabs` syntax like discussed in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2348#issuecomment-968292705

composer.json
goetas: :+1:

PR #2327, opened 2021-08-19
Add support for native PHP8 QueryParam, FileParam and RequestParam attributes

### LINKED ISSUE #2301: Missing Attribute Annotation

W0rma:
Closes #2301

### DIFF
--- a/Controller/Annotations/QueryParam.php
+++ b/Controller/Annotations/QueryParam.php
@@ -17,12 +17,42 @@ use Symfony\Component\HttpFoundation\Request;
  * Represents a parameter that must be present in GET data.
  *
  * @Annotation
+ * @NamedArgumentConstructor
  * @Target({"CLASS", "METHOD"})
  *
  * @author Alexander <iam.asm89@gmail.com>
  */
+#[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD)]
 class QueryParam extends AbstractScalarParam
 {
+    /**
+     * @param mixed $requirements
+     * @param mixed $default
+     */
+    public function __construct(
+        string $name = '',
+        ?string $key = null,
+        $requirements = null,
+        $default = null,
+        array $incompatibles = [],
+        string $description = '',
+        bool $strict = false,
+        bool $map = false,
+        bool $nullable = false,
+        bool $allowBlank = true
+    ) {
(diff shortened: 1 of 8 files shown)


Controller/Annotations/QueryParam.php
```diff
@@ -17,12 +17,40 @@
...
  */
+#[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD)]
 class QueryParam extends AbstractScalarParam
 {
+    /**
+     * @param mixed $requirements
+     * @param mixed $default
+     */
+    public function __construct(
+        string $name = '',
```
GuilhemN, line 33: Shouldn't we add bc layers for these classes, like symfony does?

W0rma: @GuilhemN I guess you refer to the BC layer for the `Route` annotation.

I'm not sure how such BC layer should look like here.
Before my changes the constructor of the annotations were just empty.

Since all newly added parameters have default values it is still possible to call `new QueryParam()` and the behavior of this call should not change.

mbabker:
How about making the annotations reader for the `FOS\RestBundle\Request\ParamReader` class optional?  Right now, the class has a hard dependency on `doctrine/annotations` and all the Symfony classes made the `Reader` argument optional when attributes support was added.

mbabker (approved):

W0rma:
> How about making the annotations reader for the `FOS\RestBundle\Request\ParamReader` class optional? 

Good idea. Thank you for the hint! See a0377b2

diegozavaletatelanto:
Amazing feature, this will help us, to use PHP Attributes instead of annotations (we need to import the Class, but looks like unused import).
Thanks for your contribution.

goetas:
@W0rma  do you have anything to add to this PR? I personally think that is fine and I would like to proceed with the merge, what do you think?

W0rma:
@goetas I tested my branch in a real life project and it worked fine. Therefore, I think it can be merged.

I'm looking forward to the next release with symfony6 and full attribute support :)

PR #2337, opened 2021-10-07
Fix compatibility with PHP 8.1

W0rma:
The remaining deprecation warnings in the PHP 8.1 pipeline will be fixed by the following PRs:

>   1x: PHPUnit\Runner\DefaultTestResultCache implements the Serializable interface, which is deprecated. Implement __serialize() and __unserialize() instead (or in addition, if support for old PHP versions is necessary)
>     1x in DebugClassLoader::loadClass from Symfony\Component\ErrorHandler

--> https://github.com/symfony/symfony/pull/43200 (PHPUnit 9.5 instead of PHPUnit 9.4 will be used)

>  2x: substr_count(): Passing null to parameter `#1` ($haystack) of type string is deprecated
>    1x in FormatListenerTest::testOnKernelControllerNegotiation from FOS\RestBundle\Tests\EventListener
>    1x in FormatListenerTest::testUseSpecifiedFormat from FOS\RestBundle\Tests\EventListener

--> https://github.com/symfony/symfony/pull/43357

>  1x: JMS\SerializerBundle\DependencyInjection\Compiler\ServiceMapPass implements the Serializable interface, which is deprecated. Implement __serialize() and __unserialize() instead (or in addition, if support for old PHP versions is necessary)
>    1x in DependencyInjectionTest::testSerializerRelatedServicesAreNotRemovedWhenJmsSerializerBundleIsEnabled from FOS\RestBundle\Tests\Functional

--> https://github.com/schmittjoh/JMSSerializerBundle/pull/872

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -35,6 +35,8 @@ jobs:
             symfony-require: "5.2.*"
           - php-version: 8.0
             composer-flags: "--ignore-platform-reqs --prefer-stable"
+          - php-version: 8.1
+            composer-flags: "--prefer-stable"
 
     steps:
       - name: "Checkout"
--- a/EventListener/BodyListener.php
+++ b/EventListener/BodyListener.php
@@ -125,7 +125,7 @@ class BodyListener
 
     private function isFormRequest(Request $request): bool
     {
-        $contentTypeParts = explode(';', $request->headers->get('Content-Type'));
+        $contentTypeParts = explode(';', $request->headers->get('Content-Type', ''));
 
         if (isset($contentTypeParts[0])) {
             return in_array($contentTypeParts[0], ['multipart/form-data', 'application/x-www-form-urlencoded']);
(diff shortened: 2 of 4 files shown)


goetas:
@W0rma does it mean that we need to wait for https://github.com/symfony/symfony/pull/43200 to be released to have the pipeline green?

W0rma:
@goetas I guess the pipeline won't get green unless all three mentioned PRs are released.

goetas:
ok, moving it to 3.2

mbabker:
> --> [symfony/symfony#43200](https://github.com/symfony/symfony/pull/43200) (PHPUnit 9.5 instead of PHPUnit 9.4 will be used)

For that one, you should be able to set the `SYMFONY_PHPUNIT_VERSION` var either in the Actions config or `phpunit.xml`, it shouldn't need to wait for an upstream release for a changed default value.

W0rma:
@goetas After considering the latest symfony and jms-serializer-bundle releases the PHP 8.1 is now green :)

goetas:
thank you @W0rma !

PR #2340, opened 2021-10-15
Add Symfony 6 support

### LINKED ISSUE #2332: Symfony 6 compatibility
Hello !

I took a little time to check what https://github.com/symfony/symfony/issues/43021 involve for this bundle. And the only blocker is https://github.com/FriendsOfSymfony/FOSRestBundle/blob/3.x/Controller/AbstractFOSRestController.php#L39 where we must add an `array` return type.
As this is a BC and respecting semver, this modification should occurs in a 4.0 release. But I want first of all discuss with you about all the possibilities.

There's is another thing, the API changes on authenticators passports, and the [PassportInterface](https://github.com/symfony/symfony/blob/5.4/src/Symfony/Component/Security/Http/Authenticator/Passport/PassportInterface.php#L25) deletion will make the tests failed by the end of november, because `symfony/security-http` is upstream to `symfony/security-bundle`.

Finally, tests are failing with `symfony/routing` v6, as this package is an upstream dependency of `symfony/framework-bundle`, it will be automatically installed with the release of 5.4 in november too. It will be more careful to block this version installation until the resolution of https://github.com/FriendsOfSymfony/FOSRestBundle/issues/2301, and then check if tests are passing.

- [ ] Make AbstractFOSRestController compatible with new API
- [ ] Make tests pass with `symfony/security-http` v6
- [ ] Make tests pass with `symfony/routing` v6

### LINKED ISSUE #2334: Use InputBag for Request::$request property when able
The `BodyListener` attempts to decode a request and replaces the parameter bag with the decoded data.  When introduced in Symfony 5.1, depending on the initialization path, the `$request->request` property could be either a `ParameterBag` or an `InputBag`, but in 5.3 the Request class was standardized and it always initializes an `InputBag`.  This bundle should use an `InputBag` when the class is available when overwriting `Request::$request`.

Note, the current way of overriding the property could result in folks missing the deprecations the `InputBag` was introduced with regarding accessing array values from the request data, so ensuring the right object is in use could be important to not blindsiding folks with that one.

mbabker:
Replaces #2330
Resolves #2332
Fixes #2334
Includes #2341

This will drop support for Symfony 5.0 thru 5.2, since these branches are already past their end of support dates this shouldn't be a huge issue.


Controller/AbstractFOSRestController.php
```diff
@@ -36,7 +36,7 @@ protected function getViewHandler()
     /**
      * {@inheritdoc}
      */
-    public static function getSubscribedServices()
+    public static function getSubscribedServices(): array
```
mbabker: Potential B/C issue, the return type is added in the FrameworkBundle

Serializer/Normalizer/FlattenExceptionNormalizer.php
```diff
@@ -37,7 +37,7 @@ public function __construct(ExceptionValueMap $statusCodeMap, ExceptionValueMap
         $this->rfc7807 = $rfc7807;
     }
 
-    public function normalize($exception, $format = null, array $context = [])
+    public function normalize($exception, $format = null, array $context = []): array
```
mbabker, line 40: Potential B/C issue, the return type is added in the Serializer

Serializer/Normalizer/FlattenExceptionNormalizer.php
```diff
@@ -74,7 +74,7 @@ public function normalize($exception, $format = null, array $context = [])
         }
     }
 
-    public function supportsNormalization($data, $format = null, array $context = [])
+    public function supportsNormalization($data, $format = null, array $context = []): bool
```
mbabker, line 77: Potential B/C issue, the return type is added in the Serializer

W0rma:
@mbabker Thank you very much for working on this!

I have submitted #2341 which should fix the BC layer of the `Route` annotation for symfony 6.

Serializer/Normalizer/FlattenExceptionNormalizer.php
mbabker: Actually, since this class is final & internal, the signature changes shouldn't matter much here.

Serializer/Normalizer/FlattenExceptionNormalizer.php
mbabker: Actually, since this class is final & internal, the signature changes shouldn't matter much here.

Tests/Functional/app/CustomGuardAuthenticator/services.php
```diff
@@ -10,10 +10,14 @@
  */
 
 return function (Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator $container) {
-    // BC layer to avoid deprecation warnings in symfony < 5.3
-    if (class_exists(Symfony\Bundle\SecurityBundle\RememberMe\FirewallAwareRememberMeHandler::class)) {
+    if (class_exists(Symfony\Component\Security\Http\Authenticator\AbstractAuthenticator::class) && method_exists(Symfony\Component\Security\Http\Authenticator\AbstractAuthenticator::class, 'createToken')) {
+        // Authenticator for use on Symfony 5.4 and newer
         $tokenAuthenticatorClass = \FOS\RestBundle\Tests\Functional\Bundle\TestBundle\Security\ApiTokenAuthenticator::class;
+    } else if (class_exists(Symfony\Bundle\SecurityBundle\RememberMe\FirewallAwareRememberMeHandler::class)) {
```
W0rma:
```suggestion
    } elseif (class_exists(Symfony\Bundle\SecurityBundle\RememberMe\FirewallAwareRememberMeHandler::class)) {
```

This should fix the stylecheck

W0rma:
@mbabker Feel free to cherry-pick my commit from https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2341

Controller/AbstractFOSRestController.php
goetas: So this will trigger a fatal error for anyone that extends the AbstractFOSRestController and overrides this method without specifying the return type, right?

goetas: I'm not sure how much worth is it, but I've published a merge request to your branch with a backward compatible version of this in https://github.com/mbabker/FOSRestBundle/pull/1

mbabker: > So this will trigger a fatal error for anyone that extends the AbstractFOSRestController and overrides this method without specifying the return type, right?

Right, once the return type is declared somewhere in the hierarchy all sub-classes have to have it.  So, your PR would handle that in a B/C way until 4.0 ships.  And given how service subscribers are somewhat encouraged and that it's more likely for folks to have an extended version of this method in controllers, I'd say your PR is worth the extra effort.

goetas: @mbabker feel free to merge my pull request. after that i think that is if the build is green for this we can merge it!

mbabker: We'll need #2341 first then we should be good.

Controller/AbstractFOSRestController.php
```diff
@@ -12,15 +12,22 @@
 namespace FOS\RestBundle\Controller;
 
 use FOS\RestBundle\View\ViewHandlerInterface;
-use Symfony\Bundle\FrameworkBundle\Controller\AbstractController;
+use Symfony\Contracts\Service\ServiceSubscriberInterface;
 
+$ref = new \ReflectionMethod(ServiceSubscriberInterface::class, 'getSubscribedServices');
+
+// Has the ServiceSubscriberInterface a return type hint
+if ($ref->getReturnType() !== null) {
```
W0rma:
```suggestion
if (null !== $ref->getReturnType()) {
```

... to fix the style check

Controller/PreSymfony6AbstractFOSRestController.php
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+ * @internal
+ */
+abstract class PreSymfony6AbstractFOSRestController extends AbstractController
+{
+    use ControllerTrait;
+
+    /**
+     * {@inheritdoc}
```
W0rma:
```suggestion
     * @return array
```

Should we add this return type hint in order to silence the deprecation notice in the symfony 5.4 pipeline?

```
2x: Method "Symfony\Contracts\Service\ServiceSubscriberInterface::getSubscribedServices()" might add "array" as a native return type declaration in the future. Do the same in implementation "FOS\RestBundle\Controller\PreSymfony6AbstractFOSRestController" now to avoid errors or add an explicit @return annotation to suppress this message.
    2x in AllowedMethodsTest::testAllowHeader from FOS\RestBundle\Tests\Functional
```

goetas:
Does anyone of you have an idea why https://github.com/FriendsOfSymfony/FOSRestBundle/runs/4215121790?check_suite_focus=true is failing?

mbabker:
PHP-CS-Fixer doesn’t support SF 6 yet, https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/6095

Adding a `composer remove` step would get around that one

W0rma:
> Adding a `composer remove` step would get around that one

@mbabker I prepared https://github.com/mbabker/FOSRestBundle/pull/2 which should fix the symfony6 pipeline. Feel free to merge it into your branch.

W0rma:
@mbabker I prepared https://github.com/mbabker/FOSRestBundle/pull/2 which should fix the symfony 6 pipeline.

See https://github.com/W0rma/FOSRestBundle/actions/runs/1475592791

The pipelines for symfony 5.4 and 6.0 only show some deprecation warnings triggered by vendor packages.

W0rma (approved):

goetas:
this to me looks good. do you have anyting against waiting for the symfony 6 release or you think that is better to merge it earlier?

mbabker:
Should be good to land now.  Plus, for that small handful of folks who might actually be beta testing 5.4 or 6.0, having this bundle be ready to use would help enable that.

goetas:
thanks everyone for your work! :tada:

PR #2341, opened 2021-10-16
Fix the BC layer of the Route annotation for symfony 6

W0rma:
Like mentioned in https://github.com/FriendsOfSymfony/FOSRestBundle/issues/2332#issuecomment-939538163 and #2340 the BC layer introduced in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2312 for the `Route` annotation does not work properly when using symfony 6.

This PR tries to fix that.

I have tested the changes together with #2340

### DIFF
--- a/Controller/Annotations/Route.php
+++ b/Controller/Annotations/Route.php
@@ -86,22 +86,31 @@ class Route extends BaseRoute
                     $env
                 );
             } else {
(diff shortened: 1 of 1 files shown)


mbabker (approved):

W0rma:
Closing since this PR is now part of #2340

PR #2343, opened 2021-10-26
Fix FlattenExceptionNormalizer without custom SerializeErrorRenderer

### LINKED ISSUE #2342: Changes in 3.1.0 breaking custom flatten exception normalizer
The change in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2294

by changing from:

```php
return $data instanceof FlattenException;
```

to:

```php
return $data instanceof FlattenException && ($context[Serializer::FOS_BUNDLE_SERIALIZATION_CONTEXT] ?? false);
```

Is breaking currently [sulu](https://github.com/sulu/sulu) installations error handling. I'm not sure why this was implemented but would prefer to revert this change. Or when not possible we need to change the behaviour on our side (https://github.com/sulu/sulu/pull/6316). But it is in some case a BC Break as the apis now not return the same errors like before and are not longer handled by the fos rest exception normalizer like before.

/cc @goetas looks for introduced #2294 did now use the normalizer only for serialization errors not for errors in the api itself, which I think should not be the case.

alexander-schranz:
Related: https://github.com/FriendsOfSymfony/FOSRestBundle/issues/2342
Related: https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2344

### DIFF
--- a/Serializer/Normalizer/FlattenExceptionNormalizer.php
+++ b/Serializer/Normalizer/FlattenExceptionNormalizer.php
@@ -86,7 +86,7 @@ final class FlattenExceptionNormalizer implements ContextAwareNormalizerInterfac
         }
 
         // we are in messenger context
-        if (!empty($context[Serializer::MESSENGER_SERIALIZATION_CONTEXT])) {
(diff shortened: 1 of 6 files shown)


goetas:
thank you!

PR #2346, opened 2021-11-01
Allow Symfony 6 in composer.json

alexander-schranz:
Test current state of symfony 6 support.

Requires:

 - [ ] [friendsofphp/php-cs-fixer](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/composer.json) Symfony 6 compatibility


mbabker:
#2340 already covers this in full.

PR #2348, opened 2021-11-13
Setup docs

### LINKED ISSUE #2329: Documentation is borked
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/3.x/Resources/doc/index.rst is borked. None of the links work.

Other pages have similar issues. External links work, but all internal links seem to be borked.

goetas:
https://fosrestbundle.readthedocs.io/en/docs/

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,3 +1,8 @@
 version: 2
+
 sphinx:
   configuration: Resources/doc/conf.py
+
+python:
+  install:
+    - requirements: Resources/doc/requirements.txt
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -2,7 +2,3 @@ version: 2
 
 sphinx:
   configuration: Resources/doc/conf.py
-
-python:
-  install:
-    - requirements: Resources/doc/requirements.txt
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,3 @@
+version: 2
+sphinx:
+  configuration: Resources/doc/conf.py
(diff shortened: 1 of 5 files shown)


W0rma (approved):

W0rma:
https://github.com/FriendsOfSymfony/FOSRestBundle/blob/3.x/Resources/doc/annotations-reference.rst#route is not parsed properly since it relies on the [`sensio.sphinx.configurationblock` extension](https://github.com/fabpot/sphinx-php) which is also used by symfony docs.

I'm not sure if it is worth installing the extension or if we should stop using this syntax.

Edit: Alternatively we could just use the `::tabs` syntax like in https://github.com/readthedocs/readthedocs.org/blob/master/docs/config-file/v2.rst - I guess this should be sufficient for our needs.

I'm glad to rework the section in the docs as soon as this PR got merged.

goetas:
@W0rma  thanks for the review, i've added the support for the custom sensio extensions, could you please check it again?

W0rma:
> @W0rma thanks for the review, i've added the support for the custom sensio extensions, could you please check it again?

Thank you for trying this out!

The code blocks get at least rendered as unordered list.
symfony-docs are using https://github.com/symfony-tools/docs-builder for rendering. However, this tool is internal.

Maybe we should just use the native `::tabs` syntax like in https://github.com/readthedocs/readthedocs.org/blob/master/docs/config-file/v2.rst - IMO it should be sufficient for our needs.

goetas:
Looks much nicer https://fosrestbundle.readthedocs.io/en/docs/annotations-reference.html

W0rma:
Great - thank you for working on this!

I'm going to adjust https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2326 after your PR got merged

PR #2368, opened 2022-04-20
Fix backward compatibility layer for rest route annotations

alexander-schranz:
I'm currently updating: https://github.com/handcraftedinthealps/RestRoutingBundle/pull/18 and did stumble over an issue. That the path here: https://github.com/handcraftedinthealps/RestRoutingBundle/blob/745e0a70c854b55c2b1ff04dc72763b39734fd1a/Tests/Fixtures/Controller/AnnotatedUsersController.php#L170 was not used when using Symfony 6.


alexander-schranz:
Looks like this was already fixed here: https://github.com/FriendsOfSymfony/FOSRestBundle/commit/f4f7471b6030dcef93ed4b129fab24a9d01b6c78 Need to find out why I can not get a newer version then 3.1.1.

PR #2371, opened 2022-06-03
Add alias for serializeFormInterface

Bukashk0zzz:
Fix issue with Symfony 6.1 and error: call_user_func_array(): Argument #1 ($callback) must be a valid callback, class JMS\\Serializer\\Handler\\FormErrorHandler does not have a method "serializeFormInterfaceTojson"

### DIFF
--- a/Serializer/Normalizer/FormErrorHandler.php
+++ b/Serializer/Normalizer/FormErrorHandler.php
@@ -96,6 +96,21 @@ class FormErrorHandler implements SubscribingHandlerInterface
         return $result;
     }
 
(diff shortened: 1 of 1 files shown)


mbabker:
I'd actually say the fix belongs in the Serializer package, not here.  When schmittjoh/serializer#1362 added support for the interface, it didn't define explicit methods, which would cause it to fall back to `JMS\Serializer\Handler\HandlerRegistry::registerSubscribingHandler()` creating the default method names that don't exist.

goetas:
solved in https://github.com/schmittjoh/serializer/pull/1432

PR #2383, opened 2023-01-26
Backport #2319 - Compat with JMS Serializer bundle 4+

Jean85:
This backports #2319, so that JMS serializer 4+ bundle works with 2.x.


deluxetom (approved):

goetas:
i would make this an opt-in feature. it has serious performance implications as it has to check the same thing for the potential parent classes

goetas:
ah, now i have noticed that this is for 2.x

PR #2387, opened 2023-04-01
fix implements HandlerRegistryInterface.getHandler

sgonchar67-dev:


goetas:
Could you please explain what is this change for?

PR #2396, opened 2023-11-06
WIP: Allow installation with Symfony 7

alexander-schranz:
I want to use this as a discussion point how we can keep this bundle compatibilty with the latest Symfony Version. The hardest part mostly will be the deprecation of the [Symfony FrameworkExtraBundle](https://github.com/sensiolabs/SensioFrameworkExtraBundle).

Some background from our side we @sulu use highly the FOSRestBundle and even still use [RestRoutingBundle](https://github.com/handcraftedinthealps/RestRoutingBundle). We are not using any FrameworkExtraBundle so one solution from our side is also skip the ExtraBundle related features when the ExtraBundle is not installed.

# Requires

 - [ ] https://github.com/schmittjoh/JMSSerializerBundle/pull/937 (@mbabker  Symfony 7 PR for JMS Serializer)

# TODO

 - [ ] Fix compatibility
 - [ ] Decide which min Version for PHP (7.4 would currently be required by the php-cs-fixer shim)
 - [ ] Decide what todo with the [Symfony FrameworkExtraBundle](https://github.com/sensiolabs/SensioFrameworkExtraBundle) requirement / features build ontop of it?
    - [ ] ParameterConverter -> ArgumentResolver?
    - [ ] BodyListener -> normal RequestListener?

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -20,10 +20,10 @@ jobs:
     strategy:
       matrix:
         include:
-          - php-version: 7.2
+          - php-version: 7.4
             composer-flags: "--prefer-lowest "
             can-fail: false
-          - php-version: 7.3
+          - php-version: 7.4
             composer-flags: ""
             can-fail: false
           - php-version: 7.4
@@ -49,7 +49,14 @@ jobs:
             symfony-require: "6.3.*"
           - php-version: 8.2
             composer-flags: ""
+            can-fail: false
+            symfony-require: "6.4.*"
+          - php-version: 8.2
+            composer-flags: ""
             can-fail: true # we don't want to fail the build if we are incompatible with the next (unstable) Symfony version
+
+    env:
+      COMPOSER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
     steps:
       - name: "Checkout"
@@ -82,7 +89,6 @@ jobs:
         env:
           SYMFONY_REQUIRE: "${{ matrix.symfony-require }}"
         run: |
-          composer remove friendsofphp/php-cs-fixer --dev --no-update
           composer update --no-interaction --no-progress ${{ matrix.composer-flags }}
 
       - name: "Run PHPUnit"
--- a/composer.json
+++ b/composer.json
@@ -28,43 +28,47 @@
             "Tests/"
         ]
     },
+    "repositories": [
+        {
+            "type": "vcs",
+            "url": "git@github.com:mbabker/JMSSerializerBundle.git"
+        }
+    ],
     "require": {
-        "php": "^7.2|^8.0",
-        "symfony/config": "^5.4|^6.0",
-        "symfony/dependency-injection": "^5.4|^6.0",
-        "symfony/event-dispatcher": "^5.4|^6.0",
-        "symfony/framework-bundle": "^4.4.1|^5.0|^6.0",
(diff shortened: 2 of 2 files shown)


mbabker:
Minimum effort that I think is needed to make this PR mergeable:

- Add a CI build that uninstalls `sensio/framework-extra-bundle` and run it on at least Symfony 7.0 (maybe 6.4 as well, there doesn't need to be many builds for this condition but considering the Sensio bundle won't be Symfony 7 compatible having at least that one build is a hard requirement); this is the smoke test to make sure this bundle at least works when `SensioFrameworkExtraBundle` isn't installed
- Add checks in `FOSRestExtension` when enabling the body converter or the view response listener and throw an exception if those are enabled and `SensioFrameworkExtraBundle` isn't installed (the body converter is hard coupled to the old param converters system and needs to be rewritten as a controller argument resolver, the `@View` annotation/attribute class extends `Sensio\Bundle\FrameworkExtraBundle\Configuration\Template` so that entire feature is unusable without the deprecated bundle)

With this, it at least makes this bundle partially functional on Symfony 7, even if a couple of its biggest features aren't available yet.  At the same time, though, a part of me wonders if it's even worth the effort to release in that state.

mbabker:
> * Decide which min Version for PHP (7.4 would currently be required by the php-cs-fixer shim)

I prefer making decisions based on real life data on this stuff, so let's go to [the Packagist stats](https://packagist.org/packages/friendsofsymfony/rest-bundle/php-stats#3).  The number of installs with PHP 7.2 or 7.3 is so minuscule that bumping to 7.4 should be a no brainer.

alexander-schranz:
Like your ideas thank you @mbabker.

> With this, it at least makes this bundle partially functional on Symfony 7, even if a couple of its biggest features aren't available yet. At the same time, though, a part of me wonders if it's even worth the effort to release in that state.

From @sulu point of view we plan to Support Symfony 7 for our current @sulu 2.x Version. We only use a very minimal set of the RestBundle features:

 - [ ] Rest Routing (not part of FOSRestBundle anymore: https://github.com/handcraftedinthealps/RestRoutingBundle)
 - [ ] ViewHandler features (https://github.com/sulu/sulu/blob/633f23052ab2e87dd5ff4cb18df36577701485f9/src/Sulu/Bundle/SecurityBundle/Controller/ProfileController.php#L106-L112)
     - [ ] ViewHandlerInterface
     - [ ] View::create
     - [ ] Context 

all other features are not used by us. So if the decision would be not to Support Symfony 7 in the official Bundle I mostly would cherrypick our used features into a Fork to avoid BC breaks in our minor release. But maybe other projects could benifit if we make a Symfony 7 release and like Sulu get a longer Time to migrate away from FOSRestBundle to directly use the Serializers and avoid BC breaks in there services.

mbabker:
I've got a more complete PR at #2400 but unfortunately, a lot of tests have to be disabled because of the `FOS\RestBundle\Controller\Annotations\View` annotation extending from `Sensio\Bundle\FrameworkExtraBundle\Configuration\Template` which breaks anything trying to parse annotations/attributes when it runs without `sensio/framework-extra-bundle` installed.  Now to go rewrite that entire component to work without the Sensio bundle 😅

alexander-schranz:
@mbabker when I got some time I can also test @sulu against your branch. There we also dont use the sensio extra bundle.

alexander-schranz:
closing this in favor of https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2400

PR #2400, opened 2023-12-29
Add support for Symfony 7 where able

mbabker:
This PR upgrades most of the bundle to support Symfony 7.  Also wrapped into this PR are:

- #2405
- #2406

The one feature not covered is the request body param converter, which has [its own dedicated PR](https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2398) to deal with implementing an argument resolver replacing that.


DependencyInjection/FOSRestExtension.php
```diff
@@ -256,7 +259,11 @@ private function loadView(array $config, XmlFileLoader $loader, ContainerBuilder
             $container->getDefinition('fos_rest.mime_type_listener')->replaceArgument(0, $config['view']['mime_types']['formats']);
         }
 
-        if ($config['view']['view_response_listener']['enabled']) {
+        if ($this->isConfigEnabled($container, $config['view']['view_response_listener'])) {
+            if (!class_exists(SensioFrameworkExtraBundle::class)) {
+                throw new LogicException('To use the view response listener, the "sensio/framework-extra-bundle" package is required.');
```
W0rma: I tried your branch in my project where I don't use `sensio/framework-extra-bundle` but the view response listener.

After removing this new check everything works like expected.

mbabker: This might've been a bit pedantic because the `#[View]` attribute extends the `sensio/framework-extra-bundle` `#[Template]` attribute and that hierarchy was giving me some headaches.  But, with https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2401 included, anything that I had to disable because of that should be able to be turned back on.

alexander-schranz:
Hello @mbabker I just got today some time to Test @sulu which has no dependency to Sensio Extra Bundle against this pull request.

https://github.com/sulu/sulu/pull/7272

Sulu itself still on Symfony 6.4 but I think it is good to know that all still works like expected also on Sulu side where the Extra Bundle is not installed. And so not any accidently bc break was introduced here which would trigger an error on such project like Sulu. Sulu uses the JMS Serializer inside FOSRest and all seems still work there also like expected.

@mbabker What do you think is required to check to move this forward?

Tests/Fixtures/Controller/ParamsAnnotatedController.php
```diff
@@ -36,12 +39,12 @@ public function getArticlesAction(ParamFetcher $paramFetcher)
     {
     }
 
-    #[QueryParam(name: 'page', requirements: '\d+', default: '1', description: 'Page of the overview')]
+    #[QueryParam(name: 'page', requirements: '\d+', default: '1', description: 'Page of the overview.')]
     #[RequestParam(name: 'byauthor', requirements: '[a-z]+', description: 'by author', incompatibles: ['search'], strict: true)]
-    #[QueryParam(name: 'filters', requirements: '\d+', default: '1', description: 'Page of the overview')]
+    #[QueryParam(name: 'filters', map: true, requirements: new NotNull())]
```
alexander-schranz: based on the CI error I think PHP 8.0 does not support `new NotNull`. This seems to be a PHP 8.1 feature: https://www.php.net/releases/8.1/en.php.

So we may need to revert the NotNull or skip this for PHP 8.0 tests.

mbabker: I'm definitely going for the "just skip on PHP 8.0" option whenever I clean this PR up, otherwise the config between annotations and attributes starts changing and may end up giving different results based on where the config comes from.

mbabker:
> What do you think is required to check to move this forward?

I think for me, in priority order, the PRs to focus on are:

- #2401
- This one
- #2398

I treat #2401 as a blocker to this because I basically disabled the whole `#[View]` attribute/annotation processing and the event listener that converts a `View` returned from a controller into a `Response` on this PR since the attribute class is extending from a class in SensioFrameworkExtraBundle.  So getting that one finished first lets me undo some of the changes I made in this PR to get a (mostly) passing CI.

Finishing this PR up after that gets to a good "we're most of the way there" milestone.

#2398 is probably the hardest to deal with because it is impossible to make a one-for-one replacement of the existing behavior.  But at the same time, it's already an optional integration, so I wouldn't let the param converter not being Symfony 7 compatible block shipping an update that adds Symfony 7 support for all other features in this bundle, and the folks relying on that can continue to use the Symfony 6.4 LTS for a little while longer.

DelphineCogi:
Is there any update here ? 🙏

TheDevick:
This PR is almost 5 months old and still not merged. Any updates about that?

mbabker:
This specific pull request isn't moving forward until its blocker, https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2401, is addressed.  From my end, nothing has been forgotten about here and there is no need for the "updates please?" comments.

Tests/Fixtures/Controller/ParamsAnnotatedController.php
goetas: If it helps I'm ok for simply dropping php 8. 0 support

mbabker: Funny enough, it's just two test cases that need to be skipped because of this.  So, I think we're good on that one for the moment.

mbabker:
For those keeping score at home, this is now rebased on top of the recently merged #2401.  There are some things that need to be undone in this PR because of the changes that the other PR worked around, those will get tackled soon.

Tests/Fixtures/Controller/ParamsAnnotatedController.php
mbabker: Eh, well, maybe not.  Running PHPUnit with the debug flag on 8.0 shows it's blowing up (I'm assuming) when the container is built for the functional tests - https://github.com/FriendsOfSymfony/FOSRestBundle/actions/runs/8543042867/job/23405950001 - skipping a couple of tests on one PHP version is one thing, but if all the functional tests have to be skipped because of that, then dropping everything before 8.1 is going to make more sense.

alexander-schranz: I'm not sure if possible another possibility may would be create a.

```php
#[\Attribute(\Attribute::IS_REPEATABLE | \Attribute::TARGET_CLASS | \Attribute::TARGET_METHOD)]
class QueryParamNotNullPHP80Bridge extends QueryParam {
    public function __construct(
        string $name = '',
        ?string $key = null,
        $requirements = null, 
        $default = null,
        array $incompatibles = [],
        string $description = '',
        bool $strict = false,
        bool $map = false,
        bool $nullable = false,
        bool $allowBlank = true
    ) {
        parent::__construct(
            $name,
            $key,
            $requirements === null ? new NotNull() : $requirements,
            $default,
            $incompatibles,
            $description,
            $strict,
            $map,
            $nullable,
            $allowBlank 
        );
    }
}
```

Can also just be in the test namespace.

mbabker: I was a little worried that wouldn't scale, but it's only 3 attribute configs that needs it.  So, back to being good here.

alexander-schranz (approved):

alexander-schranz:
Looks good from my side. @mbabker as you removed the draft this merge request is so not longer WIP?

W0rma:
I tested this branch in one of my projects and it worked fine. @mbabker Thank you for the hard work!

mbabker:
> @mbabker as you removed the draft this merge request is so not longer WIP?

I knew I forgot something when updating everything else in this PR 😆

But, yeah, aside from the one major feature callout, this should be G2G for use with Symfony 5.4, 6.4, and 7.0.

goetas:
Thanks for the exceptional quality of work! Released as 3.7.0 https://github.com/FriendsOfSymfony/FOSRestBundle/releases/tag/3.7.0

alexander-schranz:
Thx @goetas and thank you @mbabker good work 💪

PR #2406, opened 2024-03-13
feat: update new Symfony versions 6.4, 7.0 and 7.1 and PHP 8.3 matrix in CI

shakaran:
Support with tests for PHP 8.3 and new Symfony versions 6.4, 7.0 and 7.1

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -49,6 +49,18 @@ jobs:
             symfony-require: "6.3.*"
           - php-version: 8.2
             composer-flags: ""
+            can-fail: false
+            symfony-require: "6.4.*"
+          - php-version: 8.3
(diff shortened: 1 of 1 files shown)


mbabker:
Symfony 7 is not supported by this bundle right now with multiple WIP pull requests to get that taken care of, so adding new test builds that are going to always fail because of an incompatible environment isn't really benefiting anyone.

shakaran:
We can mark as allow to fail if you consider it, but I did it to see the really progress on the remaining items to improve. I did it with the idea of help with the progress to speedup the bundle to PHP 8.3 and Symfony 7.0

W0rma:
@shakaran I guess this PR can be closed since https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2400 already added the new symfony/php versions to the CI matrix

Issue #2409, opened 2024-04-08
FOSRest namespace is not defined and argument #2 must be of type array


flohw:
Hello there,

As discussed in #2400 I looked at the error a bit longer this morning and preferred to open an issue for better tracking instead of continuing the thread there. The exception is raised when sensio extra bundle is installed and the controller returns an object which is expected to be serialized by the bundle but is passed to twig instead.

I can confirm that by reverting the weight to 30 on kernel view event it fixes the issue.

The reason is because the `View` attribute extends the Template class if it exists (which exists because sensio extra bundle is installed (check made [line 16 of View](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/3.x/Controller/Annotations/View.php#L16))

As the TemplateListener of the extra bundle has a higher priority than the ViewResponseListener, the Template listener tries to render the template defined by the template guesser ([line 71 of TemplateListener](https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/src/EventListener/TemplateListener.php#L71)) the guesser uses bundle's name, route, controller to define a name (I guess, I didn't looked at it as it's not relevant for us) that's why we get the exception FOSRest namespace is not defined.

The other exception is raised because the `$parameters` returned by the controller are expected to be an array of variables passed to a twig render method which FOSRest used to bypass with its ViewResponseListener (which serialize the controller response so the TemplateListener can't be used ([line 85](https://github.com/sensiolabs/SensioFrameworkExtraBundle/blob/master/src/EventListener/TemplateListener.php#L85) returns true so return l86 is invoked)

I think I can provide a patch but not sure about the test strategy: should I only test the event weight with a comment or should I create a functional test which would trigger that particular case and resolve by changing the event weight?

Before providing the patch, it would be nice if someone else could double check what I explained if I didn't misunderstood the code or missed something else aka I don't have the right explanation of the bug as I'm not an expert of the whole bundle. :slightly_smiling_face: 

Thanks :slightly_smiling_face: 

ping to @mbabker and @benconda as you were the latest involved in the discussion.

mbabker:
I'd do a functional test here because what we want is to test to make sure the response comes back correctly even when the Sensio bundle is installed and has its view integration enabled.  This diff for the tests is enough to cover that:

```diff
diff --git a/Tests/Functional/Bundle/TestBundle/Controller/ArticleController.php b/Tests/Functional/Bundle/TestBundle/Controller/ArticleController.php
[...]

PR #2410, opened 2024-04-09
fix: Sensio extra bundle TemplateListener invocation before view response listener and serialization

flohw:
Hi,

As discussed in #2400 and pointed in #2409 (thanks for the bundles.php, not sure I would have found it without a lot of struggles :sweat_smile: ) here is the fix and test.

I relied on the existing redirect test as this existing one and the one you provided trigger the same error. A test idea which could cover more cases (and may be more common) would be that the controller returns an object to be serialized but it would require to add symfony serializer or jms into the loaded bundles and provide the necessary configuration I think.

Let me know if I really should add the new test you provide or a modified one which could return an `\stdClass` which should be serialized via the view response listener.

Thanks @mbabker

### DIFF
--- a/EventListener/ViewResponseListener.php
+++ b/EventListener/ViewResponseListener.php
@@ -170,7 +170,7 @@ class ViewResponseListener implements EventSubscriberInterface
     {
         return [
             KernelEvents::CONTROLLER => 'onKernelController',
-            KernelEvents::VIEW => ['onKernelView', -128],
+            KernelEvents::VIEW => ['onKernelView', 30],
         ];
     }
 
--- a/Tests/Functional/Bundle/TestBundle/Controller/ArticleController.php
+++ b/Tests/Functional/Bundle/TestBundle/Controller/ArticleController.php
@@ -32,14 +32,26 @@ class ArticleController extends AbstractFOSRestController
     }
 
     /**
-     * @Get("/articles.{_format}", name="get_article", defaults={"_format": "html"})
+     * @Get("/articles.{_format}", name="get_articles", defaults={"_format": "html"})
      *
      * @View()
      */
-    #[Get(path: '/articles.{_format}', name: 'get_article', defaults: ['_format' => 'html'])]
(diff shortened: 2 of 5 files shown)


flohw:
Failing test may be because of deprecation :thinking: 
Not sure how to handle them. I doubt `can-fail: true` is the correct answer and we can't `remove-sensio-bundle: yes` as it's what we want to test ^^

mbabker:
Just update the deprecation helper config.  At some point the test app/bundle configs need some cleanup so the baseline isn't so high, but a lot of it's unavoidable (especially with the Symfony 6.4 builds and the annotations deprecations).

```xml
<env name="SYMFONY_DEPRECATIONS_HELPER" value="max[self]=110&amp;max[indirect]=230"/>
```

That looks to be the right numbers based on the failing builds.

mbabker:
> A test idea which could cover more cases (and may be more common) would be that the controller returns an object to be serialized but it would require to add symfony serializer or jms into the loaded bundles and provide the necessary configuration I think.

IMO it doesn't need to go that in-depth.  The functional tests IMO really just need to be in-depth enough to cover the "make sure it doesn't break with `SensioFrameworkExtraBundle` installed" case and at least one end-to-end scenario with the framework handling a request and ensuring a valid response comes back, the unit tests cover a lot of the internal behaviors much more thoroughly and probably don't need to be re-tested with the functional test cases.

flohw:
Hi Michael,

Thanks for the env hint, I forgot about it. I am more used to gitlab and our configuration don't fail when the deprecation are raised.
I also forget to look at what is tested on unit tests.  :facepalm:  I agree with you that my idea is not necessary.

I added the functional test and have to raise the max self deprecation due to it. I let the indirect deprecation to its original value as it was higher than your recommendation.

Let me know if I have to set them exactly on the max number I found on the actions (112 and 230)

Thanks for your help.

mbabker (approved):

benconda (approved):
Thank you :pray:

PR #2412, opened 2024-08-01
Declare nullable parameter types explicitly

W0rma:
Implicitly nullable parameter types will be deprecated in PHP 8.4: https://wiki.php.net/rfc/deprecate-implicitly-nullable-types

PR #2413, opened 2024-11-03
Test against PHP 8.4

W0rma:


W0rma:
@goetas Do you plan to tag a new version?
The new tag would then contain https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2412 which fixes PHP 8.4 deprecation warnings.

PR #2422, opened 2025-11-18
Don't use deprecated `#[Route]` methods

HypeMC:
In Symfony 7.4 the `Route` attribute getters and setters [have been deprecated in favor of public properties](https://github.com/symfony/symfony/pull/61487).

### DIFF
--- a/Controller/Annotations/Route.php
+++ b/Controller/Annotations/Route.php
@@ -123,7 +123,11 @@ class Route extends CompatRoute
             );
         }
 
-        if (!$this->getMethods()) {
+        if (isset($this->methods)) {
+            if (!$this->methods) {
+                $this->methods = (array) $this->getMethod();
+            }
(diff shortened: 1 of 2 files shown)


Controller/Annotations/Route.php
```diff
@@ -123,7 +123,11 @@ public function __construct(
             );
         }
 
-        if (!$this->getMethods()) {
+        if ($this->isMethodsPropertyPublic()) {
```
alexander-schranz: Curious why not use `if (isset($this->methods)) {` to handle this bc layer? Reflection looks like overhead.

HypeMC: Good point, fixed.

alexander-schranz:
The soap test is fixed in https://github.com/FriendsOfSymfony/FOSRestBundle/pull/2423

PR #2423, opened 2025-11-20
Fix MimeTypeTest for Symfony 7.4

alexander-schranz:
Actually soap is added to the root mime types supported so the test will fail.

https://github.com/symfony/symfony/pull/61267

While adopting the tests I stumbled over that actually the previous tests has side effects because the `Request::$formats` is a static variable which is keept between the difference test. To workaround I did go with `_1` here so not reuse some from the previous test.


alexander-schranz:
@nicolas-grekas is there any other mechanic in Symfony to reset `Request::$formats`?

alexander-schranz:
think the CI errors are unrelated to the changes @goetas

PR #2427, opened 2025-12-21
Add support for symfony v8

W0rma:
This PR fixes the remaining deprecation warnings in symfony 7.4:

> Since symfony/validator 7.3: Passing an array of options to configure the "Symfony\Component\Validator\Constraints\IdenticalTo" constraint is deprecated.
>
> Since symfony/validator 7.4: Support for evaluating options in the base Constraint class is deprecated.
>
> Since symfony/validator 7.3: Passing an array of options to configure the "FOS\RestBundle\Validator\Constraints\Regex" constraint is deprecated
> 
> Since symfony/http-foundation 7.4: Request::get() is deprecated, use properties ->attributes, query or request directly instead.

With these changes it is possible to allow the installation of symfony v8


alexander-schranz:
Awesome thank you @W0rma and thx @goetas. @goetas If there is something else needed to get this released let me know.

goetas:
https://github.com/FriendsOfSymfony/FOSRestBundle/releases/tag/3.9.0-beta1

composer.json
```diff
@@ -31,14 +31,14 @@
     "require": {
         "php": "^7.4|^8.0",
         "symfony/config": "^5.4|^6.4|^7.0",
```
W0rma, line 33: Seems like I forgot to adjust this version constraint :man_facepalming: 

I'm going to create PR to fix this.

W0rma:
I tested https://github.com/FriendsOfSymfony/FOSRestBundle/releases/tag/3.9.0-beta2 in one of our projects and it works fine.
We do not use all features of RestBundle, though.

@alexander-schranz Have you already tested the beta release with Sulu? Does it work well?

alexander-schranz:
@W0rma sadly we are not there with @sulu to test against Symfony 8 as we still have other third party blocking the upgrade. I can just tell that on Symfony 7.4 the beta still works as expected and didn't break something for us.

PR #2431, opened 2026-01-10
Fix usage of null as an array offset which is deprecated in PHP 8.5

W0rma:
> Using [null](https://www.php.net/manual/en/reserved.constants.php#constant.null) as an array offset or when calling [array_key_exists()](https://www.php.net/manual/en/function.array-key-exists.php) is now deprecated. Instead an empty string should be used.

https://www.php.net/manual/en/migration85.deprecated.php#migration85.deprecated.core.using-null-as-an-array-offset

### DIFF
--- a/EventListener/AllowedMethodsListener.php
+++ b/EventListener/AllowedMethodsListener.php
@@ -41,10 +41,14 @@ class AllowedMethodsListener
 
         $allowedMethods = $this->loader->getAllowedMethods();
 
-        if (isset($allowedMethods[$event->getRequest()->attributes->get('_route')])) {
+        $route = $event->getRequest()->attributes->get('_route');
+        if (null === $route) {
+            $route = '';
+        }
+        if (isset($allowedMethods[$route])) {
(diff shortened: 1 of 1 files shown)


EventListener/AllowedMethodsListener.php
```diff
@@ -41,10 +41,14 @@ public function onKernelResponse(ResponseEvent $event): void
 
         $allowedMethods = $this->loader->getAllowedMethods();
 
-        if (isset($allowedMethods[$event->getRequest()->attributes->get('_route')])) {
+        $route = $event->getRequest()->attributes->get('_route');
+        if (null === $route) {
+            $route = '';
+        }
+        if (isset($allowedMethods[$route])) {
```
W0rma, line 48: `null` is interpreted as empty string in lower PHP versions as well: https://3v4l.org/kkPK5

Kezino:
Very much needed, please merge

deluxetom (approved):

